PR #10, opened 2019-02-18
Decoupled mode for Xiaomi switches

KapJI:
- Add documentation about decoupled mode
- Fix formatting of this page

It seems that on zigbee2mqtt.io tables are not rendering :(

### DIFF
--- a/information/mqtt_topics_and_message_structure.md
+++ b/information/mqtt_topics_and_message_structure.md
@@ -212,33 +212,40 @@ This command also allows to redefine which button controls which relay for doubl
 
 Special topics should be used:
 
-`zigbee2mqtt/[DEVICE_ID]/system/set` to modify operating mode.
+`zigbee2mqtt/[DEVICE_ID]/system/set` to modify operation mode.
 
 Payload:
 ```js
 {
-  "decoupled_mode": {
+  "operation_mode": {
     "button": "single"|"left"|"right",
-    "state": "decoupled"|"toggle_left"|"toggle_right"
+    "state": "VALUE"
   }
 }
 ```
+
+Values                | Description
+----------------------|---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Koenkk:
To fix the table rendering, you should put an empty line before and after each table

information/mqtt_topics_and_message_structure.md
```diff
@@ -191,19 +192,54 @@ Execute selftest
...
+    "button": "single"|"left"|"right"
+  }
+}
+```
+
+Response      | Description
+--------------|-----------------------------------------------
+18            | Button controls left relay
+34            | Button controls right relay
+254           | Button is decoupled from the relays
```
Koenkk: Instead of returning numeric codes, why is this converter not just returning text like: `control_left_relay`, `decoupled`? That would avoid users having to lookup the meaning of the code.

KapJI:
Because this converter only sends query. To parse this, new converter should be added to read the response. But yeah, I can update this.

information/mqtt_topics_and_message_structure.md
KapJI: Because this converter only sends query. To parse this, new converter should be added to read the response. But yeah, I can update this.

Koenkk: I don't understand, where is this response is currently visible?

KapJI: Right now only in debug logs of zigbee2mqtt.

Koenkk: With a 'No converter available..' message?

KapJI: No, it has ignore converter.
I updated it in https://github.com/Koenkk/zigbee-shepherd-converters/pull/290.

Koenkk:
Great, thanks!

PR #16, opened 2019-02-23
Update configuration.md

pixeldoc2000:
Change port to a more standard port. Note about ZLL Channels. Security Note about network_key.


Koenkk:
Thanks!

PR #26, opened 2019-03-03
Further enhance table layout

TheZoker:
Makes table layout more "stable"

(@Koenkk you can adjust the width if you like, but for me these values have the best information ratio)


Koenkk:
looks good, thanks!

PR #28, opened 2019-03-04
Update instructions for alternative-flahsing-method with CCLoader.

kianusch:
I hope this is the correct procedure to make a PR.


Koenkk:
Thanks!

PR #29, opened 2019-03-05
How to create CC2531 clone

andreasbrett:
Added some documentation on how to clone a CC2531 by copying IEEE/MAC address to a backup stick.


andreasbrett:
Thanks to orxaz who brought this to my attention in the following issue:
https://github.com/Koenkk/Z-Stack-firmware/issues/47

Koenkk:
Are you sure this is necessary? I also though this, but later figured out it wasnt needed.

andreasbrett:
Not perfectly sure as the zigbee specification is a bit vague with regards to that situation. It should be able to take down one coordinator and start up another one with the same network spanning parameters (pan Id, key etc) for the routers and end devices to automatically re-join.

Not sure though how well this works in practice as all routers will have their network tables (the mapping of nwkaddress to IEEEaddress) and depending how quickly they refresh those or check for validity routing might be disrupted temporarily as the route back to the coordinator should not work anymore. Otherwise it's fair to assume that all routers will immediately update their network tables once a coordinator spans up a network and they join. At least they would add this new coordinator's ieee address to their table and routing should work again. 

Since those steps only take a few seconds I'm happy to just do them in addition to having my backup stick flashed and in store.

Koenkk:
I just checked the following on my production network (40+ devices):
- Stop zigbee2mqtt
- Replace CC2531 with a different CC2531 (**without** copying ieee address)
- Start zigbee2mqtt
- End devices worked immediately (tested with Xiaomi motion, temperature, door/window sensor)
- Group addressing worked immediately
- Direct device addressing worked after trying a few times (I see the same when re-flashing the same CC2531, the coordinator needs to retrieve the network routes).

Based on this, I cannot conclude anything else that we don't know what value this would add. 
Also I don't actually know what the second ieee address does.

I'm happy to merge this, but first we need to find out what it does/improve.

andreasbrett:
You're absolute right and thanks for your test. It's good to know that it's not exactly necessary. I don't know what other benefit a cloned mac would have. If routing works right away... well, what more to ask.

Koenkk:
No problem, things likes this are still very interesting.

I really appreciate that you are finding out these kind of things!

joonas-fi:
I tried:

- Stop zigbee2mqtt
- Change CC2531 stick (same FW, but different MAC address)
- Start zigbee2mqtt

Controlling my lights (IKEA Trådfri power plug / router) didn't work. I had to stop zigbee2mqtt, plug the original stick back in, start zigbee2mqtt and it worked right away.

My firmware version is 20180815

PR #36, opened 2019-03-10
Trust Wireless contact sensor pairing

Kryzek:
Trust Wireless contact sensor needs to be awake a bit longer when pairing to get it fully configured.


Koenkk:
Thanks!

PR #47, opened 2019-03-14
Fix incorrect homeassistant.js mount path

GreenVine:
This PR fixes the Docker documentation where `lib/extension/homeassistant.js` should be mounted directly under `/app`, rather than `/app/node_modules`.


Koenkk:
Thanks!

PR #52, opened 2019-03-19
Add mqtt debug tool

thomasnordquist:
I use this tool a lot to integrate new devices in my node-red instance, see which topics update, and which values the different devices publish.
In fact, zigbee2mqtt was a big reason for me to write the application.

Sure, the link also advertising my app, but I really believe that this application is very helpful for everyone that is not using an out-of-the-box solution. The topic hierarchy makes MQTT super easy to understand and use.

Ohne Titel
[MQTT Explorer](https://thomasnordquist.github.io/MQTT-Explorer/?c=pr)


Koenkk:
Thanks for this useful tool!

thomasnordquist:
Thanks for adding the link 👍.

PR #56, opened 2019-03-22
added backslash newlines for docker run

renoutg:


renoutg:
added prerequisites for at least fedora, maybe gcc-c++ is required for others as well

Koenkk:
thanks!

PR #59, opened 2019-04-09
Add restart req after config to Group doc

ronschaeffer:
Was getting a "xxx group doesn't exist" error when attempting to add. I restarted Home Assistant (Hassio) and doublechecked everything several times before realizing that I needed to restart the Home Assistant add-on. I don't know what the procedure is with other types of Home Assistant implementation.


Koenkk:
Thanks!

PR #66, opened 2019-05-19
Update device_specific_configuration.md

u17194907425:
An important note about the behavior of motion sensor in "Normal User Mode"


Koenkk:
Thanks!

PR #70, opened 2019-06-03
Added information regarding Xiaomi Wall Switch (without neutral) not working as router

VivantSenior:


Koenkk:
Thanks for the pr. This file is automatically generated from `devices.js`, could you update: https://github.com/Koenkk/zigbee-shepherd-converters/blob/master/devices.js#L218?

PR #72, opened 2019-06-04
Update supported_devices.md

ChristofferG:
Added brightness up/down/stop to IKEA E1743


Koenkk:
Thanks for the PR, however this file shouldn't be edited manually, please update https://github.com/Koenkk/zigbee-shepherd-converters/blob/master/devices.js#L659 and changes will automatically land here.

PR #73, opened 2019-06-15
add support section

Data-Monkey:
add support section, it should be obvious as you have these links in the first line, but I struggled to find them ...


Koenkk:
Thanks!

PR #77, opened 2019-06-22
Instructions to factory reset hue dimmer rather than hue bulb

clockbrain:
This may address https://github.com/Koenkk/zigbee2mqtt/issues/1619#issuecomment-502476524

Instructions derived from https://www2.meethue.com/en-au/support/dimmer-switch/top/troubleshooting-a-hue-dimmer-switch

### DIFF
--- a/devices/324131092621.md
+++ b/devices/324131092621.md
@@ -14,14 +14,13 @@
 
 
 ### Pairing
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #82, opened 2019-06-30
Made /devices show content from "supported_devices" page

jabelone:
If you are in a device specific page and go back to the parent directory it 404s. This will instead just show the info from the supported devices page.


jabelone:
Ahh, I didn't quite realise how this all works. I'll close this until my device PR in zigbee-shepherd-converters goes through and reopen this.

PR #83, opened 2019-07-05
Use line thickness instead of dotted/dashed for network map

clockbrain:
Goes with matching PR for network map line style in zigbee2mqtt.


Koenkk:
Thanks!

PR #90, opened 2019-07-18
Added a little SEO for devices

G1K:
documentation and device information through google is hard to find. 
The repository is ranked mainly on github, not the site. The main reason is meta tags. 

I recommend to close the repository from indexing https://github.com/Koenkk/zigbee2mqtt.io , you can do this by simply removing the master branch, and transfer the content for example to gh-pages. 

Then Google will rank the pages of the documentation site.

PR #91, opened 2019-07-18
reorganizing files

G1K:
In explanation
https://github.com/Koenkk/zigbee2mqtt.io/pull/90#issuecomment-512897827

On the way:
- removed the double pictures the root
- corrected md files to just FrontMatter processed, and is not copied and the originals in the web
- corrected links to canonical URLs of device pages

**On a note:**

> all md files must start with a section for variables, you can even empty, otherwise the md file is copied and available on the web.
Example:
http://www.zigbee2mqtt.io/getting_started/what_do_i_need.md
```
---
---
```


G1K:
recommendations to exclude duplicates: 
1. Make 301 redirect from all versions of the site to the main. 
- http://www.zigbee2mqtt.io
- https://zigbee2mqtt.io
- http://zigbee2mqtt.io
to
- https://www.zigbee2mqtt.io

> The truth is now in the index more pages with http www , but if you already have https it is better to go to it

2. If possible, send a request to move to Google Search Console

You can do for example through cloudflare (but there is a chance that some users will not be available site due to IP locks), and perhaps through aws can be. 

I have another offer to look at the current situation in the Google webmaster tools verification code unable to send a separate PR

Koenkk:
Thanks!

Koenkk:
Something seems to be wrong http://www.zigbee2mqtt.io now shows the README.md, do you know how to fix this?

Koenkk:
I've temporarily reverted this for now.

G1K:
Unfortunately I didn't know about this feature. github-pages works correctly in normal jekyll. Found a reference here https://help.github.com/en/articles/configuring-jekyll 
> Keep in mind that if you change the source setting, your pages may not build correctly. GitHub Pages only considers source files in the top-level directory of a repository.

I tested it on my repository. 
There are no other options but. 
Even if Declare the main branch not master, and master to remove.  

Variant of solution. Keep the docs folder in the master branch and select it in the repository settings ( in case of local build it will be taken from the config). 

image

But if you delete a branch and move to gh-pages , it still takes content from the root of the repository.

PR #93, opened 2019-07-21
Wrap the Home Assistant config example with Jekyll raw tags

DominikPalo:
The first YAML config example in HA docs https://www.zigbee2mqtt.io/integration/home_assistant.html is rendered with incorrect indentation
image, so I've wrapped this block with raw tags to force Jekkyl preserving the indentation.


Koenkk:
Thanks!

PR #96, opened 2019-07-28
Update AC08562.md

str3000:
add parameter range


Koenkk:
As this applies to all brightness command I added it to a more central place, thanks!

PR #100, opened 2019-08-03
Fixes incorrect label for E1603/E1702 notes

PedroLamas:
Page for E1603/E1702 is missing the notes due to the fact that the devide is identified as `E1603/E1702` when notes indicate `E1603`.


Koenkk:
Thanks! :bee:

PR #101, opened 2019-08-04
Home assistant configuration updates

scootec:
Added some example yaml for executing Home Assistant scripts to rename and remove devices in Zigbee2mqtt.

Let me know if you would prefer this to be a part of the _Controlling Zigbee2mqtt via Home Assistant_ section of the docs rather than it's own section.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
 _site/
+.sass-cache/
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks, could you move this to the home assistant part of the documentation?

scootec:
@Koenkk I have moved to existing home assistant yaml docs, and have also updated the Home Assistant configuration for Lovelace, removed the group which is not needed anymore (Lovelace), as well as other tidy ups.

Let me know if this should be split up.

Koenkk:
No need to split up, thanks!!

PR #103, opened 2019-08-10
Reformulated and added info for Tradfri remote.

iBobik:
It is not needed to have bulb close to CC2531. If you have Tradfri remote you can use it for much more convinient pairing.

Also removing "After resetting the bulb will automatically connect." because it does not behave like this.

### DIFF
--- a/docgen/device_page_notes.js
+++ b/docgen/device_page_notes.js
@@ -288,12 +288,12 @@ small pin or paperclip to push the reset button once.
         note: `
 ### Pairing
 Factory reset the light bulb ([video](https://www.youtube.com/watch?v=npxOrPxVfe0)).
-After resetting the bulb will automatically connect.
-
-While pairing, keep the bulb close the the CC2531 USB sniffer.
-
 What works is to use (very) short “on’s” and a little bit longer “off’s”.
 Start with bulb on, then off, and then 6 “on’s”, where you kill the light as soon as the bulb shows signs of turning on.
+
+While pairing, keep the bulb close the the CC2531 USB sniffer. Pairing may also be possible with the
+[Tradfri Remote Control](https://www.ikea.com/us/en/images/products/tradfri-remote-control__0489469_PE623665_S4.JPG)
+by pressing and holding the reset button on the bottom of the remote (next to the battery) close to the bulb.
 `,
     },
     {


Koenkk:
Why is *After resetting the bulb will automatically connect.* not correct?

iBobik:
It will not connect because after resetting it is not paired. After pairing (have to be put near pairing device - gateway or remote) it will automatically connect.

Koenkk:
But after it resetting the bulb it will enter pairing mode and automatically connect, right? 

I'm observing the following: 
- reset bulb by 5x on/off fast
- wait some seconds
- bulb automatically connects

iBobik:
For me it connected only to device in near proximity what was in pairing mode. It does nto behave like it only when it is not paired already. We can call it "connect automatically", it depends on definition :-)

Koenkk:
@kiall that also what I observe indeed

In my opinion the following description would be better suited:

```
In order to pair it we need to factory reset the bulb ([video](https://www.youtube.com/watch?v=npxOrPxVfe0)). After factory resetting the bulb it will automatically connect.

If you cannot get the bulb to reset, try to use (very) short “on’s” and a little bit longer “off’s”.
Start with bulb on, then off, and then 6 “on’s”, where you kill the light as soon as the bulb shows signs of turning on.

Pairing may also be possible with the
[Tradfri Remote Control](https://www.ikea.com/us/en/images/products/tradfri-remote-control__0489469_PE623665_S4.JPG)
by pressing and holding the reset button on the bottom of the remote (next to the battery) close to the bulb.
```

However I think the tradfri remote control pairing should require some extra explanation. As from my understanding, when the tradfri remote itself is not connected to the network, it will setup an own network with only the bulb and remote in it.

iBobik:
I also paired the remote, just by holding button near the gateway. From that time the remote stopped controlling lights what were paired with it, so it seems that pairing it with the z2m gateway created new network.

Jan Pobořil
https://honza.poboril.cz/

16. 8. 2019 v 18:38, Koen Kanters <notifications@github.com>:

Koenkk:
What are the exact steps to pair a TRADFRI bulb to zigbee2mqtt using the remote control?

iBobik:
Reset the bulbs (if they are not new), pair remote, pair bulbs by the remote.

Jan Pobořil
https://honza.poboril.cz/

16. 8. 2019 v 18:58, Koen Kanters <notifications@github.com>:

Koenkk:
I've tried the following:
- Started with an unpaired bulb and remote
- Paired remote
- Tried to pair bulb by holding the reset button of the remote while holding the remote against the bulb -> Bulb doesn't respond to the remote at all
- Paired bulb by 6x on/off
- When now holding the reset button of the remote while holding the remote against the bulb the bulb does respond (changes brightness). 

Based on this I conclude that the remote can only interact with the bulb when it is already in the same network OR when the remote is not in a network yet.

Your instructions mention that you reset the bulb before doing this, but I would expect that this already pairs the bulb meaning that the remote doesn't do anything regarding the pairing.

coolguymatt:
@Koenkk I found the same behaviour. IKEA's instructions for certain devices now tell you to connect the bulb to a "steering device" (aka one of their remotes) but if you connect the bulb to the remote, they are in a separate network, and if you connect the remote to zigbee2mqtt, the remote will not communicate with the bulb.

PR #105, opened 2019-08-16
IKEA TRADFRI E1810

kiall:
Add the IKEA E1810, which appears to be a replacement for the E1524. Functionally, they appear to
work identically, the only difference I can see is front finish is matte rather than glossy, and
the internals behind the battery cover are slightly different.

**Read note on zigbee2mqtt PR @ https://github.com/Koenkk/zigbee2mqtt/pull/1866 before merging.**


Koenkk:
Thanks!

sjorge:
Looks like this broke the supported devices page, the remote no longer shows an image and the details page is not found.

kiall:
Oh! I thought that index page was generated. I'll submit a PR to fix soon. In the meantime - it's here https://www.zigbee2mqtt.io/devices/E1524_E1810.html

PR #108, opened 2019-08-22
Fixed the IKEA Tradfri remote control on the supported devices page

jorik12345:
Fixed the IKEA Tradfri remote control on the supported devices overview page after the serial number change (E1810)


Koenkk:
Thanks!

PR #109, opened 2019-08-23
Update device_page_notes.js

chunkysteveo:
Added small note for UK socket as it's "flipped" and the reset pin is at the "bottom".


Koenkk:
Thanks!

PR #115, opened 2019-09-20
Update E1743.md

chunkysteveo:
Updated pair instructions as it was using the "bulb" instructions, which is incorrect - have copied the excellent instruction from E1525 motion sensor which is very accurate, with a couple extra notes for the switch sensor.


Koenkk:
Thanks!

PR #123, opened 2019-10-02
Add docs for E1757/E1926 (IKEA Roller-blinds)

syphernl:
Added docs for E1757 (FYRTUR) and E1926 (KADRILJ) roller-blinds from IKEA.


Koenkk:
Thanks! Small note this page will be visible after the next zigbee2mqtt release (as it's currently not supported in 1.6).

PR #124, opened 2019-10-02
Link back to hass.io how to in support new device how-to

iautom8things:
Add link to help find hass.io specific directions for supporting new devices.


Koenkk:
Thankd

PR #130, opened 2019-10-09
ubisys j1 configuration & calibration information

felixstorm:
Also updated the image to the correct one for the flush-mounted version (in line with the pictures of the other ubisys devices).

Thanks,
Felix


Koenkk:
Thanks!

PR #136, opened 2019-10-13
Resetting Philips Hue Lightstrips

rosscullen:
Updated instructions for resetting Philips hue light bulbs to also include Philips Hue Lightstrips


Koenkk:
Thanks!

PR #151, opened 2019-11-03
[WIP] Move device configuration to devices page

sjorge:
- [ ] `npm run docgen` generates changes on master, move those to the proper location.
- [ ] Add 'Device specific configuration' section to 'notes' section for all devices.
- [ ] Move device configuration to per-device page from MQTT Topics and message structure page

### DIFF
--- a/docs/devices/046677476816.md
+++ b/docs/devices/046677476816.md
@@ -45,6 +45,10 @@ color temperature (if applicable) and color (if applicable) changes. Defaults to
 Note that this value is overridden if a `transition` value is present in the MQTT command payload.
 
 
+### Device specific configuration
+This devices does not have any device specific configuration.
+
+
 ## Manual Home Assistant configuration
 Although Home Assistant integration through [MQTT discovery](../integration/home_assistant) is preferred,
 manual integration is possbile with the following configuration:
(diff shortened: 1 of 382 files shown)


sjorge:
Yeah that was a wasted efford :( will revist this because I totally overlooked the docgen dir.

sjorge:
OK I started from scratch and after a checkout of master imediatly ran `npm run docgen` and there are a bunch of diffs.

I will have to get these in the correct location first it seems.

Koenkk:
Yes, it is greatly out of sync atm, I will try to bring it in sync (as it needs to be prepped for zigbee2mqtt 1.7)

sjorge:
> Yes, it is greatly out of sync atm, I will try to bring it in sync (as it needs to be prepped for zigbee2mqtt 1.7)

How flexible is docgen? Can we add a dict of 'device configuration' ?

Something like 
```
{
 'some_key': 'set some config to some value, range 0-255',
 'other_key': 'set some other thing, values on or off',
}
```

And have it auto generate the section?

Koenkk:
As flexible as we can make it, but first I need to update the docgen because currently it  is a mess.

PR #155, opened 2019-11-11
Updates the mqtt topic in sensitivity and self-test

LEJOUI:
There was a '/' missing at the end of the  mqtt topics in the sensitivity and self-test


Koenkk:
Why is this needed?

LEJOUI:
I'm not sure why, but I was only able to do the self-test and to set the sensitivity after I included the '/'. So I assumed it was a typo.

LEJOUI:
Somehow the message gets through now without the '/'. Closing.

PR #156, opened 2019-11-11
[docs] Update MCCGQ01LM pairing steps

fatmcgav:
I was having issues getting the `MCCGQ01LM` devices paired. 
Reset would get the 3 blinks followed by a few more blinks, however the device would just timeout after 10 seconds with:
`Cannot get the Node Descriptor of the Device: 0x****** (Error: Timed out after 10000 ms)`

These steps worked for me :)


Koenkk:
This was indeed the pairing procedure before zigbee2mqtt 1.7.0, but with zigbee2mqtt 1.7.0 this is not needed anymore.

fatmcgav:
Ahh, ok... Is `1.7.0` available? Or any way I can test with the latest changes to confirm they work?

Koenkk:
Yes it has been released on sunday

fatmcgav:
> Yes it has been released on sunday

Cool, I'll upgrade and give it a go... 😀

PR #157, opened 2019-11-13
New binding behaviour of IKEA remote control

didiht:
Changes due to issue IKEA E1524/E1810 remote control / new behaviour after firmware update #2325 in zigbee2mqtt repo

Since some people have trouble with binding of Enddevices to groups as I saw in serveral comments, I would propose to enhance the binding docu with an example.


Koenkk:
Thanks!

PR #158, opened 2019-11-13
Added Zigbee version column, added new columnfor commands

jasimancas:


Koenkk:
Hi thanks for the PR,  but this file is not meant to be edited manually. It is generated by: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docgen/supported-devices.js. However before making changes here we should first add the necessary information to `devices.js`.

jasimancas:
Ok, thanks.

PR #160, opened 2019-11-15
Added info about binding

nldroid:
Hey, Added som info about binding this remote. Took me a while to figure out why it didn't work.


Koenkk:
Thanks for sharing!

PR #167, opened 2019-11-23
Distinguish between product versions

fvanroie:
HS1SA is a standalone smoke sensor without Zigbee support. There are 3 versions and this commit makes clear that product code HS1SA-M is the Zigbee version.


Koenkk:
Thanks!

PR #168, opened 2019-11-26
Update E1757.md + E1926.md + E1766.md

pplucky:
Cover definition correction (to be consistent with auto-discovery when used).
Pairing instructions added.


Koenkk:
Thanks!

PR #169, opened 2019-11-27
Add per manufacturer info for reporting

sjorge:


docs/information/report.md
```diff
@@ -14,5 +14,20 @@ advanced:
 
 Zigbee2mqtt will discover devices which are eligible for reporting. The zigbee2mqtt logs shows for which devices report is configured.
 
+## Devices supporting report
+In order for this feature to work, the device itself has to support it.
+Below is an overview of devices supporting this feature per manufacturer.
+
+| Brand           | Support  |           |       |            | Reporting |           |       |            | Change Color While Off |
```
Koenkk: What is the difference between the values under 'Support' and 'Reporting'?

sjorge:
Support is what bulb support functions wise,

So for my 2 Hue RGB bulbs they support using color_temp or color_xy to set/get the color. But neither of those are reported using attReport.

sjorge:
Maybe renaming ‘support’ to ‘Get/Set’ ?

PS: I plan to update this once I get other bulbs/lamps, eyeing a trådfri ws panel atm.

Koenkk:
I don't understand, it doesn't make sense if you e.g. have a hue white bulb (brightness only). Wouldn't listing the what it reports already be enough?

sjorge:
Mostly to show the difference between what the bulbs supports and what it reports, it would make more sense if I had specified I tested the RGB bulbs form there offering vs WS or just fixed color bulbs...

sjorge:
So should I rename them to indicate I tested RGB bulbs and drop the first set of columns?

Koenkk:
RGB bulbs cover all features (on/off, brightness, color temp and color). When a Hue color bulb reports color temperature, a Hue color temperature bulb will very likely also do this (so let's assume that for now). The first column can indeed be dropped.

sjorge:
Done

Koenkk:
Thanks, last question: what does "Change color while off" mean?

sjorge:
For IKEA bulbs you can set the color while the bulb has a state of off, and you get a report back a few seconds later with the set color.

If you send a color change to a hue bulb nothing happens and it is ignore if the state is off.

This is important if you want to automate warm white -> cold white transitions based on sunset/sunrise.

Basically the IKEA bulbs still take color update and still report on the changes while the bulb state is off (but powered obviously) — I think they also take brightness so maybe this should be generalized as allows update/reports while off?

IKEA seems to be the only one of those tested so far that does this.

~ sjorge

Koenkk:
Maybe instead of adding it to the table, it should be added as a footnote like you did with the *1* in combination with the above explanation (otherwise it might be not clear what it is).

sjorge:
Done

Koenkk:
Thanks again :smile:

PR #170, opened 2019-11-30
Add pairing notes for Eurotronic SPZB0001

sjorge:
Both my test units required a reset before they would pair.

Both units shipped with firmware '20191014 22190930'


Koenkk:
Thanks!

PR #172, opened 2019-12-08
Devices are markdown files

nateww:
Also, converted all files from using DOS EOL (\r\n) to standard Unix EOL, since there was a mix of both and GIT properly handles the conversions when configured properly.


nateww:
Easier to read changes (ignores the EOL changes)

https://github.com/Koenkk/zigbee2mqtt.io/pull/172/files?w=1

Koenkk:
Thanks!

PR #174, opened 2019-12-18
Add OSRAM bulb to reporting support table

sjorge:
Another brand tested, sadly not one that will allow color changes while off like tradfri :(

But data is data!


sjorge:
Oops I made a mistake in the table, will close and open a new one as I am not near my ssh key atm

PR #175, opened 2019-12-19
Add OSRAM bulb to reporting support table

sjorge:
Corrected PR, with further testing it looks like the osram bulbs do not report color or color_temp changes :(

The colors are also a bit off when setting the color using rgb vs hue/sat


Koenkk:
Thanks again!

PR #177, opened 2019-12-20
Add hint about issues with follow-up pairings

andree182:
I had this issue with the control outlet and I think also the tradfri light controller... See https://zigbee2mqtt.discourse.group/t/cannot-pair-e1743-tradfri-on-off-switch/633/4 ...


Koenkk:
Could you create an issue for this? This shouldn't be needed.

PR #179, opened 2019-12-24
Added tamper for AC01353010G and iL07_1

jasperro:


Koenkk:
Thanks for this PR, but this will automatically be updated once the next version of zigbee2mqtt is released, therefore I'm closing this.

PR #183, opened 2019-12-29
Update binding.md (add info about target group)

Alestrix:
Added missing info about binding to a group. See comments in issue #2190 (https://github.com/Koenkk/zigbee2mqtt/issues/2190). According to the comments there, (un)binding to/from a group should be possible, though I was not successful yet.


Koenkk:
Thanks!

PR #184, opened 2019-12-29
Additional notes on this device

thetonto:
Have added reset instructions to the notes as was difficult to find information on how to do this without the manual.


Koenkk:
Thanks!

PR #186, opened 2019-12-31
Added more info for binding

nldroid:
Clearified the binding proces a bit.
Updated the link that is needed to edit the page.
Clearified the position of the led because i expected it to be at the back in the tiny hole there.


Koenkk:
Thanks!

PR #187, opened 2020-01-01
E1743: Added a note on binding

MrSnowflake:
Made it clear how to get binding to work with the IKEA E1743 switch.


Koenkk:
Thanks!

PR #192, opened 2020-01-04
Make alternative flashing methods clearer

jasperro:
Made it more clear that you can use a raspberry pi or arduino/esp8266 for flashing


Koenkk:
Thanks!

PR #193, opened 2020-01-04
Additional information about binding behaviour of this switch

grostim:


docs/devices/WXCJKG11LM.md
```diff
@@ -17,7 +17,19 @@ description: "Integrate your Xiaomi WXCJKG11LM via Zigbee2mqtt with whatever sma
 
 ## Notes
 
-None
+### Binding 
+By default the switch is bound with the Coordinator and command all the lights of the network.
```
Koenkk: 'command all the lights of the network.' do you mean that by default it control all lights?

grostim: > 'command all the lights of the network.' do you mean that by default it control all lights?

yes ! everything on the zigbee network that may undersand a generic on/off command !

Koenkk: thanks!

PR #194, opened 2020-01-06
Add information about the Paulmann 500.44 Dimmer

schwabe:
Also document the non-obvious reset procedure if I ever need to
do it again.


Koenkk:
Is this device already supported? I don't see it in https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L3922 nor do I see an open PR.

schwabe:
You merged the commit yesterday: https://github.com/Koenkk/zigbee-herdsman-converters/commit/5c4324624fe3dd68f64434ca6c49dbf46f0b01e9

Koenkk:
Sorry missed that, in that case https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/50045.md has to be edited and no new file should be created (the name of that page will automatically be updated on the next zigbee2mqtt release, just add the pairing stuff for now there).

schwabe:
How should file be called? Keep the 50045.md name or rename it to 50044_50045.md?

Koenkk:
Nope, in this case just add a note and mention the device type.

schwabe:
Like keep the two files but mention as a note that they share the device id with the other device?

Koenkk:
Keeping the two files is not possible, as they have the same `zigbeeModel` (note that the documentation is generated so it is always consistent with: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js). I would recommend adding this to the notes of https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/50045.md#notes

PR #197, opened 2020-01-07
Update pair guide for xiaomi sensor

rsimiciuc:


Koenkk:
This should be needed anymore with zigbee2mqtt 1.7 +

PR #198, opened 2020-01-07
Add note regarding pan_id to flashing instructions

SebiGo:
As discussed in https://zigbee2mqtt.discourse.group/t/unable-to-get-cc26x2r1-to-work/658. This note could potentially avoid issues.


Koenkk:
I've moved this to the supported adapter page as I think it fits better there, thanks!

PR #200, opened 2020-01-08
Update WV704R0A0902 to mention Clipsal branding

sjorge:
Update docs to also mention Clipsal (AU) branding in addition to Eberle (DE/AT) and Drayton (UK).

Also also call it 'Wiser Smart Thermostat' as in a few places as all brands refer to it as that.


Koenkk:
Thanks!

PR #201, opened 2020-01-10
Added group binding instructions

jasperro:
Many people had issues with binding IKEA remotes that only support binding to groups to lights. Added instructions on how to do this.


Koenkk:
Thanks!

PR #202, opened 2020-01-14
Added info about the two serial devices in the CC26X2R1

nldroid:
I'm not sure if the same applies to CC1352P-2


Koenkk:
Also applies to CC1352p, updated the pr, thanks!

PR #203, opened 2020-01-19
add pairing instruction to E1744 (Symfonisk remote)

CorePoint:
Manual says to press button for > 10 seconds, but i think that is for touchlink.


Koenkk:
Thanks!

PR #204, opened 2020-01-25
Fix doc rendering for Wiser radiator

sjorge:
- fix table about boost (not rendering properly)
- rename 'comission' to 'calibrate' to reflect the terminology used in the official manual
- mentioned extra twist to + to enter connect mode (as per manual)


Koenkk:
Thanks!

PR #208, opened 2020-02-04
About cc2538 adapter

kirovilya:


Koenkk:
Before making this an official supported adapter, the source code should be in https://github.com/koenkk/z-stack-firmware

kirovilya:
ok :) https://github.com/Koenkk/Z-Stack-firmware/pull/144

Koenkk:
good, I think the following things are missing before it can be used by the community:

- Instructions how to make the dongle
- Instructions how to flash the dongle

kirovilya:
These instructions are in issue and links there, like this https://modkam.ru/?p=1112 (but in russian).
there are also ready-made devices, links are exists in description

Koenkk:
Could it be translated to English? Otherwise it's not very useful for the zigbee2mqtt docs.

kirovilya:
I will try to translate. but now you can translate it with a browser too :) 
If you look at the issue, then some have already made them yourself

Gamester17:
Does the README.md for Z-Stack-firmware coordinator firmware also need to be updated for now that #144 have been merged adding support for CC2538 + CC2592 module?

https://github.com/Koenkk/Z-Stack-firmware/blob/master/coordinator/README.md

PR #212, opened 2020-02-06
add soldering note to CC1352P-2 section

CodeFinder2:
To use the CC1352P-2's external SMA antenna connector, a very tiny capacitor must be resoldered. This should be noted as it is not that easy and some people might select the CC1352P-2 (over the CC26X2R1) just for that antenna connector (not knowing that this requires soldering skills / additional work).

EDIT: see https://github.com/Koenkk/zigbee2mqtt/issues/2162#issuecomment-570286663


Koenkk:
Thanks!

PR #215, opened 2020-02-08
Adding an explainer for the `color_temp` property.

visnaut:
As I'm migrating from other platforms, which present color temperature in Kelvin, I was a little confounded by what unit the `color_temp` value is in, until I dipped into the source.

Adding some additional comments above it in the [MQTT Topics](https://www.zigbee2mqtt.io/information/mqtt_topics_and_message_structure.html#zigbee2mqttfriendly_nameset) page to hopefully assist others in making sense of it


Koenkk:
Thanks!

PR #216, opened 2020-02-12
Added debounce recommendation for some use cases

lukyjay:
This sensor spits multiple messages if used on the top of a horizontal sliding door. The magnet slides into place and will change the state whilst doing so. The debounce option allows the user to accurately receive the correct state.


Koenkk:
Thanks!

PR #217, opened 2020-02-12
Delete transition and duplicate binding text

lukyjay:
Binding was listed twice. Transition is irrelevant since this is a switch sensor, not a light. The light needs the transition setting on it.


Koenkk:
Thanks!

PR #219, opened 2020-02-15
Fix incorrect sensor config in documentation of Philips sensor.

t-liu93:
Signed-off-by: Tianyu Liu <info@jamesvillage.dev>
The extraction of json file of Philips sensor is not correct.
In the current version, the extracted illuminance is way too high (something like 20000-40000 during day time), the correct way of extraction should be illuminance_lux.


Koenkk:
Thanks!

PR #222, opened 2020-02-23
Add documentation step to "how to support new devices".

baderas:


Koenkk:
I've clarified a bit that this is an optional step, thanks!

PR #225, opened 2020-02-26
Fix docker doc

kovacsbalu:
Use network mode `host` as docker-compose

### DIFF
--- a/docs/information/docker.md
+++ b/docs/information/docker.md
@@ -22,7 +22,6 @@ docker run \
    -e TZ=Europe/Amsterdam \
    -v /run/udev:/run/udev:ro \
    --privileged=true \
-   --network host \
    koenkk/zigbee2mqtt
 ```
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Why is this needed?

kovacsbalu:
When you run mqtt broker on your host, not in this container.

kovacsbalu:
Check docker-compose example.

kovacsbalu:
Without this argument i got this logs:
```
zigbee2mqtt:info  2020-02-26 22:44:52: Connecting to MQTT server at mqtt://localhost
zigbee2mqtt:error 2020-02-26 22:45:02: Not connected to MQTT server!
zigbee2mqtt:error 2020-02-26 22:45:12: Not connected to MQTT server!
```
I run mqtt broker in my docker host, not in container. So localhost is not reached from container.

Koenkk:
This is only needed when the MQTT broker runs in a different Docker network than the Zigbee2mqtt container, updated the docs with this information. 

Thanks!

PR #230, opened 2020-03-01
Update home_assistant.md

Hellowlol:
This is useful to quickly find the device that was added without remembering the old devices and without digging in the logs.


Koenkk:
Thanks!

PR #233, opened 2020-03-05
fix yaml syntax

b-reich:
I fix the invalid YAML syntax. But I still recive this error "expected a dictionary for dictionary value @ data['packages']['sensor']"


Koenkk:
Thanks!

PR #236, opened 2020-03-08
Fixed wrong output variable

paulianttila:
Correct output value is attribute_and_json rather than json_and_attribute


Koenkk:
Thanks!

PR #237, opened 2020-03-08
Added a note to un-pair and re-pair after ota update

way2many:
Added a note to perform an OTA update, un-pair and re-pair before trying the bind commands as per DrRey comment https://github.com/Koenkk/zigbee2mqtt/issues/782#issuecomment-593781667


Koenkk:
This looks specific for the ICTC-G-1 so I think it should be added here instead: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/ICTC-G-1.md#notes

PR #239, opened 2020-03-09
Enrichment of HS1SA-* pairing doc

thomaschampagne:
- Add note to ensure device is awake during pairing
- Add a notice on **HS1SA-N**: supported too.


Koenkk:
Thanks!

PR #244, opened 2020-03-22
Update reporting support table

sjorge:
Update the on/off reporting for Hue, up to date bulbs don't report anything at all.


Koenkk:
Thanks!

PR #245, opened 2020-03-22
URL formatting fix

brd:
Just a small URL formatting fix.


Koenkk:
Thanks for the PR but this file is generated, please make the changes in https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L2310

brd:
Ahh, OK, will do.

PR #250, opened 2020-03-25
Update reporting support table for Müller Lights

sjorge:
I finally got a Müller Light Tint E27 to play with, very disappointing in both the red and green colors and reporting font :(


Koenkk:
Thanks!

PR #251, opened 2020-03-25
Update eWeLink Zigbee smart plug image

desmin88:
Current image for the eWeLink Zigbee smart plug seems to be a generic image.


Koenkk:
Thanks!

PR #254, opened 2020-04-01
add note about spaces being forbidden in FW files

CodeFinder2:
UniFlash is picky when there are spaces in the file name or the path and the flashing fails with something like "unable to open file".

See also https://github.com/Koenkk/zigbee-herdsman-converters/issues/715#issuecomment-607514386


Koenkk:
Thanks!

PR #257, opened 2020-04-10
Fix red light & add Debian

ChoiZ:
- Add debian packages (same as ubuntu)
- Add sentens for rebooting if we still with the red light


Koenkk:
Thanks!

PR #258, opened 2020-04-14
Add Wattle HCSMOKEDETECT as white label device

niksbiks:
Another white label version of the HEIMAN smoke detector.
Note that I could not get the ruby stuff running, so it is untested...


Koenkk:
The white label information of a page is generated from a whitelabel section in devices.js. Please add it here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L5541 example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L5820

PR #259, opened 2020-04-15
Updated Elko ESH 316 GLED RF device picture

wayller:


Koenkk:
Could you convert the file to .jpg and resize it to 150x150?

PR #262, opened 2020-04-16
extended instructions on uniflash

rmalchow:
installing ubuntu as described and flashing from there solved my uniflash issues.


Koenkk:
Thanks!

PR #264, opened 2020-04-17
Added outdoor temp. and time info

cptquad:
Added details on adding outdoor temperature display and getting the thermostat to display time


Koenkk:
Thanks!

PR #265, opened 2020-04-18
Improved IKEA motion sensor and Hue button images

ellnic:


ellnic:
Sorry, this is a mess. Git for web frustrates me. If you can cherry pick out 
#265 it contains improved images. Both images are 150x150, not sure why it appears to be failing?

PR #269, opened 2020-04-24
Make hint about possibly sleeping remote rather general, not group-sp…

emvau:
…ecific

Constantly got binding errors for a Hue remote and a Hue bulb (w/o using groups). After waking up the remote, it worked. So, in my view, the hint should not be group-specific.


Koenkk:
Thanks!

PR #270, opened 2020-04-26
Added ZG192910-4 to documentation

tomekpur:


tomekpur:
Closed since device has been added in different commit.

PR #280, opened 2020-05-06
Update 067771.md

jd1900:
Add all the configuration options that is available but only described in the source code.


Koenkk:
Thanks!

PR #281, opened 2020-05-06
Add pairing instruyctions for the GS361A-H04

sjorge:
Looks like lint is failing, but it was failing before this change too... I'm not sure how to fix.
I think a device was added incorrectly 3 commits ago?


Koenkk:
Thanks (don't worry about the lint, will fix this later)

PR #284, opened 2020-05-09
Added OpenHAB configuration

ruditb:
As it took me some time to figure out how to integrate the MiJia door & window contact sensor, I thought this could be useful for others too. So if you think that this is helpful I can add additional configurations to other devices as well which i have included in my OpenHab installation.


Koenkk:
Thanks!

PR #291, opened 2020-05-17
Add pin layouts

lukyjay:
The page talks about connecting pins but doesn't give clear pin layouts so I took it from your other page and included it here to help people.


Koenkk:
Thanks!

PR #298, opened 2020-05-20
Small typo in two files

wschoot:
Small typo in two files


Koenkk:
Thsnks!

wschoot:
Seems only fitting to reply with a typo in this case ;)

PR #301, opened 2020-05-20
supported adaptors: zzh requires rtscts false

sjorge:


Koenkk:
Thanks, I will edit this a bit further.

PR #302, opened 2020-05-20
fix missing alias in Home Assistant integration instructions

hmoffatt:
One of the automations in the HA integration instructions is missing an alias, which causes HA to give it a name like 'automation 13'.


Koenkk:
Thanks

PR #306, opened 2020-05-28
Add documentation to E1E-G7F

lukyjay:
The documentation is a bit light but the device is quite basic.

The device doesn't appear to have the ability to reset or pair other devices unlike the Phillips 324131092621.


Koenkk:
Thanks!

PR #307, opened 2020-05-28
Reorganise home page links

lukyjay:
I have found the homepage links to be a bit difficult to navigate at first. The link to running Zigbee2mqtt on bare-metal linux was difficult to see at first because my eyes jumped straight to the bullet points. The other categories appear to be overlapping so I have tried to simplify this.

All good if you don't agree with the changes, this is just a suggestion. Feel free to decline the pull request or amend. Thank you

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,10 +10,27 @@ Welcome to the Zigbee2mqtt documentation! Besides the documentation, you can als
 
 ### Getting started
 * [What do I need?](getting_started/what_do_i_need.md)
-* [Flashing the CC2531 USB stick](getting_started/flashing_the_cc2531.md)
 * [Running Zigbee2mqtt](getting_started/running_zigbee2mqtt.md)
-* [Pairing devices](getting_started/pairing_devices.md)
+
+### Devices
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -10,10 +10,27 @@ Welcome to the Zigbee2mqtt documentation! Besides the documentation, you can als
 
 ### Getting started
 * [What do I need?](getting_started/what_do_i_need.md)
-* [Flashing the CC2531 USB stick](getting_started/flashing_the_cc2531.md)
```
Koenkk, line 13: I think we should keep the flashing and pairing devices page in the getting started, as this section should follow what steps to execute on how to get zigbee2mqtt up and running. (https://www.zigbee2mqtt.io/getting_started/what_do_i_need.html#whats-next).

Also regarding how-tos I think it's nice to keep them together.

Is your problem solved by just moving the `Running zigbee2mqtt on..` below `Getting started`?

lukyjay: Sure! Also having the bare metal linux option bullet pointed so it's not overlooked :)

PR #309, opened 2020-06-01
Update MCCGQ01LM.md

ruditb:
OpenHAB has changed its behavior related to thing definition (OpenHAB 2.5.5-1 (Release Build)). The current version is now throwing a warning message even if it still works when using the current configuration syntax. So I changed the systax to get rid of the warning message and make the configuration more future proof. This syntax also works with the previous OpenHAB version. I also added a hint how to get the last change date time, because a lot of the original zigbee things hat this feature build in in OpenHAB.


Koenkk:
Thansk!

PR #310, opened 2020-06-05
Info about pairing Konke button

bessarabov:
Just bought a Konke button, it works fine with z2m, but the pairing instructions are missing.


Koenkk:
Thanks!

PR #313, opened 2020-06-07
Fixing number of HA ways to respond

bessarabov:
There was a commit that added a third method ( https://github.com/Koenkk/zigbee2mqtt.io/commit/0ecbb46b9d0f6a4eed209c790166a909665e4525 ) but the number wasn't changed.

(I"m a bit worried about this change. In the beginning of this file there is a text `NOTE 1: This file has been generated, do not edit this file manually!`, but I can't find the source how this file is generated )


Koenkk:
Thanks! I will update to generation part later

PR #322, opened 2020-06-16
Update to further clarify the difference seen by hass.io and docker users

zrsmithson:
The previous stating only gave a vague warning--vague enough that I accidentally ignored it and struggled to find and edit devices.js and homeassistant.js. By expanding this and noting how the "process is a little bit different" it will hopefully clarify it for future readers.

This is a minor change, but hopefully to avoid others not realizing the difference in installations between standalone, docker, and hass.io add-on.


Koenkk:
Thanks!

PR #332, opened 2020-07-03
Update HS1SA-M.md

sihui62:
HS1SA-E is the Zigbee 3.0 version of the device and is supported in the dev version of zigbee2mqtt (as of 2020-07-03)


Koenkk:
Thanks!

PR #338, opened 2020-07-08
Add note about USB3 ports and UNIFLASH

intrinseca:
Somewhat anecdotal, but both times I have tried to flash my CC1352 board it has failed until I switched to a USB 2 port. I thought this was worth noting here as a potential problem.


Koenkk:
Thanks!

PR #340, opened 2020-07-09
Create New zigbee line SNZB-03 Motion sensor

bizit524:
This is a place holder to update when the home assistant config has been discovered

### DIFF
--- a/docs/devices/SNZB-03.md
+++ b/docs/devices/SNZB-03.md
@@ -19,7 +19,7 @@ description: "Integrate your SONOFF SNZB-03 via Zigbee2mqtt with whatever smart 
 
 
 ### Pairing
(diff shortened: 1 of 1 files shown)


Koenkk:
Also meant for #341, #342 and #343 

I think all the devices are already supported but are currently marked as ewelink https://github.com/Koenkk/zigbee-herdsman-converters/blob/a0977d7b533036da594137ff65d7167ce289b054/devices.js#L9904

Have you already paired these?

bizit524:
Ahh sorry, I was just trying to fill in missing information, I thought by adding this it would give additional information and add it to supported devices. But to be fair i have never created pull requests before as I am not a dev. I was looking through supported devices and did not find them. I did pair snzb-03 (the motion sensor) and it shows up as a "MS01" which is the ewelink. I have not tested it yet but it does pair. 
Was trying to create a page like this - https://www.zigbee2mqtt.io/devices/BASICZBR3.html but for the other sonoff zigbees

Koenkk:
No problem, there was already some discussion in https://github.com/Koenkk/zigbee-herdsman-converters/pull/1378#discussion_r451114333 if we should mark these as sonoff instead of ewelink (as sonoff appear to be more popular). I believe we should so I will do this and these can be merged, thanks!

bizit524:
ah ok didn't know that there was already a thread, I will have to open up a seperate thread as i tested the motion sensor and door contact. they pair fine but never output anything :) have seen others working so not sure whats up :)

PR #344, opened 2020-07-11
Some MCCGQ11LM (Aqara door sensors) only work with a single short button press for pairing.

juanjux:
This did it for the 8 I just bought. The 5 button press didn't.


Koenkk:
Thanks!

PR #346, opened 2020-07-12
Fix minor typo in IKEA ICTC-G-1 docs

dale3h:
Noticed today that "legac" was missing the ending "y" in one occurrence in the IKEA ICTC-G-1 docs, so this is to address that.


Koenkk:
Thanks!

PR #348, opened 2020-07-16
Add Aqara D1 1 Gang smart switch (with neutral wire) [re-add] & Tuya Led Strip controller TS0503A

wolverinevn:
* Re-add Aqara D1 1 Gang smart switch (with neutral line)
  Remove .DS_Store
* Add Tuya Led Strip controller TS0503A


Koenkk:
Thanks!

PR #353, opened 2020-07-24
Add BTicino switches

maxcanna:
I added three variant of the existing BTicino/Legrand switch.

These belongs to the more common Livinglight serie:
- https://catalogo.bticino.it/BTI-L4003C-IT
- https://catalogo.bticino.it/BTI-N4003C-IT
- https://catalogo.bticino.it/BTI-NT4003C-IT

while the existing one belongs to Living Now serie: https://catalogo.bticino.it/BTI-K4003C-IT.

I provided images too.

There's no need to add support to them into z2m since they work out of the box


Koenkk:
> There's no need to add support to them into z2m since they work out of the box

How are these currently identified?

maxcanna:
> > There's no need to add support to them into z2m since they work out of the box
> 
> How are these currently identified?

_Light switch with neutral_ by _Legrand_ as the existing _K4003C_

Koenkk:
It is not possible to have pages here which are not in https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js (since the docs are automatically generated from that).

Therefore I would propose to:
- change https://github.com/Koenkk/zigbee-herdsman-converters/blob/e212c28110e0a4447f977a1afde1915f523b0aa8/devices.js#L10753 to `model: 'K4003C/L4003C/N4003C/NT4003C',`
- rename https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/images/devices/K4003C.jpg to `K4003C_L4003C_N4003C_NT4003C.jpg` and add all 4 devices to this picture.

maxcanna:
> It is not possible to have pages here which are not in https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js (since the docs are automatically generated from that).

Yes, I saw the warning in the CI. 

> Therefore I would propose to

I like it! I was thinking about doing something like this since those are just variants of the same device (in particular the color of the button for the last ones) so best thing is to have them together.

What about changing the name in the doc too? Otherwise it would look like newest ones are not supported and people like me can be confused

Koenkk:
> What about changing the name in the doc too? Otherwise it would look like newest ones are not supported and people like me can be confused

This will be done automatically, with each release the docs are automatically synced (file name, model numbers, descriptions, etc.) with zigbee-herdsman-converters `devices.js` (= single source of truth). Following the steps from https://github.com/Koenkk/zigbee2mqtt.io/pull/353#issuecomment-663732787 is enough.

maxcanna:
> This will be done automatically, with each release the docs are automatically synced (file name, model numbers, descriptions, etc.) with zigbee-herdsman-converters `devices.js` (= single source of truth). Following the steps from [#353 (comment)](https://github.com/Koenkk/zigbee2mqtt.io/pull/353#issuecomment-663732787) is enough.

Great! I did what you requested. Lmk

Koenkk:
Thanks!

PR #360, opened 2020-08-08
Update 9GED18000-009.md to add pin_code details

jhenkens:
Hi @Koenkk ,

I put the device updates on the device that we actually updated. The Kwikset lock you linked on the other PR is a 'conversion kit" that does not include an external keyboard. This "Weiser" Keypad lock is also sold as a Kwikset lock.

### DIFF
--- a/docs/devices/9GED18000-009.md
+++ b/docs/devices/9GED18000-009.md
@@ -17,8 +17,8 @@ description: "Integrate your Weiser 9GED18000-009 via Zigbee2mqtt with whatever 
 
 ## Notes
 
-### Pin Code Usage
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks, what other devices should also be updated with these docs?

jhenkens:
We haven't added the support to any other devices, since all the changes are part of a new converter in the `to/from Zigbee` endpoints. 

I'd imagine https://www.zigbee2mqtt.io/devices/9GED21500-005.html and https://www.zigbee2mqtt.io/devices/99140-002.html are compatible without changes, but there are a dozen or so keypad based locks in the supported devices list. Without having the devices in front of me, I'm not sure I'd feel comfortable making the change for any other deviecs other than the Kwikset/Weiser that we have updated so far.

Koenkk:
I agree, let's just add it for this device now. Thanks!

PR #374, opened 2020-09-02
Fix a typo in the CC2530 documentation

DominikPalo:
There is a typo (fimrware -> firmware) in the CC2530 documentation, so this PR just fixes it.


Koenkk:
Thanks!

PR #386, opened 2020-09-13
Update what_does_and_doesnt_require_repairing.md

JasperE84:
Had to re-pair all devices when changing to a 2530 to a zzh stick, didn't think I'd have to based on this page.


Koenkk:
this was an old page (also not accessible from the index.md) anymore, cleaned it up as it has been moved to: https://www.zigbee2mqtt.io/information/FAQ.html#what-does-and-does-not-require-repairing-of-all-devices

PR #390, opened 2020-09-16
Update ROB_200-008-0.md

VincentSC:
Clearer paring instructions


VincentSC:
Not sure if the whitelabel ID is correct, so I left it unchanged. See https://www.sunricher.com/zigbee-series/zigbee-remotes.html - it looks more like `SR-ZG9001K2-DIM`. The devices are mostly the same, only number of buttons are different, led-color is blue or green, and the older type tends to be more irritating with pairing (superquick clicking)

Koenkk:
Thanks!

PR #393, opened 2020-09-17
Update configuration.md

mf-social:
Homeassistant automatically pushes status messages to topic homeassistant/status, but the default configuration.yaml in the zigbee2mqtt docs suggest topic hass/status.

Homeassistant doc... 

https://www.home-assistant.io/docs/mqtt/birth_will/

This pr aligns to the zigbee2mqtt doc to match.


mf-social:
(I made this change via github on my phone so I don't have 'find and replace' capacity, might be worth a quick check through the rest of the repo to see if this also needs updating elsewhere)

PR #396, opened 2020-09-19
Add pairing instructions for TS0601 Thermostat

mhoyer:
Following the instructions in https://github.com/Koenkk/zigbee2mqtt/issues/3821#issuecomment-665960350 I was able to pair my own Moes Smart Valve/Thermostat (Tuya TS0601). Thus, I added the explanation here to help others as well.


Koenkk:
Thanks!

PR #399, opened 2020-09-20
Add a description of Zigbee Networking and the Zigbee Cluster Library, and how Zigbee2MQTT fits.

willu:
I spent a while figuring this out.  I thought it would be good to write it down and save others some time.  I'm not sure this is the right place for it though.

Also, some of the text comes across as rather harsh towards the end.  My goal here is that once scenes are working there'll be enough configuration that you can choose on a per-device basis what sort of zigbee network you want to run, basic zigbee binding between devices, or more advanced control through MQTT.

e.g. I'm thinking of putting smart switches throughout my house.  That's about 50 circuits.  But most of them don't need smart control - I just want to be able to say "turn off all the lights" and have everything turn off.  (Or poll individual lights if I really need to find out if they're on.)  Being able to stop the zigbee2MQTT controller binding to all 50 circuits and getting notified of every change (and getting the non-prompted updates that many devices send out) should reduce network traffic significantly.

### DIFF
--- a/docs/information/zigbee_network.md
+++ b/docs/information/zigbee_network.md
@@ -75,5 +75,4 @@ With this setup, when a switch is activated, it sends a message to the Zigbee2MQ
 
 This system is significantly more flexible than the base Zigbee system.  But it is also higher latency (it takes longer for the system to react to a switch being toggled) and it has more points of failure.  With the base Zigbee setup, not even the controller is involved once setup is complete.  With the Zigbee2MQTT setup there are two zigbee messages, two MQTT messages, and three extra processing steps (the main controller deciding what to do, and MQTT processing the messages in each direction).
 
-In addition, Zigbee2MQTT currently binds to each endpoint in the network.  Many of these endpoints will broadcast state information periodically to anything bound to them.  By binding to every device in the network, whether its state is relevant or not, zigbee2MQTT is causing a bunch of traffic on the zigbee network that would otherwise not be there.  This limits the practical size of the network.
-
+Note that in some commercial Zigbee systems, such as Phillips Hue, the controller node in the zigbee network is also the automation controller that can add additional smarts on top of the base Zigbee setup.  Zigbee2MQTT inserts MQTT between the two allowing them to be decoupled.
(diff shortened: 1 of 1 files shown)


docs/information/zigbee_network.md
```diff
@@ -29,3 +29,51 @@ Zigbee2MQTT logs the device type of your devices on startup, e.g.:
...
+
+This system is significantly more flexible than the base Zigbee system.  But it is also higher latency (it takes longer for the system to react to a switch being toggled) and it has more points of failure.  With the base Zigbee setup, not even the controller is involved once setup is complete.  With the Zigbee2MQTT setup there are two zigbee messages, two MQTT messages, and three extra processing steps (the main controller deciding what to do, and MQTT processing the messages in each direction).
+
+In addition, Zigbee2MQTT currently binds to each endpoint in the network.  Many of these endpoints will broadcast state information periodically to anything bound to them.  By binding to every device in the network, whether its state is relevant or not, zigbee2MQTT is causing a bunch of traffic on the zigbee network that would otherwise not be there.  This limits the practical size of the network.
```
Koenkk, line 78: - What do you mean with "binding" here?
- > zigbee2MQTT is causing a bunch of traffic on the zigbee network that would otherwise not be there
 - Could you explain this?

willu: I have some ML-ST-D200 switch adapters.  If I look in database.db I see: "binds":[{"cluster":6,"type":"endpoint","deviceIeeeAddress":"controller","endpointID":1},{"cluster":8,"type":"endpoint","deviceIeeeAddress":"controller","endpointID":1}]  (I've replaced the MAC address of my controller with the word controller to make it clearer).

This looks like the converters in zigbee-herdsman-converters are telling zigbee-herdsman to connect (bind) to various endpoints/clusters on the switch so they they can listen to state change messages.

Those devices are sending their state back to the controller periodically, even when their state isn't changing (i.e. not just when you've changed the state).  The period doesn't seem to be fixed - I think it backs off when the state hasn't changed for a while.  But just after a state change they're quite chatty.

I'm looking to get about 50 of these, and I'm a little concerned about network chatter.  I haven't tested it yet though.  My plan was to modify devices.yaml to add a "passive: true" setting which stopped zigbee-herdsman from making those bindings.

That was going to happen in my spare time though... so perhaps by 2022?

But we could take that section out of this text.

Koenkk: By default no bindings are made, are you using the  report feature? https://www.zigbee2mqtt.io/information/report.html

willu: No, I'm not using report.  Why don't we drop the last paragraph of this text.  I'll dig into where those binds are coming from.

Is there anything else you think should change?

willu: I added a different paragraph at the end comparing with Phillips Hue.

Koenkk: looks good, thanks!

PR #402, opened 2020-09-25
Make powerful adapters more visible in getting started

ciotlosm:
Supported adapters should be in getting started. People checking out what they need to start should understand the different types available before they buy a flashed CC2531 and find out it doesn't cover enough devices or range for them.


Koenkk:
There is already a note about this here: https://www.zigbee2mqtt.io/getting_started/what_do_i_need.html (note 1). I think we should make the newer adapter more prominent in the first place, also in what do I need.

ciotlosm:
I agree. The order of actions is now different with more supported adapters, so the flow for new users should have the adapter choice and some pros & cons as 1st step, with the "easy way" or "quick way" still pointing to a setup with CC2531.

Koenkk:
I've made some major changes. I will merge this already as I believe it's easier to review on https://www.zigbee2mqtt.io/ . Let me know in case you have any feedback.

PR #406, opened 2020-10-01
add pairing info for TS0044

blakadder:
add pairing info since the manual refers to first button which it is as an endpoint


Koenkk:
Thanks!

PR #412, opened 2020-10-07
Fix typo on nginx proxy configuration

jvgutierrez:
Tiny typo on the listen stanza for IPv6, it should read `ssl` instead of `sll`


Koenkk:
Thanks!

PR #416, opened 2020-10-12
Update notes section

Tche333:
Would be very useful when combined with a link form the object in the Zigbee2MQTT web page


Koenkk:
Thanks!

PR #422, opened 2020-10-13
Binding: Mention Numeric Form for Endpoints

felixstorm:
Mention that endpoints can also be specific in numeric form when binding endpoints (and not only in text form like `left` or `right`).


Koenkk:
Thanks!

PR #424, opened 2020-10-14
Update flashing_the_cc2531.md

sergeygrigorev:
It worked on Debian only after installing `libglib2.0-dev` as proposed [here](https://github.com/nfc-tools/libnfc/issues/373#issuecomment-280785351).


Koenkk:
Thanks!

PR #428, opened 2020-10-18
Run bundle update to get rid of Jekyll deprecation warnings.

ronaldevers:
Jekyll was throwing hundreds of instances of a single deprecation warning. This updates Jekyll to a fixed version.

The issue was reported on Jekyll here: https://github.com/jekyll/jekyll/issues/7947.

For me when running `bundle exec jekyll serve`, with the old version, I got tons of these:

    /usr/local/lib/ruby/gems/2.7.0/gems/jekyll-3.8.5/lib/jekyll/tags/include.rb:194: warning: Using the last argument as keyword parameters is deprecated


Koenkk:
Thanks!

PR #432, opened 2020-10-24
Update supported_devices to list best-on-top

cadavre:
Update supported_devices to highlight least-headache-generating devices on top. Previous order was favoring CC2531 which is extremely fakeable on Aliexpress and in most cases is good just for playground – not for production use.

This is extreme problem in tons of community forums. People are asking "Conbee II or CC2531" like the second is a keyword for Zigbee2MQTT. I guess, in the end, it makes a lot of headache and lots of people lost their trust to Z2M.

### DIFF
--- a/docs/information/supported_adapters.md
+++ b/docs/information/supported_adapters.md
@@ -2,66 +2,7 @@
 
 Zigbee2MQTT officially supports the following adapters:
 
-### Texas Instruments CC2531
-![CC2531](../images/cc2531.jpg)
-
-- USB connected Zigbee adapter with PCB antenna.
-- Cheap but not very powerful (+- $5), may not be powerfull enough for larger networks (20+ devices).
-- Limited range, ~30m line of sight
-- Requires additional hardware to be flashed (CC debugger and downloader cable)
-- Coordinator firmware: [Zigbee 1.2 (recommended)](https://github.com/Koenkk/Z-Stack-firmware/tree/master/coordinator/Z-Stack_Home_1.2/bin) and [Zigbee 3.0](https://github.com/Koenkk/Z-Stack-firmware/tree/master/coordinator/Z-Stack_3.0.x/bin)
-- Router firmware: [Zigbee 1.2](https://github.com/Koenkk/Z-Stack-firmware/tree/master/router/CC2531/bin), can be re-paired pressing the S2 button for 5 seconds.
-- Can be bought on [AliExpress](https://www.aliexpress.com/wholesale?catId=0&initiative_id=SB_20191108075039&SearchText=cc2531)
-- [How to flash with CC debugger](../information/flashing_the_cc2531.md) or [alternative flashing methods](./alternative_flashing_methods.md)
-
-### Electrolama zig-a-zig-ah! (zzh!)
-![zzh](../images/zzh.jpg)
-
-- USB connnected Zigbee adapter
-- **Very powerful**, will easily handle networks of 100+ devices.
-- Very good range (due to external antenna)
-- Can be bought on [Tindie](https://www.tindie.com/products/electrolama/zzh-cc2652r-multiprotocol-rf-stick/#product-reviews)
-- Coordinator firmware: [Zigbee 3.0](https://github.com/Koenkk/Z-Stack-firmware/tree/master/coordinator/Z-Stack_3.x.0/bin) (use **CC26X2R1_*.zip**)
-- [Flashing guide](https://electrolama.com/projects/zig-a-zig-ah/#flash-firmware) (requires no additional hardware to flash)
-- When migrating from another dongle (e.g. CC2531), make sure to modify your `pan_id` in your [configuration](configuration.md), otherwise Zigbee2MQTT won't start.
-
-### slaesh's CC2652RB stick
-![slaesh's CC2652RB stick](../images/slaeshs_cc2652rb_stick.jpg)
-
-- USB connnected Zigbee adapter
-- **Very powerful**, will easily handle networks of 100+ devices.
-- Very good range (due to external antenna, SMA female connector)
-- Can be bought on [Tindie](https://www.tindie.com/products/slaesh/cc2652-zigbee-coordinator-or-openthread-router/#product-reviews)
-- Coordinator firmware: [Zigbee 3.0](https://github.com/Koenkk/Z-Stack-firmware/tree/master/coordinator/Z-Stack_3.x.0/bin) (use **CC2652RB_*.zip**)
-- [Flashing guide](https://slae.sh/projects/cc2652/#flashing) (requires no additional hardware to flash)
-- When migrating from another dongle (e.g. CC2531), make sure to modify your `pan_id` in your [configuration](configuration.md), otherwise Zigbee2MQTT won't start.
-
-### Texas Instruments CC2530 (optionally with a CC2591 or CC2592 RF frontend)
-![CC2530](../images/cc2530.jpg)
-
-- Serial connected Zigbee adapter with most of the times and external atenna.
-- Cheap but not very powerful (+- $10), may not be powerfull enough for larger networks (20+ devices).
-- Good range, ~50-60m line of sight, sensitivity can be increased when used together with a CC2591 or CC2592
-- Requires additional hardware to be flashed (CC debugger)
-- Coordinator firmware: [Zigbee 1.2 (recommended)](https://github.com/Koenkk/Z-Stack-firmware/tree/master/coordinator/Z-Stack_Home_1.2/bin) and [Zigbee 3.0](https://github.com/Koenkk/Z-Stack-firmware/tree/master/coordinator/Z-Stack_3.0.x/bin)
-- Router firmware: [Zigbee 1.2](https://github.com/Koenkk/Z-Stack-firmware/tree/master/router/CC2530/bin), can be re-paired if you power on/power off the device three times (power on, wait 2 seconds, power off, repeat this cycle three times)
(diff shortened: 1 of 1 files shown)


andriej:
Great idea.
Also would be super to remove "powerful" from devices which does not have amplify in antenna-chip.
Signal strength is as important as chip in bigger networks and 6dBm of maximum transmit power vs 20dBm of i.e. Conbee makes huge difference.
Coordinator can make more stable connections with routers.

Koenkk:
My opinion about this:
- The slaesh and the zzh are the most recommended adapters for now due to it's external antenna, powerful chips and ability to flash without any additional hardware
- I don't recommend the launchpads; they use the same chips as the zzh and slaesh but are far more bulky, don't have an external antenna and are more expensive.
- I don't recommend the CC2538 because the firmware is not developed by Texas Instruments anymore, the chips on the zzh/slaesh appear to be more powerful while the price is similar and it requires a JTAG flasher to flash the firmware
- I don't think the CC2530 should be recommended as it requires some DIY

This list should be consistent with: https://www.zigbee2mqtt.io/getting_started/what_do_i_need.html#supported-zigbee-adapter

andriej:
Both zzh and slaesh use chip without amplifier, which is like setting WiFi AP to 6dBm instead of 20dBm allowed by regulations for 2.4GHz :/

Koenkk:
@andriej maybe @omerk or @slaesh can also comment here, but I'm not convinced an amplifier makes a big difference. In my practical test (https://docs.google.com/document/d/1MM6z-uBwQ-DkHnAA00rR2IauyC4c58JIHbatKahRdyw/edit?usp=sharing) the differences were very small.

andriej:
I have 3 floor (solid concrete in between) house and amplifier makes h_u_g_e difference.
Difference between CC2531 and CC2531+CC2591 is difference between no-stable link between 1 concrete supported floor vs stable link between bottom and top floor. 

I woulde've repeat test, but have thrown out all cc253x and only left 2 with amplifier as main routers in house.
After that my lack of signal disappeared.

Koenkk:
@andriej the CC2531 antenna is not good indeed (this is also what I found in my test).

omerk:
@Koenkk The answer to the question whether a PA makes a difference or not is that you do not always need a PA/LNA, as long as you follow a simple set of fundamentals. Building floorplan and materials, Zigbee network topology along with the levels of 2.4GHz spectrum utilisation all play a massive role though so there are always exceptions!

Placing your Zigbee coordinator in a sensible position away from a WiFi AP, choosing a Zigbee channel that is relatively free of interference from any other 2.4GHz signal (mainly WiFi) and having routers distributed around the area you are trying to cover, in most cases, is enough to get a healthy network with just the standard sticks we have.

If the issue is interference with a nearby WiFi AP then you *might* see some improvements with a higher power coordinator although I'd argue that remedial action such as moving the coordinator away from the AP (if possible) and changing channels used on both the WiFi AP and the Zigbee coordinator will probably work better. A busy WiFi AP (think streaming video, high speed downloads etc) that is on the same channel as the Zigbee coordinator will drown out the Zigbee traffic even if the coordinator has a PA/LNA. See [this article](https://www.metageek.com/training/resources/zigbee-wifi-coexistence.html) for WiFi and Zigbee channel utilisation.

If you are trying to cover distance (say, multiple floors in a building) or have a purely end device based network without routers then of course having a PA/LNA on your coordinator would help. But really, adding routers would probably be a better move for the overall stability of your network.

zzh has a removable antenna that is a good "all-rounder" but if you're working with a larger area, using a higher gain antenna can also help.

More power doesn't always mean better results. W.r.t to [this comment](https://github.com/Koenkk/zigbee2mqtt.io/pull/432#issuecomment-716007524), if you actually look at how enterprise WiFi networks are deployed, you don't always go full power on every AP around. Instead, you look at densities, coverage, spectrum utilisation and other factors and tune your APs accordingly. Same principles apply to large Zigbee networks as well, think about topology (and routers!) and spectrum utilisation before simply going higher power.

andriej:
Sure. But now, coming back to user groups related to (mostly HA) support - most people ten to buy stuff they see on website and moan.
I can barely count houses I know that someone buyin AP (not even Zigbee stuff) was aware of anything like interference.

And I know a lot of users who bought cc2531 and some Osram Plugs which didn't make anything stable. Guess what. I've starded with such setup. 4 OSRAM Plugs didn't manage to make any stable link between 1 floor (concrete one, with floor heating). Two osram plugs didn't make it. One conbee - did.

andriej:
Still to be done I guess?

cadavre:
Yes, I'll get back to this as soon as my professional work will allow me.

Hedda:
> @Koenkk The answer to the question whether a PA makes a difference or not is that you do not always need a PA/LNA, as long as you follow a simple set of fundamentals. Building floorplan and materials, Zigbee network topology along with the levels of 2.4GHz spectrum utilisation all play a massive role though so there are always exceptions!
> 
> Placing your Zigbee coordinator in a sensible position away from a WiFi AP, choosing a Zigbee channel that is relatively free of interference from any other 2.4GHz signal (mainly WiFi) and having routers distributed around the area you are trying to cover, in most cases, is enough to get a healthy network with just the standard sticks we have.
> 
> If the issue is interference with a nearby WiFi AP then you _might_ see some improvements with a higher power coordinator although I'd argue that remedial action such as moving the coordinator away from the AP (if possible) and changing channels used on both the WiFi AP and the Zigbee coordinator will probably work better. A busy WiFi AP (think streaming video, high speed downloads etc) that is on the same frequency as the Zigbee coordinator will drown out the Zigbee traffic even if the coordinator has a PA/LNA. See [this article](https://www.metageek.com/training/resources/zigbee-wifi-coexistence.html) for WiFi and Zigbee channel utilisation.
> 
> If you are trying to cover distance (say, multiple floors in a building) or have a purely end device based network without routers then of course having a PA/LNA on your coordinator would help. But really, adding routers would probably be a better move for the overall stability of your network.
> 
> zzh has a removable antenna that is a good "all-rounder" but if you're working with a larger area, using a higher gain antenna can also help.
> 
> More power doesn't always mean better results. W.r.t to [this comment](https://github.com/Koenkk/zigbee2mqtt.io/pull/432#issuecomment-716007524), if you actually look at how enterprise WiFi networks are deployed, you don't always go full power on every AP around. Instead, you look at densities, coverage, spectrum utilisation and other factors and tune your APs accordingly. Same principles apply to large Zigbee networks as well, think about topology (and routers!) and spectrum utilisation before simply going higher power.

Based on recommendations from @omerk above and several Zigbee threads in the open-source community I suggest adding something like this to FAQ troubleshooting section:

# General tips for extending Zigbee signal range

One of the more commonly asked questions is in regards to improving signal transmission and reception or resolving poor link quality. Fact is that bad signal reception can be due to many different things and experienced Zigbee users in the community will normally advise that there are a few simple fundamental things that you should do first to better optimize your Zigbee devices equipment that can resolve many basic problems related to signal strength. There are always exceptions however if possible to try follow at least some of these general recommendations before you take more advanced steps like reporting issues to developers and submitting debug logs. Just following these simple tips should help reduce signal interference and thus improve transmission and reception. 

1. Add more Zigbee routers between the far away devices and the next nearest router, distributing more Zigbee routers around areas with poorer reception. Zigbee network topology uses a "mesh network" design which means that each device that acts as a Zigbee router will extend the range of your Zigbee network. While there are exceptions, as a rule of thumb, almost all permanently powered devices will serve as a router. This is a part of Zigbee design concept, simple add more powered permanently Zigbee devices to get better coverage, (there are purpose-specific Zigbee routers which you can find by doing a community search for "Zigbee signal repeater" or "Zigbee range extender"). Devices that can not act as routers are known as "end devices" and you should know that some end device models from brands like Xiaomi/Aqara have issues connecting through routers.

2. Utilize a USB extension cable to move Zigbee coordinator adapter and place your Zigbee coordinator adapter in a sensible position, away from WiFi access-points/routers or other sources of 2.4GHz signals in a location that it is relatively free from strong signal interference. The best location will depend on your building's floorplan though normally you want to place it somewhere in the middle. Building materials are also very relative as example dense/thick concrete, bricks, masonry, etc. dampen will all wireless signals and therefore it is best to place your Zigbee coordinator adapter some distance from walls, ceilings and floors. You could also try different orientation of your Zigbee antenna (or your whole Zigbee coordinator adapter if it has an internal antenna). This is because the connection between the Zigbee coordinator radio and your other devices also depends on the way the antenna oriented in space.

3. Check your WiFi access-points/routers channels in use to see if your router/access point is overlapping the channel your Zigbee network is using. If so change the channel of the access-points/routers and lock it. It will normally be easier than changing the channel of your Zigbee network and then you will have to re-join/re-pair all of your Zigbee devices. A busy WiFi access-points/routers that are operating the same frequency range (overlapping channels) as your Zigbee coordinator will drown out the Zigbee traffic, and even more so if they are located close to each other. See this article for WiFi and Zigbee channel s overlapping frequency ranges -> https://www.metageek.com/training/resources/zigbee-wifi-coexistence.html

4. Update the firmware on your Zigbee coordinator adapter and your Zigbee devices. Note that depending on your hardware the latest Zigbee coordinator firmware might not always be the one recommended by the community so it is advised to ask before upgrading.

5. If you Zigbee coordinator adapter a removable antenna with an SMA-connector then you have the option of buying a high-gain antenna as an upgrade. In addition, you also have the option to use an antenna extension-cable if needed (though for various reasons that is not a good alternative to only using a USB extension cable with your Zigbee coordinator adapter if possible). This should really only be needed if you are trying to cover a long distance, like to another building or very dense/thick walls, ceilings and floors.

6. Buy more powerful Zigbee radio hardware with better radio range, preferably with an external antenna. If you are not only experimenting but want a permanently stable and healthy Zigbee network with potentially many devices then you should consider upgrading to a more powerful Zigbee coordinator USB adapter. Generally, those with an external antenna will have better range, therefore you will also want to avoid buying an internal adapter unless it has an external antenna.

Vismuthxxx:
Can i add a single-key and three-key ewelink switch? And only you have a two-key switch eWeLink ZB-SW02

PR #440, opened 2020-10-29
Improve Homeassistant integration

Bojan023:
This pull contains two commits related to controlling Z2M with Homeassistant:

1. Adding a slider to control the joining timer duration. Even if not used, it will behave like before (default 2 minutes)
2. Defining empty strings to text inputs to prevent possible confusion and improve usability


Koenkk:
Thanks!

PR #441, opened 2020-10-31
Hints about waking up the device when (un)binding

gitolicious:
As I stumbled upon this and only found hints in other Github issues, I think a quick note on the necessity of waking up the device are helpful for others.


Koenkk:
Thanks!

PR #450, opened 2020-11-10
Update 9290023349.md

u17194907425:
An update proposal related to https://github.com/Koenkk/zigbee2mqtt/issues/4902

### DIFF
--- a/docs/devices/9290023349.md
+++ b/docs/devices/9290023349.md
@@ -45,8 +45,9 @@ by pressing and holding the reset button on the bottom of the remote (next to th
 
 
 ### Power-on behavior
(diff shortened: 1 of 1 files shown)


u17194907425:
There is a comment about the bulb behavior when `hue_power_on_behavior` is set to `off`. I can't be certain it is similar for all the  other bulbs however for this specific model it is as described.

Might be an important note to highlight since it renders the bulb absolutely useless when the smart home system is down for some reason.

Koenkk:
Thanks! Assuming https://github.com/Koenkk/zigbee2mqtt/issues/4902 can be closed now.

PR #452, opened 2020-11-11
Added documentation for adding unsupported Tuya devices

drzony:
As promised in https://github.com/Koenkk/zigbee-herdsman-converters/pull/1742

### DIFF
--- a/docs/how_tos/how_to_support_new_devices.md
+++ b/docs/how_tos/how_to_support_new_devices.md
@@ -16,7 +16,7 @@ The first step is to pair the device with Zigbee2MQTT. It should be possible to 
 Once you successfully paired the device you will see something like:
 ```
 Zigbee2MQTT:info  2019-11-09T12:19:56: Successfully interviewed '0x00158d0001dc126a', device has successfully been paired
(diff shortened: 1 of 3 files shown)


docs/how_tos/how_to_support_new_tuya_devices.md
```diff
@@ -0,0 +1,157 @@
+---
+---
+# How to support new devices
```
Koenkk, line 3: Should be 'How to support new TuYa devices' I think

docs/how_tos/how_to_support_new_devices.md
```diff
@@ -28,7 +28,13 @@ The next step is the to add an entry of your device to `node_modules/zigbee-herd
 
 ```js
 {
-    zigbeeModel: ['lumi.sens'], // The model ID from: Device with modelID 'lumi.sens' is not supported.
```
Koenkk, line 31: The `zigbeeModel` is the preferred way of adding a device and is the easiest one. I think we should not mention fingerprint here but only in the how_to_support_new_tuya_devices

drzony: Fixed

docs/how_tos/how_to_support_new_tuya_devices.md
drzony: Fixed

Koenkk:
Thanks!

PR #455, opened 2020-11-13
Add Philips Hue Bloom 929002375901 and 929002376001

renini:
Added documentation for Philips Hue Bloom 929002375901 (white) and 929002376001 (black).


Koenkk:
Thanks!

PR #458, opened 2020-11-15
mixed up boolean

andreasbrett:
this one confused me a bit... turns out it was just a mixup of true and false. verified through other pages that referenced that "legacy = false" will DISABLE the legacy mode


Koenkk:
Thanks!

PR #464, opened 2020-11-24
Update LED1623G12.md

hdoedens:
Having trouble pairing a new bulb, I've updated the reset proces just a little. It seems that on new bulbs the reset sequence must be exactly 6 "on's". That is what caused me a lot of time to figure out as many people suggest to 'play safe' and use more than 6 "on's" in which case the bulb (maybe limited to newer bulbs) will start glowing, but does not reset!


Koenkk:
Thanks!

PR #465, opened 2020-11-24
Update SNZB-02.md

ahue:
Added a note on temperature change reporting only being issued for quite large changes or every hour while humidity changes are reported instantly, even if on 2nd digit level.


Koenkk:
Thanks!

PR #466, opened 2020-11-25
Added docs for Silvercrest (Lidl) HG06467: Smart LED string

MichielMag:


Koenkk:
I've updated the trigger effects text a bit (since its not device specific configuration but a device specific message). If you agree I will merge this.

MichielMag:
Yea, looks good!

PR #479, opened 2020-12-02
Filled in missing climate topics in SPZB0001.md

SoulOfNoob:
I noticed the topics are missing from the manual home-assistant climate config.
I tested this with my `SPZB0001` (new firmware) which doesn't work with the home-assistant default MQTT discovery.


Koenkk:
Thanks!

PR #483, opened 2020-12-03
Invalid json in example message

wafprotocol:
The example message for effects contains an extra comma at the end of the list of colors. Took me an embarrassing amount of time to figure out why it wasn't working.


Koenkk:
Thanks!

PR #489, opened 2020-12-07
Update HG06335.md

lxh7:
Better described what the "tamper" property means


Koenkk:
Thanks, moved this a bit since the text you edited is generated and cannot easily be changed.

lxh7:
Great! I saw discussions about the tamper-thing, and since I got the ConBee finally working with zigbee2mqtt and was still viewing the logs, testing the button was a matter of seconds 😊

Hopefully it works for others. 

 

And a thank you for the work and effort you put in zigbee2mqtt! What I’m missing is a diagram that shows what exactly it is, how the modules inside it work, how to change it (messages through the mqqt-side, it took me a couple of hours to discover that) etc.

 

Maybe when I get a better understanding I could draw up some diagrams for you?

 

Cheers!

Alex

PR #492, opened 2020-12-08
Update and rename HG06337.md to HG06337-CH.md

flejur:


Koenkk:
Different models have been unified to one named HG06337 so this is not needed.

PR #500, opened 2020-12-11
Improved instructions on enabling herdsman debugging

Rene-Sackers:
Maybe I'm just stupid, but the existing instructions for enabling herdsman debugging under HASS os/supervised were a little confusing to me. It tells you to add what looks like JSON (`"zigbee_herdsman_debug": true`) to `the configuration`, but the configuration you get when you have it as an add-on in hass is YAML.

Added a more specific instruction on how to enable it.


Koenkk:
Thanks!

PR #502, opened 2020-12-13
Correct climate systemMode from 'fan only' -> 'fan_only'

ownaginatious:
Follow-up related to [this](https://github.com/Koenkk/zigbee-herdsman-converters/pull/1896) PR.


Koenkk:
Thanks!

PR #505, opened 2020-12-16
Apex Smart Plug 16A

TheStigh:
Added support for APEX SMart Plug 16A switch from Datek AS.


Koenkk:
Can you also make a PR to add support for it in: https://github.com/koenkk/zigbee-herdsman-converters ?

TheStigh:
> Where can this plug be bought? (personally very interested in it)

Afaik it can only be bought at www.elkjop.no or in their stores, but tell me how many you want and I can mail them to you. I purchase them in the local store. I know they want to get an international market. It does also have capability of measuring external temperature using an external sensor (I haven't seen or know more of this yet). I will get more of their new devices which I will integrate in the best system in the world - Z2M !

PR #509, opened 2020-12-19
Update SEA801-Zigbee.md

A33a:
Hi, 

I found out that there is an undocumented value which is very helpful with this thermostat. It is called `local_temperature_calibration` and really helpful to control the device. For instance, the one I've got stops heating only after 3-4 degrees above the setpoint and I need to set the calibration minimum to 3 in order to achieve desired temperature in the room. 

Best, 
Azat


Koenkk:
Thanks!

PR #513, opened 2020-12-20
Added support for Namron 89665, and deleted ghost file ' SLT3.jpg'

TheStigh:
Deleted ' SLT3.jpg' because of invalid filename. Correct 'SLT3.jpg' already exist. Deleted file gave errors on local repositories as it started with a space.


Koenkk:
Thanks!

PR #516, opened 2020-12-21
Document serial to network bridge using tasmota

ayavilevich:
Latest version of Tasmota have a "Serial to TCP Bridge".
I found Tasmota to work better than espeasy in some cases. Good to have as an option for z2m.


Koenkk:
Thanks!

PR #522, opened 2020-12-24
Changed CO2 to CO for Carbon Monoxide sensors

rduivenvoorde:
In the docs I see for carbon monoxide something in the text docs like:
"Indicates if CO2 (carbon monoxide) is detected."
This should be CO (monoxide) compared do CO2 DIoxide

Grepped for 'CO2'  in all files.

For 2 sensors:
MCLH-08.md
TPZRCO2HT-Z3.md
I think these are actually CO2 sensors... but not 100% sure.


Koenkk:
Thanks!

PR #526, opened 2020-12-27
updating the actions for hue dimmer

lutierigb:
the release actions are called on-hold-release, for example, not on-release. updated all of them.


Koenkk:
Thanks!

PR #529, opened 2020-12-28
Recommend using docker to run npm ci

Skaronator:
Just noticed while creating the debug log for https://github.com/Koenkk/zigbee-herdsman-converters/issues/1985#issuecomment-751697805 that running npm ci on debian/linux isn't that straight forward. Installing NPM via apt results in a incompatible npm version with the installed node version. I could install it directly without the package manager but I don't really want npm on my server in the first place.

Since this part of the documentation is already only for docker users it makes sense to use a docker command here to run npm ci as well.

The upside is that we can make sure that the user has a up2date npm and node version.


Koenkk:
Thanks!

PR #531, opened 2020-12-29
Update ZYCT-202.md

timoline:
added binding section

### DIFF
--- a/docs/devices/ZYCT-202.md
+++ b/docs/devices/ZYCT-202.md
@@ -23,6 +23,10 @@ Factory reset the remote by holding the 0 (off) button for +-20 seconds.
 To establish a connection keep the remote within 2 meters from the hub.
 Press and hold the smart group button (button with two bulbs) and wait until the lights, below the channels, flash.
 
+### Binding
+Currently, binding to a device or group is not possible from zigbee2mqtt
+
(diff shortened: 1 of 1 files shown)


docs/devices/ZYCT-202.md
```diff
@@ -23,6 +23,10 @@ Factory reset the remote by holding the 0 (off) button for +-20 seconds.
 To establish a connection keep the remote within 2 meters from the hub.
 Press and hold the smart group button (button with two bulbs) and wait until the lights, below the channels, flash.
 
+### Binding
+Currently, binding to a device or group is not possible from zigbee2mqtt
```
Koenkk, line 27: I think this should be rephrased to something like: "This device does not support binding". Now it looks like it's a Zigbee2MQTT issue.

timoline: I understand...
but, technically zigbee2mqtt doesn't support it, you can bind with touchlink according the manual

Koenkk: Those should be the same bind commands as: https://www.zigbee2mqtt.io/information/binding.html

timoline: I'm sorry but binding dins't work...
I already sent the device back to the store, so i can't test anymore

PR #533, opened 2020-12-29
Remove actions that don't exist

andyshilton:
I have tested several of these and there is no action for releasing buttons.


Koenkk:
Thanks!

PR #534, opened 2020-12-29
Removing unsupported action

andyshilton:
Sorry @Koenkk  I think this may be the same. I don't have one to test, but given all the other battery powered Xiaomi devices don't have a "release" action, I would presume this one doesn't either.


Koenkk:
I expect it's indeed the same, thanks!

PR #537, opened 2020-12-31
Update GL-C-008-1ID.md

Zehelein:
Added missing line breaks that made markdown render especially the `Transition` headline as a sub-element of the list of colors.


Koenkk:
Thanks!

Zehelein:
Yw, and thanks for all your great work!

PR #538, opened 2021-01-01
Update RTCGQ11LM.md

flimmy2911:
Added another Troubleshooting section, the steps mentioned in the comments on issue 839 helped me fix the issue after days of trying other things to make the sensor pair again.


Koenkk:
Thanks but I think this should not be added to the device docs (since it can apply to every device). It's already mentioned in the FAQ: https://www.zigbee2mqtt.io/information/FAQ.html#no-logging-is-shown-at-all (last point)

flimmy2911:
Makes sense.
Now I just i wish i would have looked into the FAQ 3 days ago.

Btw: Thanks for the fast reply and the great project, happy new year to you.

PR #542, opened 2021-01-03
device: Calex smart wall switch image

svrooij:


Koenkk:
Can you rename the file to `421782.jpg`?

Koenkk:
> How did it generate that new device number? Is that something we can replicate?

It's the `model` defined in `devices.js`

> How can I add device specific information about this device? Because getting it to pair isn't described yet and it supports touchlink which is nice because that way you can control colors of some lights without the coordinator in between.

Duplicate a page of a similar device and fill in the desired info under `## Notes`. Don't spend to much time correcting the other information since that will be generated on the next z2m release.

svrooij:
Thought I needed the `zigbeeModel` name, will also add the docs page.
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2015/files#diff-5520dd49755e0a353d3857b8963a91cb4b8d87a01b6b8eb7bb6cce1c180eac83R9077

Koenkk:
Nope, the model is checked to be unique. Note that not all devices are identified through the `zigbeeModel` (as some devices share the same), e.g.  https://github.com/Koenkk/zigbee-herdsman-converters/blob/c9ed6bb75a9682cf60b41e4e7647c653480f9a6c/devices.js#L1096 and https://github.com/Koenkk/zigbee-herdsman-converters/blob/c9ed6bb75a9682cf60b41e4e7647c653480f9a6c/devices.js#L1120

svrooij:
There you go this time with additional documentation.

Koenkk:
Thanks!

svrooij:
@Koenkk Is it possible to do the generation of the documentation locally and preview the result? Maybe people are also interested in seeing which device are already supported in the dev version of zigbee2mqtt but not in latest. Then if people have a new device they See `Support added in version x` and have a choice to either wait for an update of zigbee2mqtt or to switch to dev. But they don't have to create a new issue for a device that is already in there.

And would it be possible to have a `zigbee2mqtt/experimental/update-herdsman` topic that would just install the latest version of herdsman and restart. Then you can still use the latest (tested version) but with new devices added sooner.

Koenkk:
@svrooij looking at devices.js is much easier then setting up the docs locally. 

This is mentioned here: https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html

> Before starting, first check if you devices is not already supported in the Zigbee2MQTT dev branch! This can done by checking devices.js. Search for example on the Zigbee model from step 1 below.

svrooij:
Hi @Koenkk 

This part seems to be missing from the device page.

https://github.com/Koenkk/zigbee2mqtt.io/pull/542/files#diff-a794cec2e3bb1b67bfbd81dc2119df1bcb197f5d82a5476598b3fde097a43e1cR20-R31

Koenkk:
@svrooij somehow missed this in my sync, re-added it, thanks.

PR #544, opened 2021-01-04
Docs update for hassos supervisor add-on org change

ciotlosm:


Koenkk:
Cool, let me know it's ready for merge.

Off topic question: do user need to do something to migrate from the danielwelch version?

ciotlosm:
We will keep danielwelch repo for now, but only with the configs for hassos not the rest of the code. The source of the images will be the new one, and only the new jobs will run for build. 

We will probably keep it like that for 6 months, keeping the versions in sync just by copying configs after a successful release, maybe with some delay to absorb some of the bugs we miss to allow people an easy rollback option. 

A release would be:
- Update config.json in dev on z2m org
- Publish a release tag and allow build
- Merge to master once build completes
- Copy config.json to danielwelch ( at the same time at start, and delayed by a week after 2nd successful release).

ciotlosm:
@Koenkk I think this and https://github.com/Koenkk/zigbee2mqtt/pull/5546#issuecomment-772914382 are ready for merging. I will update https://github.com/danielwelch/hassio-zigbee2mqtt after merge to clean it up to use https://hub.docker.com/orgs/zigbee2mqtt/repositories repositories and remove code from the repo and just leave configs.

Koenkk:
Thanks, really appreciate your efforts!

ciotlosm:
I should have renamed it when I wrote github support to break the fork history from the main one. I think it's a bit late now with quite a few people on the new repo. I'll think about it and find a solution, but you are correct, it's HassOS, Home Assistant Supervisor and Home Assistant Core as names, HassIO is no longer there.

PR #549, opened 2021-01-08
Update F20T60A.md

Giancky79:
Brand legrand but in Italy


docs/devices/F20T60A.md
```diff
@@ -9,7 +9,7 @@ description: "Integrate your BTicino F20T60A via Zigbee2MQTT with whatever smart
 
 # BTicino F20T60A
 
-| Model | F20T60A  |
+| Model | F20T60A (same Legrand 412015)  |
```
Koenkk, line 12: These docs are generated from the `model` in devices.js, which do not allow this. Probably better to add it to description.

PR #557, opened 2021-01-11
Update ZNCZ04LM.md

peca89:
Tested device myself. There seems to be mechanical switch inside a plug that detects inserted prongs. You can even attached a plug without a cable and "consumer_connected" it is still detected.


peca89:
The description in exposes.js would also need to be edited

https://github.com/Koenkk/zigbee-herdsman-converters/blob/6361070033828dd540d2107d90fd7d53b835f6ee/lib/exposes.js#L378

Koenkk:
Updated `exposes`, changed slightly to fit better with generation.

PR #558, opened 2021-01-12
ZB-SW03

castorw:
Adds docs and image for ZB-SW03 supported by https://github.com/Koenkk/zigbee-herdsman-converters/pull/2066.


Koenkk:
Thanks!

PR #560, opened 2021-01-12
update pairing and binding docs for E1766

asdftemp:
When trying to bind the IKEA E1766 remote to the blinds, I noticed some missing docs for the IKEA E1766. This pull request Should fix that and improve the write down of the pairing process. It's copied from the IKEA E1743 so kudo's to original writer.


Koenkk:
Thanks!

PR #562, opened 2021-01-13
Small fixes

tels7ar:
According to the manual and my own testing, you only need to hold down the buttons for 3 seconds to initiate pairing mode.


Koenkk:
Thanks!

PR #566, opened 2021-01-15
changed up-press, down-press and stop to new values

Mr-Groch:
I was suprised that my automation stopped working, and figured out in zigbee2mqtt logs that action was renamed...


Koenkk:
The listed actions are still correct, they only changed if you have `legacy: false` set.

Mr-Groch:
OK, then IMO it should be listed both values - for legacy false and true. I've updated PR

Koenkk:
Thanks!

PR #568, opened 2021-01-15
Added Schwaiger HAL300 E27 Light

marcr77:
I got a Schwaiger HAL300 E27 9,5W 806lm 1800-6500K + RGB
It describes itself as ZBT-RGBWLight-GLS0844

This is my first Change Request - I don't really know what to do - I hope it is correct

### DIFF
--- a/docs/devices/HAL300.md
+++ b/docs/devices/HAL300.md
@@ -1,0 +1,9 @@
+// Add HAL300 by Schwaiger to your devices.js
+// Icon could be like zbt_e27_rgbw.png
+    {
+        zigbeeModel: ['ZBT-RGBWLight-GLS0844'],
+        model: 'HAL300',
+        vendor: 'Schwaiger',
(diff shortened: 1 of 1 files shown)


marcr77:
I hope my request is correct this way

Koenkk:
Hello, looks like this merge request should be made against https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js , please let me know if you can do this, otherwise I can add it myself.

marcr77:
> Hello, looks like this merge request should be made against https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js , please let me know if you can do this, otherwise I can add it myself.

Hi,
I am not very firm in github things. Could you please add it. 
Best regards,
Marc

Koenkk:
Added!

device will be supported in the 1 February release of z2m

PR #569, opened 2021-01-16
Added missing details about exposed properties

joaodsantos:
After updating my zigbee2mqtt bridge from 1.13.0 to 1.17.0 I noticed that this plug is now exposing 4 new details that are not referenced in the documentation.
New properties are Voltage, Energy, Power and Current.
Added to both the "Exposes" section and also to the "Manual Configuration". Copied a lot of the text from a similar Xiaomi plug as it's exactly the same properties and types.
As a note, and this is probably a comment for the herdsman project, the HA integration is automatically creating a power_outage_memory sensor that the plug doesn't seem to support.
(copied from commit)

### DIFF
--- a/docs/devices/4000116784070.md
+++ b/docs/devices/4000116784070.md
@@ -19,6 +19,9 @@ description: "Integrate your Lonsonho 4000116784070 via Zigbee2MQTT with whateve
 
 None
 
+### Pairing
+Press and hold the power button on the device for +- 5 seconds (until a blue light starts blinking).
+After this the device will automatically join.
 
 ## Exposes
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Can you share the data/database.db entry of this device? Looking at the definition `4000116784070` doesn't support power measurements.

joaodsantos:
Is a screenshot of the MQTT entity enough? Happy to dig up more detail if needed.
You can also see what I meant by the power outage memory sensor being present but plug doesn't seem to actually support it.
lonsonho_tuya_plug

Koenkk:
> this plug is now exposing 4 new details that are not referenced in the documentation.
New properties are Voltage, Energy, Power and Current.

These are documented here: https://www.zigbee2mqtt.io/devices/TS0121_plug.html#exposes

What makes you think the plug does not support these? Looking at your screenshot e.g. power has a value

joaodsantos:
I think that might be where the confusion is, I didn't buy it as a Tuya UK plug. Bought it as a Lonsonho Tuya EU Plug as you can see from my order's description:
image

From the list of compatible devices that's the one that matches the brand, description and look of what I bought, hence why I thought this was the one. 
Since Lonsonho is a sub-brand of Tuya I wouldn't be surprised if the firmware is the same/very similar and that's why it shows up with the same name.

PR #572, opened 2021-01-17
Fixed a Typo in docs ("tripple" to "triple")

alexmerm:
confirmed that the output is indeed "triple" with my local instance Screenshot is of node red, reading it from home assistant through mosquitto mqtt, so it's not direct "proof" but like.. this is obviously a typo
image


Koenkk:
Thanks!

PR #574, opened 2021-01-18
Update TS0216.md

tarbax:
better explanation of triggering alarm


docs/devices/TS0216.md
```diff
@@ -31,7 +31,8 @@ The unit of this value is `%`.
 
 ### Alarm (binary)
 Value can be found in the published state on the `alarm` property.
-It's not possible to read (`/get`) or write (`/set`) this value.
+To read (`/get`) the value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/get` with payload `{"alarm": ""}`.
```
Koenkk, line 34: `/get` doesn't seem to be supported, only `/set`

PR #581, opened 2021-01-24
Update possible action payload values for IKEA E1744

flonix8:
I have just set up a brand new SYMFONISK sound controller (E1744) and it sends action events with the payloads according to my edit.
Maybe an older version behaves differently, but I feel like this might be a copy/paste issue, as brightness* actions don't make a ton of sense with this sound controller.


Koenkk:
I have to fix this in the docs, but these values will be send once you disable the legacy integration as described here: https://www.zigbee2mqtt.io/devices/E1744.html#legacy-integration

PR #583, opened 2021-01-26
Update E1524_E1810.md

martin-g-it:
Added color temperature workaround when binding feature is used


MattWestb:
Perhaps it can being good to mention its only IKEA devices that is working with the "magic grupe".
Then i writing devices is also include plugs and blinds (not tested but they is being setted up the same way if looking in sniffed data).

Koenkk:
Thanks!

PR #595, opened 2021-01-31
New button actions

denes44:
I just tested, this switch also supports double and hold actions.


Koenkk:
These actions were already added but docs were not updated yet (will be done with every release which I just did).

PR #596, opened 2021-02-01
Add voltage reading instructions for TS0201

sineverba:
Add configuration to read voltage for TS0201. Personally verified.


Koenkk:
Thanks!

PR #602, opened 2021-02-03
Add Heiman HM1RC-2-E Device

arnaud-gouder:
Copied all documentation from HS1RC-N, and replaced the name everywhere.


Koenkk:
Thanks!

PR #603, opened 2021-02-03
Update E1812.md by adding pairing instructions

erikshausen:
Addition of pairing instructions just like with E1524/E1810 (copied and modified a bit).


Koenkk:
Thanks!

PR #613, opened 2021-02-07
Clarify Home Assistant Birth/Will topic

shbatm:
After Koenkk/zigbee2mqtt#3918 was resolved, the docs were not completely updated to match the new Home Assistant birth/will topics that zigbee2mqtt was listening for.

A comment was added to the example code, but the example code still used `hass/status` instead of `homeassistant/status` which is what zigbee2mqtt was updated to listen to.

See comments made on Reddit here: https://www.reddit.com/r/homeassistant/comments/lel4u0/zigbee2mqtt_stopped_working_after_rebooting_home/


Koenkk:
Thanks!

PR #614, opened 2021-02-07
Add note about frontend and WebKit bug

davidjb:
Due to this long-standing  WebKit bug, the frontend can’t be used when basic auth is configured. Swapping to token based auth is a workaround, explained in this PR.


Koenkk:
Thanks!

PR #615, opened 2021-02-07
Rework formatting and content of network range how-to docs

davidjb:
Previously, this URL was rendered as text in the documentation (GitHub’s web UI worked but it uses a different markdown renderer)


nurikk:
better use proper markdown sytax

`[this is a link to google](http://google.com)`

[this is a link to google](http://google.com)

davidjb:
@nurikk This standard syntax is an autolink - https://daringfireball.net/projects/markdown/syntax#autolink - just to replace like-for-like so it renders correctly.  That said, summarising the title / content of the linked page and using that as the link's text would be more accessible - so I'm happy to adjust the PR if that's preferable.

Koenkk:
I like the suggestion from @nurikk , title: "Zigbee and WiFi coexistence" (?)

davidjb:
@Koenkk Updated and rebased on latest `master`. I reworded that sentence so it flowed a bit more and while I was there reviewed the rest of the doc for formatting, wording and grammar.

Koenkk:
Thanks!

PR #616, opened 2021-02-08
Clarify override config when changing exposed entity type

crowbarz:
When exposing a switch as a light, any device overrides need to be configured under light rather than switch, otherwise they are ignored. This PR corrects the example and adds a comment to note this.


Koenkk:
Thanks!

PR #627, opened 2021-02-14
OTA control explanation

flatsiedatsie:
- Set OTA check interval in minutes
- Disable automatic checking completely


Koenkk:
Thanks!

PR #629, opened 2021-02-16
Added Tradfri occupancy_timeout device config option

isorin:
Similar to Xiaomi Aqara motion sensor device option, added the ability to override the default detection cooldown timeout.
This allows custom motion detection loop control in HA automation by setting the value lower than the re-trigger interval of the device.

Related PR:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2251


docs/devices/E1525_E1745.md
```diff
@@ -30,6 +30,8 @@ The E1745 can be bound to groups using [binding](../information/binding).
 This device will pause detection of presence for approximately 1 min 37 sec after detection. In other words; with continuous detection, every 1:37 a new publish (occupancy: true) will be done.
 Cooldown of detection is 3 min. In other words; 3 min after last published detection (and trigger no longer present) occupancy will be published as false.
 
+The cooldown detection timeout can be overriden by using the following device config option:
+* `occupancy_timeout`: Timeout (in seconds) after which the `occupancy: false` message is sent.
```
Koenkk, line 34: It would be good to mention the minimal value. Does this e.g. work when setting occupancy_timeout to 5 seconds? 

e.g. when setting to 5:
- 0 seconds: motion, occupancy true is send
- 5 seconds: occupancy false send
- 7 seconds: motion detected, is occupancy true send now?

isorin: Yes, thank you for your comments and good point about the minimal value. Maybe it's worth mentioning that the device will re-trigger "occupancy: true" only when motion is detected after the device cooldown value (which is 90 seconds in my experience for these sensors).
So, in your example:
- 0 seconds: motion, occupancy true is sent
- 5 seconds: occupancy false sent
- 7 seconds: motion detected, nothing happens
- 95 seconds: motion detected, occupancy true is sent

For simple automations, setting a occupancy_timeout greater than the device re-trigger cooldown is probably desired and this setting allows to override the default fixed 3 minutes to whatever value people desire.

My use case for the motion sensors (I have both Xiaomi and Tradfri) is to set the occupancy_timeout just below the device re-trigger time (I use 58 seconds for Xiaomi and 1m28sec for Tradfri) and controlling the device cooldown / motion clear time through HA automations.
The ability to control this cycle (detection -> idle -> detection -> idle) in HA helps with a lot of situations, for instance, the ability to restart either HA or Zigbee2MQTT and being able to re-trigger and continue with the motion-controlled automations, preventing timeouts (the device triggers "occupancy true/false" every minute or so), switching from motion-control to timer-lights (by pressing a switch), etc.
I find this type of behaviour/control very powerful for advanced automations.

This behaviour already exist for Xiaomi Aqara motion sensors, I wanted to merely extend it to Tradfri motion sensors.

isorin: Should I add a note to say:
Please note that if you set the occupancy_timeout lower than the device trigger cooldown time then the device will only trigger "occupancy true" again for a motion event after the cooldown time has passed.

Koenkk: Yes please.

If I remember correctly this device has a knob on the back to tweak the detection time right? Does this influence the detection period?

isorin: The device has two knobs on the back. One is "night/day" (only trigger when it's dark), the other is "dimmed/bright" (sends brightness 30% for dimmed setting) - they don't influence the detection period.

I have updated the PR with the note. Hope it's fine now.

docs/devices/E1525_E1745.md
```diff
@@ -30,6 +30,12 @@ The E1745 can be bound to groups using [binding](../information/binding).
...
+
+Please note that setting `occupancy_timeout` lower than the device re-trigger time (around 1m30s) means `occupancy: true` will only be sent again when motion is detected after the re-trigger time.
```
isorin, line 36: Updated the PR with the note. Hope it's fine now.

Koenkk:
Thanks!

docs/devices/E1525_E1745.md
Koenkk: perfect!

PR #633, opened 2021-02-17
LED Indication info for Philips Hue Motion Sensor

uncle-fed:


Koenkk:
Thanks!

uncle-fed:
I am not sure if the same applies / works for the other Hue motion sensor (outside). 
I don't have that device.
https://www.zigbee2mqtt.io/devices/9290019758.html

Koenkk:
It does (I have this device). Docs will be automatically updated with this on the next release.

uncle-fed:
Does this mean that my change will be automatically copied to the other device?
Because I only modified one page, not both.

Koenkk:
Yes (these docs are generated so your text will slightly change)

PR #634, opened 2021-02-18
Feature/aldi bulbs

BMOD89:
Add documentation for Aldi bulbs (https://www.aldi-sued.de/de/p.lightway-smart-home-led-leuchtmittel.490000000000707059.html)


Koenkk:
Thanks!

P.s. no need to update descriptions manually here after changing it in devices.js
, since they will automatically be updated with next release.

PR #639, opened 2021-02-19
Update docs for SmartThings Multisense v4 (F-MLT-US-2) with accelerat…

declanshanaghy:
…ion properties


Koenkk:
Thanks! No need to update this after changing `devices.js`/`exposes.js` because this part is generated.

PR #640, opened 2021-02-21
Resetting bulb with TRADFRI Remote

JsBergbau:
Updated device description since getting 3rd generation remote control is probably possible with IKEA Tradfri remote control. This worked in my case see https://github.com/Koenkk/zigbee2mqtt/issues/296#issuecomment-782835434


Koenkk:
Thanks!

PR #645, opened 2021-02-23
Change picture for E1812, better picture without background

denes44:
I changed the picture for the E1812 shortcut button for another picture without background. The picture is from IKEA's website.


Koenkk:
Thanks!

PR #647, opened 2021-02-24
Update 014G2461.md

Jodrik:
Added steps for pairing the device (according to their own manual inside the gateway app)


Koenkk:
Thanks!

PR #648, opened 2021-02-26
Update 014G2461.md

Jodrik:
Added some more values that are exposed


docs/devices/014G2461.md
```diff
@@ -40,6 +44,9 @@ It's not possible to read (`/get`) or write (`/set`) this value.
 The minimal value is `0` and the maximum value is `100`.
 The unit of this value is `%`.
 
+### Mount mode (boolean)
```
Koenkk, line 47: On which property is this exposes? I cannot find anything about e.g. `mount_mode`.

Koenkk: Nevermind, it's `mounted_mode`.

PR #654, opened 2021-03-03
Check spelling and normalise case in notes

davidjb:
This PR reviews and fixes spelling in the device page notes file, as well as normalising the case of Zigbee, MQTT and Zigbee2MQTT.  There's one case where sensors are named with the typo `ambiant` in the name -- I've adjusted the heading and text but not the underlying sensors as they'd need separate changes in herdsman-converts but also deprecation etc to change the identifiers.


Koenkk:
Thanks!

PR #656, opened 2021-03-05
Update QBKG23LM.md

WarmCatUK:
Unlike the 2 gang variety, the Aqara D1 1 gang with neutral, unfortunately does not publish 'double' and 'triple' actions. Amended to reflect.


Koenkk:
Thanks!

PR #659, opened 2021-03-06
Remove float support from AC03647 docs

ai-xyz:
This commit removes the non-existing float support from the AC03647 according to [#6574](https://github.com/Koenkk/zigbee2mqtt/issues/6574)


Koenkk:
Thanks!

PR #660, opened 2021-03-08
Remove icon property

nevries:
Looks like in current ha version, devices.X.homeassistant.light.icon throws a 'unknown property' error when ha reads the entity config from mqtt.


Koenkk:
Thanks!

PR #663, opened 2021-03-09
Add documentation for Eurotronic specific attributes

t0mas:
Add documentation on Eurotronic specific eurotronic_trv_mode and eurotronic_valve_position attributes as supported by Zigbee-herdsman. Based on page 14 of the user manual of the device: https://eurotronic.org/wp-content/uploads/2021/02/Spirit_ZigBee_BDA_web_EN_2021.pdf

### DIFF
--- a/docs/integration/home_assistant.md
+++ b/docs/integration/home_assistant.md
@@ -162,7 +162,6 @@ devices:
         object_id: light
       light:
         name: my_switch
-        icon: mdi:desk-lamp
       # OR if your devices has multiple endpoints (e.g. left/right)
       switch_left:
         type: light


Koenkk:
Thanks! This part of the docs is generated based on the `exposes`: https://github.com/Koenkk/zigbee-herdsman-converters/blob/7df420616f4aca9ace70879189cfab5231527eb0/devices.js#L10880.

Could you make a pr to update that instead?

e.g. for `eurotronic_trv_mode` you can use: 

```js
exposes.enum('eurotronic_trv_mode', exposes.access.STATE_SET, [1, 2]).withDescription('Select between direct control of the valve via the `eurotronic_valve_position` or automatic control of the valve based on the `current_heating_setpoint`. For manual control set the value to 1, for automatic control set the value to 2 (the default). When switched to manual mode the display will show a value from 0 (valve closed) to 100 (valve fully open) and the buttons on the device are disabled.')
```

t0mas:
Apologies, didn't see this part was also generated. I'll add it to devices.js and the rest to docgen/device_page_notes.js and create a new pull request.

PR #665, opened 2021-03-10
Update running_zigbee2mqtt.md

pixeldoc2000:
Note added about Docker mount command does not support certain symbols like colon.


Koenkk:
Thanks!

PR #666, opened 2021-03-11
Make user extension example actually work

kristianheljas:
Currently, the example extension in the docs does not work because:

1. Extensions are actually loaded from `data/extension` directory (https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/externalExtension.js#L17:L17)
2. MQTT connection is not available in constructor


kristianheljas:
Oh, the relative require work only when working inside `zigbee2mqtt` source directory.

So I removed the relative require and provided full example just like in https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/extension.js

Koenkk:
Thanks!

PR #675, opened 2021-03-22
Explanation what reportable_change means

ai-xyz:
Hi,

i had to lookup a bit the code of the zigbee-herdsman-converters in order to ensure that 100 in the reportable_change for a reporting configuration stands for 1°C. The Configure Reporting Command in the ZigBee Cluster Library is also not noting what reportable_change really means. From the frontend you don't know that the device is sending in temperature*100°C steps formatted as integer, hence it would be nice to note somewhere that the reportable_change is the device's reported raw value. I am not sure if my explanation is good enough but maybe it can help one or the other.


Koenkk:
Thanks!

PR #676, opened 2021-03-24
Information for flashing cc2531 extended.

glad-wolf:
On newer Ubuntu versions cc-tool can simply be installed via apt (no need to clone and  build cc-tool)


Koenkk:
Thanks!

PR #679, opened 2021-03-26
Add image for Linkind key fob

Tediore:
Pairing instructions: hold the upper right (arm home) and lower left (disarm) buttons until the green LED at the top blinks rapidly, then release. Key fob should now be in pairing mode.

Related to https://github.com/Koenkk/zigbee-herdsman-converters/pull/2416


Koenkk:
Thanks!

PR #682, opened 2021-03-27
Added option for not logging into syslog

JsBergbau:
I had the same problem Zigbee2MQTT filling my syslog. Since Zigbee2MQTT logs also in /opt/zigbee2mqtt/data/log/ you can avoid this redudancy. Especially on Raspberry PI with SD Card this also saves SD card lifetime. 
For original issue see https://github.com/Koenkk/zigbee2mqtt/issues/288


JsBergbau:
sorry wrong formatting. Will try it again

PR #685, opened 2021-03-27
Add documentation on Iluminize 5120.1210

deltaphi:
See https://github.com/Koenkk/zigbee-herdsman-converters/pull/2421


deltaphi:
Actually not deeded, as the documentation claims that this page will be auto-generated on the next release.

PR #688, opened 2021-03-30
Update how_to_connect_to_a_remote_adapter.md

Hedda:
Update how_to_connect_to_a_remote_adapter.md with a recommendation not to use WiFi-based Serial-to-IP bridges (serial proxy servers) with Silicon Labs EZSP adapters.

https://github.com/zigpy/bellows#warning-about-zigbee-to-wifi-bridges

### DIFF
--- a/docs/how_tos/how_to_connect_to_a_remote_adapter.md
+++ b/docs/how_tos/how_to_connect_to_a_remote_adapter.md
@@ -3,7 +3,9 @@ This how-to explains how to run Zigbee2MQTT with an adapter on a remote location
 We will use ser2net for this which allows to connect to a serial port over TCP.
 In this way you can e.g. setup a Raspberry Pi Zero with the adapter connected while running Zigbee2MQTT on a different system. The instructions below have to be executed on the system where the adapter is connected to.
 
-DeCONZ Conbee II is not supported over tcp connections using this or any other configuration at this time. 
+DeCONZ Conbee II is not supported over TCO connections using this or any other configuration at this time. 
+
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Hedda:
> Thanks! Does this mean the sonoff zigbee bridge is not recommended?

Correct, at least WiFi based should not generally be recommended to everyone who thinks that stability is a requirement, like in your "production" environment at home with a lot of devices that needs always be working and be available to achieve high WAF.

But there are of course exception WiFi environments where and when it works fine, as it will still work without any problems if you have extremely stable WiFi at home, (for example such WiFi environments as offered by Ubiquiti UniFi with several access points for redundancy and have features rolling upgrades for no downtime).

FYI, I posted a deeper technical explaination and more detailed arguments here -> https://github.com/home-assistant/home-assistant.io/issues/17170

PS: Have also submitted a similar PR for Home Assistant's ZHA integration here -> https://github.com/home-assistant/home-assistant.io/pull/17179

Hedda:
Tip! If still need a network-attached remote EZSP adapter then recommend to instead use a (wired) Ethernet-based solution:

Example:

https://github.com/tube0013/tube_gateways

and

https://github.com/zigpy/zigpy/discussions/584

Ethernet-based (wired) non-WiFi solution of course also include using ser2net on a single-board-computer like a Raspberry Pi:

https://www.zigbee2mqtt.io/how_tos/how_to_connect_to_a_remote_adapter.html

PR #697, opened 2021-04-05
Update FAQ.md

sibero80:
Proposing changes for alternate method of permanently giving users permissions over device.

I came to this method when used Node.js installation method, faced permission issues accessing device and tried the suggested method in the FAQ with no success.

Found this https://github.com/esp8266/source-code-examples/issues/26 forum that provided the following approach:

Add user to the `uucp ` ,  `tty `  ,  `dialout `   groups:
```
sudo usermod -a -G uucp $USER
sudo usermod -a -G tty $USER
sudo usermod -a -G dialout $USER
```
The particular group that would make it work depends on the type of serial connection and linux distribution.

I think this would be a more simple solution but just wanted to bring some light to this with the hopes that someone more knowledgeable could provide a more educated opinion.

Cheers!


Koenkk:
Thanks!

PR #698, opened 2021-04-06
Update how_to_connect_to_a_remote_adapter.md

zombielinux:
Added specific instructions for using a Slaesh stick with ser2net, previously not working. Solution found and verified:

https://github.com/cminyard/ser2net/issues/46


Koenkk:
Thanks!

PR #700, opened 2021-04-07
Add some explanation to actions and pairing

afoeder:
Note that the pairing information is the
same as on [1] but it's the same author
anyways.

[1] https://github.com/blakadder/zigbee/commit/ddda552661e456f13bb0c9de01d3d93a43694c86


Koenkk:
Thanks, moved the action explanation to a different section since the section you edited is automagically generated.

PR #701, opened 2021-04-08
Added one-liner as hint

tillz:
As I didn't want to convert the numbers manually (like stated in the tutorial), I wrote a minimal python one-liner.
I just want to share this for others who are too lazy to convert 16 decimal numbers to hex.
Feel free to reject the PR if you don't want such experimental snippets in the docs.


Koenkk:
Thanks!

PR #703, opened 2021-04-12
Update supported_adapters.md with a note about EZSP and WiFi

Hedda:
Update supported_adapters.md with a note about EZSP and WiFi


Koenkk:
Thanks!

PR #705, opened 2021-04-16
Add 57600 as another common baud rate speed example

Hedda:
Add 57600 as another common baud rate speed example to match PR https://github.com/Koenkk/zigbee2mqtt/pull/7094 which in turn ain to confusions about EZSP baud speeds like in https://github.com/Koenkk/zigbee-herdsman/issues/319


Koenkk:
Thanks!

PR #709, opened 2021-04-18
Update docker.md

Moltey:
I propose to change the Docker documentation to provide insights on how to increase security by running the zigbee2mqtt container as a non-root user. The executing user needs to be added to the dialout group (or the corresponding group in the respective distro) with the command "--group-add dialout".

### DIFF
--- a/docs/information/docker.md
+++ b/docs/information/docker.md
@@ -11,11 +11,17 @@ Before executing `docker run` pull the correct image with `docker pull koenkk/zi
 
 First run the container, this will create the configuration directory. Change `configuration.yaml` according to your situation and start again.
 
-## Running
+## Running as root
 Run by executing the following commmand:
 
+1. Identify your device:
 ```bash
-docker run \
+$ ls -l /dev/serial/by-id
+```
+
+2. Run docker, whereas the "--device" statement should match the previous output:
+```bash
+$ docker run \
    -it \
    -v $(pwd)/data:/app/data \
    --device=/dev/ttyACM0 \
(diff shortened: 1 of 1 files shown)


docs/information/docker.md
```diff
@@ -30,12 +36,44 @@ docker run \
 * `--device=/dev/ttyACM0`: Location of adapter (e.g. CC2531)
 * `-v /run/udev:/run/udev:ro --privileged=true`: is optional, only required for autodetecting the port
 * Optional: in case your MQTT broker is running on `localhost` and is not within the same Docker network as the Zigbee2MQTT container also add `--network host \`.
+* Note: Running in "--privileged" mode is a security risk.
```
Koenkk, line 39: Is this note really needed? Running as root is already a security risk, this might only confuse

Moltey:
```suggestion
* Note: Running a docker container as root, especially in "--privileged" mode, is a security risk.
```

Moltey: I agree, I am not insisting on that line but maybe the root part should also be mentioned ;)

Koenkk: What about removing the whole root section?

docs/information/docker.md
```diff
@@ -30,12 +36,44 @@ docker run \
...
+```
+$ ls -l /dev/tty*
+```
+
+3. Check the user&group id you want to execute the docker image with:
+```
+$ id
+```
+
+4. Start the docker container (note: interface, user&group ID must match the outputs above, priveleged mode is not required):
```
Koenkk, line 58: It's not necessary to mention things that are not required, I suggest to remove `priveleged mode is not required`

docs/information/docker.md
Moltey: If you mean the line: 
Fine for me.

If you mean the root paragraph: 
The root mode could be beneficial for auto-discovery mode for USB devices. However, I never tried auto-discovery, therefore, I cannot tell, how that feature works.

Koenkk: I see, lets remove the line but also add a note that discovery only works when running as root (since it requires --privilliged).

dryros:
The integration in HA no longer works well in version 1.18 of zigbee2mqtt, also this device would need a switch entity to be able to trigger it from the automation panel, otherwise it is impossible, please check it

Koenkk:
Thanks!

PR #715, opened 2021-04-24
Add binding issue for Busch-Jae 6737

afoeder:
This adds an explanation and a reference
to the discussion about the binding issue
of the four-gang variant 6737's bottom row.

Preview of the rendering:
image

Footnote:
image

Related: #7009


Koenkk:
Thanks!

PR #724, opened 2021-05-03
Update AC0251100NJ_AC0251700NJ.md

unverbraucht:
Add information on endpoints for pairing, caused me some headaches

### DIFF
--- a/docs/devices/AC0251100NJ_AC0251700NJ.md
+++ b/docs/devices/AC0251100NJ_AC0251700NJ.md
@@ -28,11 +28,18 @@ If the Switch is connected hold Middle and Arrow Up Buttons for 3 Seconds to dis
 
 * `legacy`: Set to `false` to disable the legacy integration (highly recommended!) (default: true)
 
+### Groups and pairing
+By default this switch sends commands to group 0, which is the default group. For me it meant that it switched all lights in my Zigbee network! Once you pair it with another group or device it stops sending to the default group, which probably means that it can only be paired to one partner.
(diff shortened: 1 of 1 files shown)


docs/devices/AC0251100NJ_AC0251700NJ.md
```diff
@@ -28,11 +28,18 @@ If the Switch is connected hold Middle and Arrow Up Buttons for 3 Seconds to dis
 
 * `legacy`: Set to `false` to disable the legacy integration (highly recommended!) (default: true)
 
+### Groups and pairing
```
Koenkk, line 31: This should already be done out-of-the-box: https://github.com/Koenkk/zigbee-herdsman-converters/blob/98f1722c90f1a6963b21f99a98797ea2fd484c34/devices/osram.js#L321

Can you try pairing again, do you see it is "Successfully configured.."  in the log?

unverbraucht: Haven't gotten around to it yet, I'll try it out later this week.

Reading the code at that point it seems that it should bind reporting to all three endpoints - wouldn't reporting only concern that Z2M can update the group state? My issue was that sending commands from the middle or lower button by default would switch all lights in the network because only endpoint 1 was bound to the group, endpoints 2 & 3 were still bound to group 0.

Koenkk: First check if the code is wrong. My point here is that the configure should setup the device correctly which means not sending commands to group 0. (basically what you do manually here should be done there)

unverbraucht: Sorry for the long wait, I finally got around to re-pairing it. After completely resetting it and removing it the switch did pair cleanly and did not join group 0. I did see "Successfully configured ..." in the log. Don't know what happened the first time.
Thanks for the support, this PR can be rejected.

PR #728, opened 2021-05-04
Add Note for Hue reset w/dimmer switch

afoeder:
This adds a direction to the Notes section of every Philips device
on how to factory-reset the bulbs using the Dimmer Switch.
In a second commit it contains the rendering of this addition.


Koenkk:
Thanks!

PR #734, opened 2021-05-06
Added documentation for Philips Hue Iris Gold Limited edition

gusi1994:
I added the support page for the Philips Hue Iris Gen 4 Gold Limited edition. Based on the model 929002376801 i copied the documentation. It seems that these models are the same but only with different model identifiers.

The model integration on the converters repo is [here](https://github.com/Koenkk/zigbee-herdsman-converters/pull/2554)

Tell me if I can improve on that.


Koenkk:
Thanks!

thomasvoni:
Can you also add the Support for the Philips Hue Iris Gen 4 Silver Limited edition. Based on the model 929002376701?

gusi1994:
Sure, tomorrow I will push an update. I think that the silver lamp should have the same config.

PR #739, opened 2021-05-07
Added support document for philips hue iris silver edition

gusi1994:
As requested by @thomasvoni in #734 I added the same data for the silver philips hue iris limited edition.


Koenkk:
Thanks!

PR #747, opened 2021-05-13
Correct HS3SA image

etatto:
This is a PR to fix the current HS3SA image in the documentation which is not the correct one.


Koenkk:
Thanks!

PR #748, opened 2021-05-13
Remove unwanted </a> tags

MariusRumpf:
Removes unwanted </a> tags from text content of newly added supported adapters description text.


Koenkk:
Thanks!

PR #752, opened 2021-05-16
Adds documentation for SMaBiT AV2010/21C

cdemi:
Further to Koenkk/zigbee2mqtt/issues/7485 this is the documentation for the SMaBiT AV2010/21C. This is my first pull request in this repository so if you need different changes or I am doing something incorrectly just tell me and I will fix.

In any case thanks for the Zigbee2mqtt project I use it daily


cdemi:
Upon re-reading the documentation it states that the device page will automatically be generated, so maybe this pull request isn't necessary, but in that case, where will it get the image from?

Koenkk:
Only the markdown will be generated picture not, thanks!

PR #763, opened 2021-05-23
FAQ.md: Mention a hacky method of "fixing" the non-responsive CC2531 …

konstantint:
…dongle

Spent quite a while looking for a way to make my freshly flashed dongle work, tried everything on this page and elsewhere until finding a working solution in [this obscure comment](https://www.gitmemory.com/issue/Koenkk/zigbee2mqtt/2997/605655045). Plugging the dongle into a laptop and then back actually worked, and I'm apparently not the only one with this issue, so it this method is for sure worth mentioning here (it would be especially nice if someone could explain what is actually happening, but the advice is valuable even without this information).

### DIFF
--- a/docs/information/FAQ.md
+++ b/docs/information/FAQ.md
@@ -89,6 +89,7 @@ Most of the times this is caused by zigbee-herdsman not being able to communicat
 2 common reasons of this error:
 1. The port of your serial adapter changed. Check [this](../getting_started/running_zigbee2mqtt.md#1-determine-location-of-the-adapter-and-checking-user-permissions) to find out the port of your adapter.
 2. If you are using a CC2530 or CC2531; it is a common issue for this adapter to crash (due to its outdated hardware). Reflashing the firmware should fix the problem. If it happens often consider upgrading to a [more powerful adapter](../getting_started/what_do_i_need.md#supported-zigbee-adapter).
+3. If you see this happening right after you flashed the dongle with your Pi, try inserting the dongle into a laptop, have it sit there for a few seconds to be recognized by the system, and then put it back into your Pi. Scientists are still researching the reason of why this procedure works, but it actually does. 
 
 ### Verify that you put the correct port in configuration.yaml
 Execute the following command to find out the correct path:


docs/information/FAQ.md
```diff
@@ -89,6 +89,7 @@ Most of the times this is caused by zigbee-herdsman not being able to communicat
 2 common reasons of this error:
 1. The port of your serial adapter changed. Check [this](../getting_started/running_zigbee2mqtt.md#1-determine-location-of-the-adapter-and-checking-user-permissions) to find out the port of your adapter.
 2. If you are using a CC2530 or CC2531; it is a common issue for this adapter to crash (due to its outdated hardware). Reflashing the firmware should fix the problem. If it happens often consider upgrading to a [more powerful adapter](../getting_started/what_do_i_need.md#supported-zigbee-adapter).
+3. If you see this happening right after you flashed the dongle with your Pi, try inserting the dongle into a laptop, have it sit there for a few seconds to be recognized by the system, and then put it back into your Pi. Scientists are still researching the reason of why this procedure works, but it actually does.
```
Koenkk, line 92: How many times were you able to reproduce this? Do you get the same result when just pulling the USB for 10+ seconds?

konstantint: It's the kind of a Heisenbug I don't think I can reproduce. The device didn't respond after flashing, nor after reflashing, nor after attempts to reinstall zigbee2mqtt, nor after putting it back and forth into the Raspberry in various ways while performing magical dances with a tambourine around it.

Then I found that strange hint about inserting the dongle it into a laptop, so I did exactly that - inserted it, observed the autodetected (Windows) driver being installed, put the dongle back into the Raspberry voila, it magically started working. I didn't try reflashing it again to see whether it becomes unresponsive, but I could do this experiment (in a couple of days perhaps) if you're curious.

I understand this sounds weird, but the fact of the matter is - following the advice helped me, hence the proposal to include it in the docs here - it's an easy thing to do when one is desperately trying to figure out why it doesn't work.

qqgg231: LGTM

karolzlot:
@Koenkk 
LGTM

docs/information/FAQ.md
Koenkk: @karolzlot are you able to reproduce this?

karolzlot: @Koenkk No, I didn't have this issue, but I think it won't hurt to add this info.

Koenkk: I think we should first be able to reproduce it and confirm it indeed fixes such issue. I don't want to add something just because it helped once without being clear what exactly solved it.

PR #771, opened 2021-05-31
fixed incorrect image for TS0115

qqgg231:
New image created from: https://www.aliexpress.com/i/4000981707898.html

I have this device.

Btw. 150x150 seems to be too small resolution. I suggest to make a rule that higher resolutions are also accepted.


Koenkk:
Thanks!

PR #777, opened 2021-06-02
added note

flatsiedatsie:
I bought this device without looking clearly, hoping it would allow for remote unlocking of the door. But this is not the case.


Koenkk:
thanks!

PR #780, opened 2021-06-03
 Add documentation for Home Assistant legacy entity attributes configuration

frenck:
This PR add documentation for the `homeassistant_legacy_entity_attributes` configuration option, introduced in https://github.com/Koenkk/zigbee2mqtt/pull/7683

### DIFF
--- a/docs/information/configuration.md
+++ b/docs/information/configuration.md
@@ -141,6 +141,12 @@ advanced:
   homeassistant_discovery_topic: 'homeassistant'
   # Optional: Home Assistant status topic (default: shown below)
   homeassistant_status_topic: 'homeassistant/status'
+  # Optional: Home Assistant legacy entity attributes (default: show below), when enabled:
+  # Zigbee2MQTT will send additional states as attributes with each entity. For example,
+  # A temperature & humidity sensor will have 2 entities for the temperature and
+  # humidity, with this setting enabled both entities will also have
+  # an temperature and humidity attribute.
+  homeassistant_legacy_entity_attributes: false
   # Optional: Home Assistant legacy triggers (default: shown below), when enabled:
   # - Zigbee2mqt will send an empty 'action' or 'click' after one has been send
   # - A 'sensor_action' and 'sensor_click' will be discoverd


docs/information/configuration.md
```diff
@@ -141,6 +141,12 @@ advanced:
   homeassistant_discovery_topic: 'homeassistant'
   # Optional: Home Assistant status topic (default: shown below)
   homeassistant_status_topic: 'homeassistant/status'
+  # Optional: Home Assistant legacy entity attributes (default: show below), when enabled:
+  # Zigbee2MQTT will send additional states as attributes with each entity. For example,
+  # A temperature & humidity sensor will have 2 entities for the temperature and
+  # humidity, with this setting enabled both entities will also have
+  # an temperature and humidity attribute.
+  homeassistant_legacy_entity_attributes: false
```
Koenkk, line 149: I believe the default should be `true`, otherwise this is a breaking change: https://github.com/Koenkk/zigbee2mqtt/discussions/7249#discussioncomment-678457

frenck: Yes, it is a breaking change.

The use should be prevented/minimized, as it is actually really bad for the system to push and process all that additional data.

The config parameter is here to provide user who do rely on them (which they should not), still can avoid the breaking change and migrate at their own pace.

Switching the default, makes this (what is considered for quite some time now) bad practice never go away.

Koenkk: > Additionally, new users should start with sane defaults any time.

Totally agree with this, otherwise you indeed end up in a situation where even new users still keep using the legacy things (this is already happening with some of the legacy options).

I've added a mechanism which disables legacy options on a new network start: https://github.com/Koenkk/zigbee2mqtt/blob/7b13b2d213ef7db1e36c2dea97e0b40e94d38579/lib/controller.js#L120

For this option it would be:
```
settings.set(['advanced', 'homeassistant_legacy_entity_attributes'], false);
```

frenck: Nice! 👍 

Implemented the behavior in the parent PR for this documentation PR. Adjusted this PR to match the new default.

Koenkk:
Thanks!

PR #782, opened 2021-06-03
Update E1524_E1810.md

Spiffo:
added brightness_down_hold to the enum of actions, and sorted the enum a bit more logically.


Spiffo:
was on work VPN

PR #786, opened 2021-06-07
ADD Woox R7049 documentation

mattleh:
copied most parts from 5p1vj8r
but should be correct
picture copied from reichelt and converted


Koenkk:
Thanks!

PR #789, opened 2021-06-09
Add note about Axis Gear and AC Adaptor

sjorge:
Ref: https://twitter.com/theaxislife/status/1402313486026354694

When using the AC Power Adapter the battery value can be ignored as it will hover around 30%.


Koenkk:
Thanks!

PR #795, opened 2021-06-15
Add note about restarting HA when disabling legacy entity attributes

frenck:
Added a small note for when someone disables `homeassistant_legacy_entity_attributes`, it requires a Home Assistant restart.

See: Koenkk/zigbee2mqtt#7802


Koenkk:
Thanks!

PR #796, opened 2021-06-16
added Hue Outdoor Spot Lily (1742830P7)

fs79:
cloned the device page 1741830P7 to 1742830P7 and changed all occurences from 1741830P7 to 1742830P7.
Didn´t changed the picture link because device looks similar than 1741830P7.


Koenkk:
Thanks!

PR #804, opened 2021-06-26
add power measurement function to SP2600ZB

charlesdagenais:
complete documentation after adding power measurement feature to SP2600ZB in zigbee-herdsman-converters


Koenkk:
Thanks!

PR #806, opened 2021-06-28
Update documentation for QBKG26LM decoupled mode.

ruifung:
Now it only states the relevant info for this model instead of for some other 2 gang model that isn't relevant here.

Though I am unsure about the GET operation mode part, it doesn't seem to work properly on my device.


Koenkk:
This is currently generated from https://github.com/Koenkk/zigbee2mqtt.io/blob/60be84f223aef242f86104cb2159c8702d4b01ef/docgen/device_page_notes.js#L2998 so I prefer to keep it as is

ruifung:
No longer relevant due to adding relevant exposes.

PR #808, opened 2021-06-29
Update W2049.md

MattWestb:
The name of the type is **not** production year and week.
The real name shall being used and not YYWW.

The name is [E2001 (stainless)](https://fccid.io/FHO-E2001) and [E2002 (white)](https://fccid.io/FHO-E2002).

### DIFF
--- a/docs/devices/W2049.md
+++ b/docs/devices/W2049.md
@@ -1,19 +1,19 @@
 ---
-title: "IKEA W2049 control via MQTT"
-description: "Integrate your IKEA W2049 via Zigbee2MQTT with whatever smart home
+title: "IKEA E2001/E2002 control via MQTT"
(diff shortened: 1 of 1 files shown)


MattWestb:
CSA certificate: https://zigbeealliance.org/zigbee_products/styrbar-remote-control/

Koenkk:
Thanks!

MattWestb:
Sorry i see i have missing one E  in the last E2001/E2002.

But i think its easier you is fixing it then the index file also need being renamed that is still W2049 in the device list  :-((

Koenkk:
@MattWestb I see it's already correct in zigbee-herdsman-converters, so these docs will be updated with the next release. (tomorrow)

PR #809, opened 2021-06-29
Add Nue WL-SD001-9W device

bayview252:
Add device markdown with jpg for new Nue WL-SD001-9W RGB Downlight.  It has narrower color_temp range than currently defined Nue lights.


bayview252:
I have tested by replacing this file in /app/node_modules/zigbee-herdsman-converters/devices and restarting the container. It works as expected. Thanks for your work.

Koenkk:
Thanks!

PR #812, opened 2021-06-30
FAQ updates

SNoof85:
I just migrate from a CC2531 to a zzh! and by reading the doc, I thought it could be better :)

remove duplicate advice + more details
add hardware path thing for migration to new coordinator


Koenkk:
Thanks!

PR #815, opened 2021-07-04
Update supported_adapters.md

MattWestb:
EFR32 is not only Zigbee radios so shall being written EFR32MG that is the Zigbee family of the EFR32.
EM35X is also supported then its have enough hardware and we have ported EZSP 6.7.8.0 to it and its working OK.
From the host system its transparent wot radio hardware its running only the version of the firmware that is making it  compatible or not.
List of radios with verified working firmware https://github.com/zigpy/zigpy/wiki/Coordinator-Firmware-Updates


Koenkk:
Thanks!

PR #818, opened 2021-07-05
Update ubisys J1 docs to reflect HA's changes to *_template MQTT configuration attributes

felixstorm:
Since a few Home Assistant versions ago one must also remove `tilt_status_template` for covers without tilt and `set_position_template`, `position_template` and `tilt_status_template` for covers with neither lift nor tilt from HAs MQTT auto discovery message for the covers to still be discovered


Koenkk:
Thanks!

PR #821, opened 2021-07-09
SJCGQ11LM pairing instructions correction

Chriserus:
Pairing instructions needed to be adjusted. It is about 5 seconds till we can see the blue light blinking three times.


Koenkk:
Thanks!

PR #822, opened 2021-07-10
Update pairing information for Bitron devices

wluberti:
Since the official website of Bitron is no longer available I though it wise to add their pairing information.


Koenkk:
Thanks!

PR #839, opened 2021-07-31
Add troubleshooting to Bitron AV2010/34

Julusian:
I spent quite a few hours trying to figure out why I wasnt receiving the actions data.
After a lot of googling and reading more into zigbee I discovered that the cause was the device not part of a group and so was not broadcasting the scene change.


Koenkk:
Thanks!

PR #852, opened 2021-08-05
HG06668 is a door bell _button_, not a door bell

ernstkl:
This device is just a button (which is clear from the "exposes" section, and from descriptions found online).
But the name "wireless door bell" sounds like it were a device that can deliver a signal. [Wikipedia says:](https://en.wikipedia.org/wiki/Doorbell) "A doorbell is a signaling device..." (I nearly wrongly ordered one of these)


Koenkk:
Thanks!

PR #863, opened 2021-08-16
Add Philips Hue (BT) to reporting table

sjorge:
Someone mentioned it was working for the newer Philips Hue with BT bulbs (Manu=Signify), I bought one to verify. Everything except color_mode is supported.


Koenkk:
Great, thanks!

PR #864, opened 2021-08-17
Documentation corrections for Sunricher ZGRC-KEY-013 / sr-zg9001k12-dim-z4

meinarsson:
Changed to the correct picture and added information about how to reset and pair the remote


Koenkk:
Thanks, also fixed the model number so it will display correctly in the next z2m release.

PR #867, opened 2021-08-17
remove long shipping times hint from Slaesh's CC2652RB stick.

tioan:
remove long shipping times hint from Slaesh's CC2652RB stick.
https://slae.sh/projects/cc2652/#get-one


Koenkk:
Thanks!

PR #875, opened 2021-08-20
Updated and new device documentation

TheStigh:


TheStigh:
Updated documentation for:
SRAC-23B-ZBSR
SMSZB-120

Added documentation for:
SIRZB-110
HSE2919E

docs/devices/HSE2919E.md
```diff
@@ -0,0 +1,111 @@
...
+
+| Model | HSE2919E  |
+| Vendor  | Datek  |
+| Description | Water leak sensor |
+| Exposes | battery, battery_low, temperature, water_leak, tamper, linkquality |
+| Picture | ![Datek HSE2919E](../images/devices/HSE2919E.jpg) |
+
+## Notes
+
+This device support `iasZone` and `iasWd`
```
Koenkk, line 20: I don't think this should be mentioned here, most users probably don't know what this means

TheStigh: I can remove it if you like, but for those who know what it is and perhaps searching for devices supporting it (like I did), it is handy information. I've been contacted by other users about this after updating the devices to support this feature.

Your call Koen 👍

Koenkk: I'm fine with it, but only if an explanation is added what it actually means (for the user)

TheStigh: Ok, removed the "**This device support `iasZone` and `iasWd**`" as there is no easy way of explaining iasZone and iasWd.

Koenkk:
Thanks!

PR #887, opened 2021-08-30
Missing quotation marks from example payloads and responses

clemwo:
The example responses from Scan and the example payload from Identify are missing quotation marks


Koenkk:
Thanks!

PR #891, opened 2021-09-02
Update Notes section for all Hive receivers

robertalexa:
Hi @Koenkk 

This is an update to the documentation as a result of PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/2866

This was thoroughly discusses in issue https://github.com/Koenkk/zigbee2mqtt/issues/8184 and I have made the promise to update the Notes to explain how to use all the devices commands after the PR is in official release.

There is a bit of repetition on the dual channel receivers, but i found that to be the most clear explanation, because each endpoint has little quirks that the other doesn't, and same the other way around.

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 _site/
 .sass-cache/
+.idea
 .DS_Store
 .idea/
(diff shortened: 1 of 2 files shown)


robertalexa:
Also, I wasn't sure if I need to re-build the devices .md files myself, or just provide the modified notes file and the file rebuilding gets handled at your end. Let me know if you need me to do this.

PS: I have another PR in z2m.io in the same position. I can update them both if needs be

Koenkk:
I will do the docgen update right away, could you check the result in a few mins?

PR #892, opened 2021-09-02
Replace json type with js

robertalexa:
@Koenkk 

Apologies for having to create this PR.

I have replaces `json` with `js` for the code markdown as it was looking hidious.

I have corrected the notes files, but also the devices, so no need for a docgen, just a merge.

The docs for Alarms and Hive look good apart from the above.
Thanks for merging


Koenkk:
no problem!

PR #893, opened 2021-09-02
Updated pairing instructions and description

owangen:
Added pairing instruction for Elko dimmer and Namron 4512726.
Also added instructions on changing color modes for Namron 4512726


Koenkk:
Thanks!

PR #894, opened 2021-09-03
Include incomplete ZCL values for Alarm Keypads

robertalexa:
Hi @Koenkk 

As a results of the new issue https://github.com/Koenkk/zigbee2mqtt/issues/8630 I have reviewed the notes and realised I have not added all the valid ZCL values for `arm_mode`

I have now added them to the notes file but also the device.md so no need for a docgen, just a merge.

Apologies for giving you extra work.

PS: I will handle the above mentioned issue. It is just a case of users having to change their automation flows to use the standard logic, but all the functionality is available and correct.


Koenkk:
Thanks!

PR #898, opened 2021-09-08
MFKZQ01LM.md - complements the pairing instructions

cidus:
Just got this device, and in order to pair it I had to keep doing short presses to the button or the interview process would not finish. 
Using zigbee2mqtt from homeassistant and the cc2531.


Koenkk:
Thanks!

PR #905, opened 2021-09-15
Added E3007 starkvind

derfloDev:
Support for the new Ikea Starkvind air purifier.
At the moment only the fan control is working.
The device also has a air quality sensor (heimanSpecificPM25Measurement) but I wasn't able to get it work.


Koenkk:
Thanks!

MattWestb:
Sorry but _Starkvind_ is **not** E3007.

The standalone is E2007 https://fccid.io/FHO-E2007 and the table is E2006 https://fccid.io/FHO-E2006.

Both models have external power supply so is more then likely the same devices sold all over the world then its being released in all regions with different PSU.

PR #908, opened 2021-09-16
Ikea E1524 `action: toggle` was missing

cturra:
really small update here, just adding `action: toggle` to the document as it was missing. the toggle action is sent when the middle "power" button on the remote control is pressed.


Koenkk:
Thanks!

PR #912, opened 2021-09-18
Add images for devices

Asleep-at-the-wheel:
These two devices I just had PR requests done to fix proper support for them.

They don't look anything like what they show up as right now (they use the chipset image).

Might help new user uptake and avoid confusion if they actually see the devices.


Koenkk:
Could you resize the images to be 150x150?

Asleep-at-the-wheel:
My bad, should be done.

Koenkk:
Thanks!

PR #915, opened 2021-09-23
Update SNZB-02.md

mgwoj:
I think it can be valuable for every user to know what battery type is used inside the device.


docs/devices/SNZB-02.md
```diff
@@ -15,6 +15,7 @@ description: "Integrate your SONOFF SNZB-02 via Zigbee2MQTT with whatever smart
 | Exposes | battery, temperature, humidity, voltage, linkquality |
 | Picture | ![SONOFF SNZB-02](../images/devices/SNZB-02.jpg) |
 | White-label | eWeLink RHK08 |
+| Battery type | CR2450 |
```
Koenkk, line 18: This part is generated (and thus it will automatically be removed on regeneration), please move to notes.

Koenkk:
Thanks!

PR #918, opened 2021-09-27
Update notes about ConBee

DubhAd:
A _lot_ of people miss the required `configuration.yaml` entry. This is to hopefully make it easier for those people.


Koenkk:
Thanks!

PR #924, opened 2021-09-30
Update E1812.md

Deshmir:
Added extra information, since the button will pair itself to a nearby light bulb

### DIFF
--- a/docs/devices/E1812.md
+++ b/docs/devices/E1812.md
@@ -22,6 +22,7 @@ description: "Integrate your IKEA E1812 via Zigbee2MQTT with whatever smart home
 Pair the remote to Zigbee2MQTT by holding it close to the coordinator and
 pressing the inside button, next to the CR2032 battery, 4 times.
 The red light on the (front of the) remote will now flash a few times.
+Make sure you have any nearby IKEA lightbulbs turned off from power, else the button will pair with the light instead of the coordinator.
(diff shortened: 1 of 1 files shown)


docs/devices/E1812.md
```diff
@@ -22,6 +22,7 @@ description: "Integrate your IKEA E1812 via Zigbee2MQTT with whatever smart home
 Pair the remote to Zigbee2MQTT by holding it close to the coordinator and
 pressing the inside button, next to the CR2032 battery, 4 times.
 The red light on the (front of the) remote will now flash a few times.
+Make sure you have any nearby IKEA lightbulbs turned off from power, else the button will pair with the light instead of the coordinator.
```
Koenkk, line 25: What is the problem of this?

Deshmir: If a light is on power there is a high probability the button binds itself to the light instead of the coordinator.
Happened to me multiple times; holding it next to the coordinator and it still binds itself to the ceiling bulb.

Koenkk: That should only happen when you hold the reset button, if you press the button quickly 4 times it should only pair and never bind itself (I assume with binding you mean a direct control)

Deshmir: I guess I did it wrong then.

PR #925, opened 2021-09-30
Update LS12128.md

kds69:
After trying to change my own zigbee network to different channel than 11, I figured out with a sniffer the LS12128 is only polling on channels 11, 15, 20, 25. Smart choice given the WiFi overlap bandwidth, unfortunately doesn't take channel 26.


Koenkk:
Thanks!

PR #930, opened 2021-10-03
Replace the image of DLKZMK11LM

keniji:
The current one should be from `SSM-U02` (without neutral). Model `DLKZMK11LM` is with neutral, so I replaced it with `SSM-U01`'s (with netral).


Koenkk:
Thanks!

PR #933, opened 2021-10-07
Docs update for exposing scenes and setting scene name

Koenkk:
See https://github.com/Koenkk/zigbee2mqtt/pull/9056


docs/information/scenes.md
```diff
@@ -9,6 +9,7 @@ Scenes can be created in two ways, by storing the current state (`scene_store`)
 
 ### `scene_store`
 This is the easiest way to create a scene. First set the device or group in the desired state. Next store the current state as a scene by sending a command to `zigbee2mqtt/[GROUP_OR_DEVICE_FRIENDLY_NAME]/set` with payload `{"scene_store": SCENE_ID}` where `SCENE_ID` is a number (e.g. `1`).
+To also provide a name for the scene (shown in e.g. the frontend) send e.g. `{"scene_store": {"ID": 1, "name": "My scene"}}`.
```
sjorge, line 12: Is this name only use for the frontend? As there is also a scenename attribute in the ZCL so if the same constraint is used it can probably be written to the device as wel! (and retrieved if that ever gets implemented)

Koenkk: Yes it will be used in the frontend, @nurikk just added it.

I know scene names can be written to a device, however I don't see the benefit of it and I don't know if all devices support this.

PR #940, opened 2021-10-11
Added initial documentation for how to find Tuya Data Points

danieledwardgeorgehitchcock:
As the popularity of TuYa device requests is increasing, and people are struggling to map TuYa data points to their functions, I thought it would be best to document my process for finding the data point functions via the TuYa Cloud Platform and TuYa gateway which I discovered through trying to implement my own TuYa device.

I hope this will help some people! It really does take a LOT of the hassle out of the trial and error / sniff approach in determining the data points!

### DIFF
--- a/.github/workflows/merge_master_to_dev.yml
+++ b/.github/workflows/merge_master_to_dev.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Merge master -> develop
-        uses: devmasx/merge-branch@v1.3.1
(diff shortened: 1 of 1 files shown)


Koenkk:
Agree, lets keep it separate, but I think it will be good to add references to this how-to in https://www.zigbee2mqtt.io/how_tos/how_to_support_new_tuya_devices.html

danieledwardgeorgehitchcock:
@Koenkk done!

Koenkk:
Great, big thanks!

PR #972, opened 2021-11-01
Outline / in friendly_name

psi-4ward:


docs/guide/configuration/devices-groups.md
```diff
@@ -43,6 +43,11 @@ Every Zigbee Device supports the following list of options.
 **`friendly_name`**  
 Used in the MQTT topic of a device. By default, this is the device ID (e.g. `0x00128d0001d9e1d2`).
 
+::: tip
+You can use the `/` separator in `friendly_name` to _group_ devices.  
+For example, using a friendly-name like `Kitchen/Floor Light` would result in a corresponding MQTT structure with `Kitchen` as folder containing `Floor Light`.
```
Koenkk, line 48: There is one catch here, a friendly name is not allowed to end with `/` + one of the possible endpoint names: https://github.com/Koenkk/zigbee2mqtt/blob/cf68507ec3a11826ad794035208b83f4b37ee8d7/lib/util/utils.ts#L30

Z2M already handles this and will fail to start, but I think it would be good to mention it here too.

psi-4ward:
Should we add the warning in MQTT-Topics page too?

docs/guide/configuration/devices-groups.md
```diff
@@ -43,6 +43,17 @@ Every Zigbee Device supports the following list of options.
...
+:::
+
+::: warning
+Note that a `friendly_name` is **NOT** allowed to end with `/` or use one of the
```
Koenkk, line 52: It is not allowed to end with a `/` or `/` with one of the endpoint names (e.g. `/left` is not allowed)

docs/guide/configuration/devices-groups.md
```diff
@@ -43,6 +43,17 @@ Every Zigbee Device supports the following list of options.
 **`friendly_name`**  
 Used in the MQTT topic of a device. By default, this is the device ID (e.g. `0x00128d0001d9e1d2`).
 
+::: tip
+You can use the `/` separator in `friendly_name` to _group_ devices.
```
Koenkk, line 47: I think "to group devices" is confusing here, this may sound like it creates zigbee groups. Maybe change it to structure devices?

docs/guide/configuration/devices-groups.md
```diff
@@ -43,6 +43,17 @@ Every Zigbee Device supports the following list of options.
 **`friendly_name`**  
 Used in the MQTT topic of a device. By default, this is the device ID (e.g. `0x00128d0001d9e1d2`).
 
+::: tip
+You can use the `/` separator in `friendly_name` to _group_ devices.  
+For example, using a friendly-name like `Kitchen/Floor Light` would result in a corresponding MQTT structure with `Kitchen` as folder containing `Floor Light`.
```
Koenkk, line 48: "as folder containing" this depends on the MQTT client you use right?

Koenkk:
Thanks!

PR #984, opened 2021-11-03
Add pseudo image for envilar 2ch box relay

tbowmo:
Adds pseudo image for 2ch envilar box relay (they all look the same)


Koenkk:
Thanks!

PR #991, opened 2021-11-05
Fix naming convention frontend to WebUI

psi-4ward:


Koenkk:
I'm still not sure what definition is better, if I look at the definition of frontend it seems to match with what we have: https://techterms.com/definition/frontend

My concern is that people get confused if we rename this to WebUI since it is referred to as the frontend in github issues, discussion and in the `configuration.yaml`. 

@nurikk what do you think?

nurikk:
What's wrong with existing naming?) What's the benefit or this change?

psi-4ward:
So in terms a frontend can be anything which makes use of a service (or something which provides an interface to access data). MQTT-Explorer is a frontend for MQTT, NodeRED is a frontend for MQTT, our WebUI is a frontend for Z2M. So my idea was to specify the _type-of_ frontend. 

But if you argue that "frontend" is well known as the web-ui here we can also drop the renaming but than, we should also rename `webui.md` to `frontend.md`.

Greets ;)

Koenkk:
Lets keep it frontend, I've changed all references of webui to frontend: https://github.com/Koenkk/zigbee2mqtt.io/commit/4ed536efe3bdf3804806146fb173a8761fae6b30

PR #996, opened 2021-11-05
Fix docker-compose in quickstart

psi-4ward:
serialport derives infos about the device from udev in some cases

### DIFF
--- a/docs/guide/getting-started/README.md
+++ b/docs/guide/getting-started/README.md
@@ -59,7 +59,7 @@ First, we create the `docker-compose.yml` file which defines how Docker would ru
 
 
 ```yaml
-version: '3.5'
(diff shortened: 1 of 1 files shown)


docs/guide/getting-started/README.md
```diff
@@ -77,6 +77,7 @@ services:
     image: koenkk/zigbee2mqtt
     volumes:
       - ./zigbee2mqtt-data:/app/data
+      -v /run/udev:/run/udev:ro
```
Koenkk, line 80: `-v` looks like a typo ( should be  `-` i think)

psi-4ward: thx, force pushed it :)

PR #1003, opened 2021-11-10
remove warning about shipment issues, as they seem to be resolved for…

RubenKelevra:
… a while now.


Koenkk:
Thanks, hopefully service stays good and we never have to add this back

PR #1014, opened 2021-11-12
Update README.md

fschaeck:
The link for the CC2538-BSL tool for flashing the CC1352/CC2652/CC2538 adapters was pointing to the Texas Instruments FLASH PROGRAMMER 2 as well.
I changed that link to now point to https://github.com/JelmerT/cc2538-bsl


Koenkk:
Thanks!

PR #1028, opened 2021-11-22
Update adapter-settings.md

DubhAd:
Adding a little piece on why, if you're manually configuring it, to use the `/dev/serial/by-id/` path.


Koenkk:
Thanks!

PR #1033, opened 2021-11-24
Add 8719514302235

LevitatingOrange:
This change adds the device EAN 8719514302235 to the list of devices. See also [#3409](https://github.com/Koenkk/zigbee-herdsman-converters/pull/3409). I just copied the docs from another philips hue filament device and changed a few things, hope that's ok/enough. The device behaves just as any other philips filament bulb in my tests.


Koenkk:
Thanks!

PR #1034, opened 2021-11-25
Update PTM_215Z.md

sandervandenberg:
Added an alternative way of resetting the switch, found here:
https://www.senic.com/pages/support-center
(under "I have problems with the FOH...")
thanks to this post (in Dutch) https://gathering.tweakers.net/forum/list_message/69560682#69560682

I needed to do this with my Senic Friends of Hue switches to get them to pair with Zigbee2MQTT (using a Conbee II)

BTW The Senic switches are not in the Z2M device list yet, and I have no idea how to add them, but they use the PTM215Z as well.
Furthermore, I think I read somewhere that the Conbee II also supports Green Power devices, which means you would not need a Hue light to get them to work, but have not been able to find this again, or to verify this. But maybe you already know ;-)


Koenkk:
Bedankt! 😄

PR #1035, opened 2021-11-25
Update Eco-Dim.07.md

sandervandenberg:
From the manual (and partly my own experience) here:
https://www.ecodim.nl/eco-dim07-zigbee-basic.html


Koenkk:
Thanks!

PR #1037, opened 2021-11-26
Update README.md with link to llama-bsl

Hedda:
Update README.md with link to [llama-bsl](https://github.com/electrolama/llama-bsl) (developed by omerk of Electrorama ZZH fame) which is fork of cc2538-bsl script but with added features to make it more user-friendly to new users, such as board type detection + automation download of coordinator and router firmware (features that only make sense in a limited context, therefore unlikely to be ever merged into the upstream cc2538-bsl script).


Koenkk:
FYI @omerk

PR #1040, opened 2021-11-30
Update configuration filename

8none1:
The docs stated that the config filename was "configuration.yml" where as actually `zigbee2mqtt` will fail to start unless the file is named `configuration.yaml`.


Koenkk:
Thanks!

PR #1041, opened 2021-11-30
Fix CC-Debugger pins image error

rakete1:
The CC-Debugger pins image was not displayed due to a faulty div alignment. The alignment therefore was removed.


Koenkk:
Thanks!

PR #1045, opened 2021-12-02
Docs - Add MQTT broker requirement in getting started guide

teameh:
I'm an absolute beginner. It was not clear to me I had to make sure a broker was running until I figured this out from the logs.


teameh:
See also https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/pull/246/files

teameh:
Would it also make sense to add it to the "Zigbee2MQTT fails to start" FAQ? In my case it wouldn't start because I had no broker running.

Koenkk:
This is fine, thanks!

teameh:
\0/ thanks for the quick reply!

PR #1046, opened 2021-12-04
Update TS004F.md

theetron:
Hereby some more info about this device.

### DIFF
--- a/docs/devices/TS004F.md
+++ b/docs/devices/TS004F.md
@@ -29,10 +29,9 @@ pageClass: device-page
 
 
 ## Notes
-### Serialnumer
+### Serialnumber
 |Location|Serial|
 |-----|-----|
-|On the backside|ESW-OZAA-EU|
 |Above the battery|ESW-OZBA-EU|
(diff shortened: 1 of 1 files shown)


docs/devices/TS004F.md
```diff
@@ -28,6 +28,26 @@ pageClass: device-page
 <!-- Notes END: Do not edit below this line -->
```
Koenkk, line 28: - Please add everything above this line
- I'm not sure if the serial number is useful information, this device will probably come with many different serial numbers since the module is used in a lot of devices.

theetron: I added the serial because the inside differs from the outside. On the TS0044 page the serial is also mentioned. Maybe only mention the inside serial?

theetron: I see a typo in the serial number header..

theetron:
As this is my 1st contribution, do I need to do something else now?

Koenkk:
Thanks!

PR #1057, opened 2021-12-08
Image for Namron 3802961

ivareri:
Image for Namron 3802961, support added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/3470


ivareri:
huh, pushed a button that did not do what i expected it too. Let me know if the latest commit (sync with master) causes problems

Koenkk:
Thanks!

PR #1059, opened 2021-12-08
SLZB-05 Zigbee LAN adapter

sergge1:
Added Smartlight's SLZB-05 Zigbee\LAN adapter to the list of supported adapters.


Koenkk:
Thanks

PR #1060, opened 2021-12-09
Update calibration instructions for Tuya TS130F

McGiverGim:
According to the instructions of the product, you need to press the "pause" button after fully closed. I tried without it and it did not work, with it all worked perfectly.

I've removed too the mosquitto command to put it into calibration mode, I think it can produce confusion to people. There are easier ways to enable/disable de calibration mode, like using the z2m UI or HA UI.

The LoraTap white-label image of the product is quite different from the Tuya one. Is there a way to add different images for the same product here? Can I simply add another Picture to the table?

### DIFF
--- a/docs/devices/TS130F.md
+++ b/docs/devices/TS130F.md
@@ -31,10 +31,12 @@ Press on stop button for 10 seconds to enter pairing mode
 
 ### Calibration
 
-* Open curtains completely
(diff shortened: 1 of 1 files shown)


Koenkk:
For whitelabel devices a different picture can be added by naming it `SC400.jpg`, note that this picture will only appear on the overview page: https://www.zigbee2mqtt.io/supported-devices/

PR #1061, opened 2021-12-09
Fix broken link for Cluster Library

tioan:
Correct Link for https://github.com/Koenkk/zigbee-herdsman/blob/master/docs/07-5123-08-Zigbee-Cluster-Library.pdf


Koenkk:
Thanks!

PR #1063, opened 2021-12-09
TS011F_plug_1 - update white label with BW-SHP15

MaciejR89:
Checked and confirmed that BlitzWolf BW-SHP15 plug (discovered as TS011F_plug_1) works with all available options for this device.


Koenkk:
Thanks!

PR #1066, opened 2021-12-11
dimmer switch exposes more values

Kat-CeDe:


Koenkk:
This part of the docs is generated and should not be changed directly, please make a pr to update https://github.com/Koenkk/zigbee-herdsman-converters/blob/47db78733a48c6cbbb268aadb60140ac6cb66f31/devices/philips.js#L1741 instead.

PR #1070, opened 2021-12-12
Conbee II also works with Green Power devices

sandervandenberg:
I wanted to know if the Conbee II stick could also handle Green Power devices. So this is what I did:

I unscrew all Hue bulbs in my house, also unscrew any other lights (ikea) that I could easily reach. Then turned off the mains power to every room that had smart devices in it (built-in smart dimmers, smart plugs) which also shut down my Hue Bridge.
By now, my server (with HA and Zigbee2MQTT and Conbee II stick) and network equipment (switch, router, modem) were running on the UPS.
I then got into HA on my tablet and went to Node Red and switched to the tab where I could see the messages from the debug nodes, so I could see the clicks on the Senic switch come in.
And clicking it worked! So I can hereby confirm that the Conbee II stick also handles green power devices.

(Also added a newline more to the end)

### DIFF
--- a/docs/devices/PTM_215Z.md
+++ b/docs/devices/PTM_215Z.md
@@ -30,7 +30,7 @@ pageClass: device-page
 ### Green Power
 This is a Zigbee Green Power device which allows it to be very energy efficient.
 Messages from Green Power devices cannot be "understood" by normal Zigbee devices, therefore they need to be "translated" first.
-Not all Zigbee devices can do this translation, currently the only devices known to do this are Philips Hue devices. This means that the Green Power device has to be in range of a Philips Hue device in order to use it.
+Not all Zigbee devices can do this translation, currently the only devices known to do this are Philips Hue devices and the Conbee II stick. This means that the Green Power device has to be in range of a Philips Hue device or the Conbee II stick in order to use it.
 
 Green Power devices don't support binding and are not included in network scans.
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Please also update the docs of other supported green power devices.

sandervandenberg:
Instead of updating all the other green power devices, I think it would be better to make a section dedicated to green power devices. I'm thinking under the Advanced->Zigbee menu.
Then for each green power device we can just link to that section instead of having to duplicate the same text everywhere.
Not sure if this is something I can do?

Also, I think it would be a good idea to have a flag or something for each device that indicates whether it is a green power device (just like you have "model", "vendor", "description", "exposes", etc?
And in addition to that a flag that indicates whether the device can "translate" green power messages (would be true for hue lights and conbee ii)

What would be nice as well is to have a clear way to have multiple switches (Gira Senic, Trio2Sys, etc) point to the Enocean white label module, but still have the original picture of the switch attached to it). Maybe this is possible already, but I have no idea how to do this.

Koenkk:
> Then for each green power device we can just link to that section instead of having to duplicate the same text everywhere.
Not sure if this is something I can do?

Sounds good

> Also, I think it would be a good idea to have a flag or something for each device that indicates whether it is a green power device (just like you have "model", "vendor", "description", "exposes", etc?

What do you want to do with this? If it is a green power device is already documented on the device page itself


> And in addition to that a flag that indicates whether the device can "translate" green power messages (would be true for hue lights and conbee ii)

This is not really maintainable, it would be better to give a list of devices which may support green power proxies (e.g. Hue, TuYa) (can be done on the same page as point 1)

> What would be nice as well is to have a clear way to have multiple switches (Gira Senic, Trio2Sys, etc) point to the Enocean white label module, but still have the original picture of the switch attached to it). Maybe this is possible already, but I have no idea how to do this.

This can be done with the whiteLabel property, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/5a8e477a40a979611fa91cde75eb5a62f6ba6e68/devices/tuya.js#L55

PR #1073, opened 2021-12-15
Update TS0201.md

vdebrist:
Max precision 1 (xx,x).


Koenkk:
The provided precision of this sensor could differ from the max settable precision (the option).

PR #1076, opened 2021-12-15
Update ZB003-X.md

ip-ua:
Device accepting setting change only when powered via USB cable. 
Battery powered device just ignore any changes like Reporting_time, Temperature_calibration, Pir_enable, etc.


Koenkk:
Thanks!

PR #1084, opened 2021-12-21
Kwikset Lock

LowHanger:
This adds the 914 Kwikset lock documentation as well as its picture to the repository.


Koenkk:
Thanks!

PR #1085, opened 2021-12-22
Completed info on pairing and resetting

VincentSC:
It stated wrongly that connection will happen after a reset. Corrected the info on pairing and resetting


Koenkk:
Thanks!

PR #1087, opened 2021-12-25
Update ota_updates.md

lukyjay:
The OTA test server doesn't provide many / any updates anymore and the main OTA server is working again, so there's no reason for users to have this option enabled as of Dec-21. It could be increasing the number of github issues where OTA updates are not available.


Koenkk:
Thanks!

PR #1095, opened 2021-12-29
Adapt documentation to Tuya "multiple DP per packet" change

gmbnomis:
In the "from Zigbee" converters, DPs should be consumed using a loop now.

Describe the alternative approach to send multiple DPs at once for the "to
Zigbee" converters.


gmbnomis:
Depends on https://github.com/Koenkk/zigbee-herdsman-converters/pull/3572

PR #1096, opened 2021-12-29
Add a note about pressing a button on the remote prior to binding

Stamper-repo:
For IKEA Tradfri E1810 remote.


Koenkk:
thanks!

PR #1101, opened 2021-12-31
Clarify logging twice with systemd unit

Phyks:
As discussed in https://github.com/Koenkk/zigbee2mqtt.io/pull/1100


Koenkk:
Thanks!

PR #1104, opened 2022-01-02
add image for namron 3802966

engvik:
Adds image for device support added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/3605.


Koenkk:
Thanks!

PR #1113, opened 2022-01-05
Updated FLASH-PROGRAMMER-2 Flash description

HeimdallMidgard:
Updated FLASH-PROGRAMMER-2 Flash description " in the right corner under Secondary MAC check "Retain secondary IEEE" " else the IEEE is not beeing used and the migration doesn't work


Koenkk:
good to know, thanks!

PR #1119, opened 2022-01-06
add Flashing the firmware via cc2538-bsl

der-pw:
In this example I am using a Sonoff dongle and flashing it via cc2538-bsl.

### DIFF
--- a/docs/guide/adapters/flashing/flashing_via_cc2538-bsl.md
+++ b/docs/guide/adapters/flashing/flashing_via_cc2538-bsl.md
@@ -1,0 +1,67 @@
+---
+---
+# Flashing the firmware via cc2538-bsl
+
+TI CC13xx/CC2538/CC26xx Serial Boot Loader  
+
+CC2538-bsl is a python script that communicates with the boot loader of the Texas Instruments CC2538, CC26xx and CC13xx SoCs (System on Chips).  
+
+In this case we use a Sonoff Zigbee 3.0 USB Dongle Plus and flash the firmware without opening the case to press the "Boot" button.
+
+## Prepare your system (debian based)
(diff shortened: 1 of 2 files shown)


docs/guide/adapters/flashing/flashing_via_cc2538-bsl.md
```diff
@@ -0,0 +1,73 @@
...
+```
+2. Install needed python packages.  
+```bash
+sudo pip3 install pyserial intelhex
+```
+## Download cc2538-bsl
+3. We need the ITead_Sonoff_Zigbee-delay "bootloader" feature.  
+```bash
+mkdir z2m_fw-upgrade
+cd z2m_fw-upgrade
```
Koenkk, line 25: Doesn't seem to be necessary, the unzip wil already create a separate directory.

Koenkk:
Made some changes, could you check if it is OK?

docs/guide/adapters/flashing/flashing_via_cc2538-bsl.md
der-pw: Yes, You're right, but I wanted everything in one directory to keep things tidy.
If You want, i can give a hint, for my reasons.

der-pw:
> ... could you check if it is OK?

Yes, it's okay! Thank You!

docs/guide/adapters/flashing/flashing_via_cc2538-bsl.md
Koenkk: I will rename the directory to cc2538-bsl instead (makes more sense then `z2m_fw-upgrade`)

PR #1121, opened 2022-01-06
fix description

Giancky79:
cleaning and fix some description


Koenkk:
This page is generated based on https://github.com/Koenkk/zigbee-herdsman-converters/blob/cd336fcea042b8e96dc7d658ccc514a64512f4a8/devices/bticino.js#L71 and should not be modified directly.

Giancky79:
perfect, but F20T60A it's a same 
https://www.zigbee2mqtt.io/devices/412015.html 
and not to 
https://www.zigbee2mqtt.io/devices/FC80CC.html

Koenkk:
thanks, updated descriptions, docs will be regenerated with the next release (1 February)

PR #1132, opened 2022-01-11
Update 01_linux.md for i386 hardware

kyrios:
i386 hardware has no official node support anymore. But it's still working.


Koenkk:
thanks!

PR #1135, opened 2022-01-12
Update V3-BTZB_V3-BTZBE.md

XLDom:
Update for the pin code

### DIFF
--- a/docs/devices/V3-BTZB_V3-BTZBE.md
+++ b/docs/devices/V3-BTZB_V3-BTZBE.md
@@ -33,7 +33,15 @@ If pairing failed, try the followings:
 - Replacing the batteries of the danalock.
 
 ### App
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

smulle48:
How does this work?

I can't get it to work. I'm using HomeAssistant with zigbee2mqtt edge (also tried non edge).

I've tried to set the pinched both with the MQTT and within the zigbee2mqtt frontend, but the pinched does not show up in the iOS app, and I don't see the pinched in the debug window

Received Zigbee message from 'Hoveddør Lås', type 'commandOperationEventNotification', cluster 'closuresDoorLock', data '{"data":0,"opereventcode":1,"opereventsrc":1,"pin":{"data":[],"type":"Buffer"},"userid":65535,"zigbeelocaltime":4294967295}' from endpoint 1 with groupID 0

PR #1148, opened 2022-01-18
zigbee-herdsman-converters PR3698 model image

mdehaas:
I will also create documentation tomorrow. I believe this will however enable to show this image instead of the Zigbee logo in the devices list of Z2M, is this correct?


Koenkk:
> I believe this will however enable to show this image instead of the Zigbee logo in the devices list of Z2M, is this correct?

Yes, after the next release on 1 February

PR #1158, opened 2022-01-22
Update TS0207_repeater.md

DJBenson:
Added pairing and re-pairing instructions for this device as this procedure seems to be largely undocumented.


Koenkk:
thanks!

PR #1160, opened 2022-01-23
update flashing_via_cc2538-bsl.md

der-pw:
Instructions adapted to the new master of cc2538-bsl and also simplified (hopefully) the whole process.


Koenkk:
thanks!

PR #1173, opened 2022-01-26
New device TS0601_smart_air_house_keeper

1stsetup:


Koenkk:
thanks!

andre68723:
Hi @1stsetup,

using the same device, but noticed the pic is in z2mqtt is wrong. 
The one uploaded by you is correct, the one used  within the WebUI is wrong - https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/TS0601_smart_air_house_keeper.jpg

same for the docs 
https://www.zigbee2mqtt.io/devices/TS0601_air_quality_sensor.html

any idea how to fix this?

PR #1180, opened 2022-01-28
Update BW-SS7_2gang.md

windows97:
Seems that my previous commit was only valid for pairing for the first time.


Koenkk:
thanks!

PR #1187, opened 2022-01-31
Update WL4200 image

Maelstrom96:
The current picture is for the wrong model. Currently depicts the WL4200S with the external water sensor.


Koenkk:
thanks!

PR #1192, opened 2022-02-01
Add images for Mercator Ikuü devices

davidjb:
This PR adds various images for all the Mercator Ikuü devices that previously didn't have images and defaulted to white-labelled device images instead.


Koenkk:
It seems one of the jpgs is corrupt, could you check this?

image

davidjb:
@Koenkk Hmm, how strange - thanks for the heads up. The images load okay in browser/image editor but something must be janky.  I've re-saved all the images, rebased on latest master and pushed -- `npm run test` is passing locally now so that should be it.

Koenkk:
thanks!

davidjb:
@Koenkk Thanks for merging. This looks like it's been built and deployed, images like https://www.zigbee2mqtt.io/images/devices/S9E27LED9W-RGB-Z.jpg are being hosted but the supported devices page for Mercator (https://www.zigbee2mqtt.io/supported-devices/#v=Mercator%20Iku%C3%BC) still uses the white-label images.  When I build on my local machine, it looks fine.  Any thoughts?

Koenkk:
@davidjb should be fixed in an hour now.

PR #1198, opened 2022-02-02
Update HG06335.md

Evil95:
Will be added as "Modell HG06335/HG07310". Because of that the sensor has no picture and the model hyperlink goes to https://www.zigbee2mqtt.io/devices/HG06335_HG07310.html#lidl-hg06335%252Fhg07310 instead of https://www.zigbee2mqtt.io/devices/HG06335.html.
Fixed tamper description. False = button pressed = no tamper.


Koenkk:
thanks!

PR #1200, opened 2022-02-02
Added md and image for Stelpro Orléans

MrNickll:
Added image file and markdown file for Stelpro Orléans fan heater.


Koenkk:
thanks!

PR #1208, opened 2022-02-03
Update JT-BZ-01AQ/A

Otnow:


Otnow:
It also looks like _Docgen_ incorrectly generate exposes `linkage_alarm`: _switch (state)_ instead of _linkage_alarm_ and there is no description for it.

Koenkk:
Linkage alarm should be a binary, not a switch. Could you change [this](https://github.com/Koenkk/zigbee-herdsman-converters/blob/208cdbfee9a74682a5b8c0c26bfff9b01230dcac/devices/xiaomi.js#L1177), test and make a pr?

Otnow:
> Linkage alarm should be a binary, not a switch. Could you change [this](https://github.com/Koenkk/zigbee-herdsman-converters/blob/208cdbfee9a74682a5b8c0c26bfff9b01230dcac/devices/xiaomi.js#L1177), test and make a pr?

Ok, I'll try to fix it.

PR #1215, opened 2022-02-06
add pairing info for TS0043 which doesn't have a pairing button on th…

RubenKelevra:
…e PCB


RubenKelevra:
Here's an image of the PCB:

IMG_20220206_165706

Koenkk:
thanks!

PR #1227, opened 2022-02-10
Update Schneider Electric 550D6001 with information about ELKO Smart range

evest:
Added note about the use of this device in the ELKO Smart EKO07117 SmartSwitch.


Koenkk:
thanks!

evest:
My pleasure! 

After creating the PR, I found [other Schneiders](https://www.zigbee2mqtt.io/devices/CCT5010-0001.html) had "White-label" settings as well.

Don't know where that is supposed to be changed, but this should probably be added for better search results:
`| White-label | ELKO EKO07117 |`

PR #1246, opened 2022-02-15
Xiaomi OTA retrieval

Koenkk:
@zvldz please check. I have the first generation Xiaomi gateway (DGNWG02LM); I paired my Xiaomi plug but it doesn't have a firmware update option. Can I assume OTA updates are only supported through Xiaomi gateway 3 (ZNDMWG03LM/ZNDMWG02LM)

image

image

### DIFF
--- a/docs/advanced/more/tuya_xiaomi_ota_url.md
+++ b/docs/advanced/more/tuya_xiaomi_ota_url.md
@@ -33,10 +33,13 @@ Start Android Studio and create a new empty project:
 ![](../../images/tuya_ota.png)
 
 ## Step 3b: get the Xiaomi OTA url
-NOTE: This only works for Xiaomi gateways (e.g. ZNDMWG03LM works, DGNWG02LM doesn't work) which can be connected to the Aqara home app (not the Xiaomi home app!)
+NOTE: This only works for native Aqara gateways which can be connected to the Aqara Home app (not the Xiaomi home app!)
+- Works: ZHWG16LM/HE1-G0, ZHWG15LM, ZHWG12LM
+- Doesn't work: ZNDMWG03LM, DGNWG02LM
+
 1. Download the Aqara home app apk from [this link](https://apkpure.com/nl/aqara-home/com.lumiunited.aqarahome), tested with version 2.2.5. Drag the downloaded apk on the emulator (which will install it).
 1. Open the Aqara home app in the emulator, log in with your account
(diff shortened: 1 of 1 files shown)


zvldz:
Not only via Xiaomi gateway 3.
I tried it via Aqara e1 hub as well.
I think other new gateways should also work - m1s, m2. I can't say anything about DGNWG02LM, but it may not fully support new devices.

Xiaomi gateway 3 does not work with Aqara Home.
And the new firmware is mostly only available in Aqara Home, not in MiHome.

zvldz:
Almost all zigbee devices for Xiaomi were made by Aqara(Lumi)

Koenkk:
@zvldz so do I understand correctly that this works for all except DGNWG01LM and DGNWG02LM gateways? If yes I will update the guide with this.

zvldz:
@Koenkk 
There's no point in catching firmware via the MiHome app. There will be no new firmware in 90%. Xiaomi rarely puts out an update via its app.  Xioami and Aqara have stopped working together. New Aqara devices no longer work with the MiHome app at all.
You MUST use the Aqara Home app to get both new and old firmware.
Most Xiaomi/Aqara devices work with the Aqara Home app (there are a few exceptions)
Aqara Home needs its native gateway (not Mijia).
For example, the gateways are Aqara Hub E1 CN/EU (recommended), M2, M1S, G3, etc.

Koenkk:
> There's no point in catching firmware via the MiHome app. There will be no new firmware in 90%. Xiaomi rarely puts out an update via its app.

The thing is that it even doesn't show a check for firmware update button. Hence my suspicion that upgrades aren't supported at all via DGNWG01LM and DGNWG02LM

zvldz:
I have Xiaomi Gateway 3(ZNDMWG03LM) in MiHome also do not have this button. This is a limitation of MiHome. I updated via the Aqara Home app.

Koenkk:
Thanks, I've updated the guide, can you do a final check before merge?

zvldz:
> NOTE: This only works for Xiaomi gateways (e.g. ZNDMWG03LM works, DGNWG02LM doesn't work) which can be connected to the Aqara home app (not the Xiaomi home app!)

replace to:
NOTE: This only works for native Aqara gateways (e.g. ZHWG16LM/HE1-G01, ZHWG15LM, ZHWG12LM works, ZNDMWG03LM/DGNWG02LM doesn't work) which can be connected to the Aqara Home app (not the Xiaomi home app!)

Koenkk:
updated, thanks!

PR #1259, opened 2022-02-22
[DOCS] Fix dashes that should be underscores in action

deviantintegral:
As we can see, the actual published values from the Philips 324131092621 remote use underscores:

Screen Shot 2022-02-22 at 8 06 43 AM


Koenkk:
thanks!

PR #1262, opened 2022-02-22
Update AC26940_AC31266.md

KopfKrieg:
It seems that Ledvance uses the same identifier for multiple devices. I've got an Ledvance Smart+ Nightlight Plug which identifies as LEDVANCE AC26940/AC31266. I've updated the device file (not sure if that's the correct way though, due to the warning that this page was auto-generated (but I can't find the source)).

The referenced image should maybe also be changed.

Here's the link: https://lb.ledvance.de/consumer/produkte/smart-home/smart-home-produkte-mit-zigbee-technologie/smart-home-komponenten/smart-nightlight-plug/index.jsp

### DIFF
--- a/docs/devices/AC26940_AC31266.md
+++ b/docs/devices/AC26940_AC31266.md
@@ -17,7 +17,7 @@ pageClass: device-page
 |-----|-----|
 | Model | AC26940/AC31266  |
 | Vendor  | LEDVANCE  |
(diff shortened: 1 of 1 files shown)


KopfKrieg:
Whoops, sorry, didn't see your comment. I'll try to update the branch and then report back.

KopfKrieg:
Don't close it yet - I just didn't have time.

PR #1269, opened 2022-02-27
Add Waxman 8850100/leakSMART Valve 2.0

kevin-david:
Not sure if this is going to "work" yet or I need to manually add the other sections?

Converter added here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/3933

I think that needs to be merged first, but wanted to get a skeleton up.


docs/devices/8850100.md
```diff
@@ -0,0 +1,25 @@
+<!-- !!!! -->
```
kevin-david, line 1: As I mentioned in https://github.com/Koenkk/zigbee-herdsman-converters/pull/3933/files#r815450403, Not sure what the naming convention should be here given my verified test with one model but assumed compatibility with other models. Happy to make whatever changes are necessary.

Koenkk:
thanks!

PR #1273, opened 2022-03-01
Update README.md

SvenCSchmidt:
Changed 'ConBee II' to 'ConBee II / RaspBee II' because unexperienced user maybe don't know this option applies for both adapters.


Koenkk:
thanks!

PR #1281, opened 2022-03-11
Update F-MLT-US-2.md

amaisano:
This tip works for the older unit as well. Saved me a LOT of trouble. You really need to spam the contact magnet on/off while pairing for all the sub-sensors to get picked up.


Koenkk:
Thanks!

PR #1291, opened 2022-03-18
Extra clarification for adding new devices

JanJaapKo:
Some things that I would have found helpful to understand what to do.


docs/advanced/support-new-devices/01_support_new_devices.md
```diff
@@ -96,9 +97,9 @@ const definition = {
     model: 'WSDCGQ01LM',
     vendor: 'Xiaomi',
     description: 'MiJia temperature & humidity sensor',
-    fromZigbee: [fz.temperature], // <-- added here
-    toZigbee: [],
-    exposes: [e.battery(), e.temperature(), e.humidity()],
+    fromZigbee: [fz.temperature], // <-- added here all clusters reported from zigbee
+    toZigbee: [], // <-- add here all clusters to send commands to zigbee
+    exposes: [e.battery(), e.temperature(), e.humidity()], // <-- this will define all fields published to MQTT on the device's topic
```
Koenkk, line 102: > this will define all fields published to MQTT on the device's topic

What is published to MQTT is defined by fromZigbee, `exposes` defines the things exposed to the frontend/Home Assistant

JanJaapKo: @Koenkk the elements in the exposes array are then populating some message at startup? I'm integrating with Domoticz so the pointers to Home Assistant put me of at first... So Domoticz Zigbee2mqtt plugin also takes the initial message to create the widgets for its front end. What's the name of this message?

Koenkk: Yes the domiticz plugin also users this info, so maybe the following is bettery:

```
// <-- this will define all fields exposed by this device (e.g. frontend, Home Assistant, Domoticz)
```

JanJaapKo: right, but it is not so much as exposing the actual measurements, it is about exposing the definition on the 'zigbee2mqtt/bridge/devices' topic, correct? As you mentioned in your earlier comment, the fromZigbee/toZigbee parameters also define what goes onto MQTT. So perhaps a phrase like this:
`// <-- this will define which fields will be exposed in the definition message to configure any front end (e.g. to the frontend, Home Assistant, Domoticz)`

Koenkk: > right, but it is not so much as exposing the actual measurements, it is about exposing the definition on the 'zigbee2mqtt/bridge/devices' topic, correct?

Exactly

> So perhaps a phrase like this:

fine by me

JanJaapKo:
Well then hopefully this will provide a bit more guidance for new users....

Koenkk:
thanks!

PR #1300, opened 2022-03-26
Added TS0202_2 support

SignalStealer:
Another variant of TS0202 (reported & in print at the bottom of the device), slightly different than the existing 2.


Koenkk:
Thanks!

PR #1301, opened 2022-03-27
Document changed BW-SHP13 and broken power reporting of that new revision

nohn:
Firmware (maybe Hardware) für BW-SHP13 has changed in early 2022 and has broken power reporting. See https://github.com/Koenkk/zigbee2mqtt/issues/11800 for details


Koenkk:
thanks!

PR #1308, opened 2022-04-01
Update installation/01_linux.md: add RestartSec to systemd service

ferbar:
zigbee2mqtt.service: add restart delay, prevents that a raspberry pi gets hot in case of an error


Koenkk:
Thanks!

PR #1309, opened 2022-04-01
Added a markdown file and image for the 5060930P7 Philips Hue Centris…

marblecoma:
… 3 spot black lamp. The markdown file is a copy of the 5060931P7 markdown file with all instances of 5060931P7 replaced with 5060930P7.


Koenkk:
thanks!

PR #1310, opened 2022-04-01
Fix incorrect markdown in Niko Switch pages

sjorge:
Looks like without the header the table does not render correctly.


Koenkk:
thanks!

PR #1315, opened 2022-04-04
Correct indentation for external_converters

adyjoy:
Z2M wouldn't start without 'external_converters' being indented

### DIFF
--- a/docs/advanced/support-new-devices/01_support_new_devices.md
+++ b/docs/advanced/support-new-devices/01_support_new_devices.md
@@ -51,8 +51,8 @@ Now set the Zigbee2MQTT `log_level` to `debug` and enable the external converter
 ```yaml
 advanced:
   log_level: debug
-external_converters:
-  - WSDCGQ01LM.js
+  external_converters:
+    - WSDCGQ01LM.js
 ```
 
 Once finished, restart Zigbee2MQTT and trigger some actions on the device. You will see messages like:


Koenkk:
The current indentation should be ok, can you provide the error message?

adyjoy:
Hi, 

in the guide under "2. Adding your device" it shows the syntax for adding the 'external_converters' option to be:

```
advanced:
  log_level: debug
external_converters:
  - WSDCGQ01LM.js
``` 
with external_converters at the beginning of the line, and therefore not part of the 'advanced' options. When I added it like this for a dev branch converter I needed, Z2M would not start up and kept rebooting every few seconds, each time leaving a zero-byte log file (this is with logging set to 'debug'). 

I then changed the indent to 
```
advanced:
  log_level: debug
  external_converters:
    - skydance.js
```
so that external_converters and the referenced file were indented and part of 'advanced', and Z2M started and worked correctly .

Koenkk:
> so that external_converters and the referenced file were indented and part of 'advanced', and Z2M started and worked correctly .

Probably something is wrong with your external converter, if you define it under advanced no external converter will be loaded

adyjoy:
> > so that external_converters and the referenced file were indented and part of 'advanced', and Z2M started and worked correctly .
> 
> Probably something is wrong with your external converter, if you define it under advanced no external converter will be loaded

Ok, thanks for looking in to it.

PR #1321, opened 2022-04-09
add beok thermostat TGR-85-ZB

nicoxygen:
for new device support 

[New device support]: Beok TGR-85-ZB #11707


Koenkk:
Please also create a pr for https://github.com/Koenkk/zigbee-herdsman-converters

PR #1326, opened 2022-04-10
Adapters config notes

mdziekon:
During my latest migration to Z2M I've struggled a bit with the initial configuration of the bridge, which later on turned out to be a simple missing param for my ConBee II adapter. However, it was easier to find a solution for this by going through various HA-specific problem threads on their forum, rather to find that being mentioned in this documentation.

I think it wouldn't hurt to mention it a bit more prominently in the "Getting started" and "Installation" sections, to let users know that besides the usual config, there might be this or that parameter required for their specific adapter. Therefore, let's recommend that they should venture to the `Supported Hardware > Adapters` section to find that out before they run into any troubles (or at least provide a quick guidance if they already did). I've also added a mention about adapter's specific config in "Zigbee2MQTT fails to start" section.

In addition to that, I took the liberty of mentioning the recommended way of configuring adapter's device path using the "by ID" mapping. The recommendation was already there, in the "Configuration > Adapter settings" section, but I think it would be better to mention it right away during the installation process, to promote the best practice right from the start, rather than an afterthought.


Koenkk:
Thanks!

PR #1327, opened 2022-04-11
Update PTM_216Z.md

wb441:
I received confirmation from Ubisys that their
D1
S2
C4
... modules support direct binding from PTM216Z modules (wire- & batteryless switch) to their modules.
I plan on testing it myself.

So far, I had to use an external tool like NodeRed to make these batteryless switches do stuff. (which works great!)
But for many applications, a direct link would not need that extra service running & could be preferable reliability-wise.


Koenkk:
Thanks!

PR #1330, opened 2022-04-15
Update RTCGQ11LM.md

danstreeter:
Added additional potential pairing option for when the standard sync is not working.


Koenkk:
Thanks!

PR #1332, opened 2022-04-18
Support for Z2M-setup now integrated in openhabian

Larsen-Locke:
short update of documentation:
- Setup of Z2M ist now integrated in openHABian
- link to examples is updated


Koenkk:
Thanks!

PR #1334, opened 2022-04-19
Add JavaScript heap out of memory to common start failures

Impact123:
I'm a bit conflicted if this is common/important enough to be listed here. You be the judge.

### DIFF
--- a/docs/guide/installation/20_zigbee2mqtt-fails-to-start.md
+++ b/docs/guide/installation/20_zigbee2mqtt-fails-to-start.md
@@ -122,6 +122,18 @@ In case you see message like below when running `dmesg -w` you are using a bad p
 
 When you have a SSD connected to the Pi, try connecting the adapter via a powered USB hub.
 
+## Error: JavaScript heap out of memory
+On memory constrained deviced it might be necessary to run 
+```bash
+export NODE_OPTIONS=--max_old_space_size=256
+```
+before executing `npm start`.
+
+It can also be run in one line like this
+```bash
+NODE_OPTIONS=--max_old_space_size=256 npm start
+```
+
 ## Make sure the extension cable works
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Just pushed a fix for this, this option will automatically be added on systems with < 600MB ram: https://github.com/Koenkk/zigbee2mqtt/commit/ec2bf1a95d6c45c093f68df832bc6bcbdd372a5a

Could you test this? For this you will need to switch to the dev branch and execute: `rm -rf dist && npm start`  (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html)

Impact123:
I don't actually have this problem myself. I'll test this in a small VM tomorrow if I can remember. Fixing the problem in code instead of documenting a workaround definitely sounds like a better solution. The only suggestion I have is to add/concat the option to the variable instead of potentially overwriting it.

Koenkk:
> The only suggestion I have is to add/concat the option to the variable instead of potentially overwriting it.

Done

Impact123:
I'm sorry but I couldn't reproduce this with the `master` branch on my 500MB VM. If I disable swap then `npm ci` gets killed. If I disable swap only after having successfully ran `npm ci` then `npm start` takes a minute or so and then gets killed too. I couldn't produce the `JavaScript heap out of memory` error. I tested with a basic headless `Debian 11` install. I don't own a Raspberry PI. I also tried to test this in docker because why not and apparently `os.totalmem()` doesn't work with it. At least it doesn't give reliable results. Maybe something to keep in mind.

As for the concatenation, It's a little verbose but I thought about something like this
```js
if (env.NODE_OPTIONS) {
    if (!env.NODE_OPTIONS.includes("--max_old_space_size")) {
        env.NODE_OPTIONS += " --max_old_space_size=256";
    }
} else {
    env.NODE_OPTIONS = "--max_old_space_size=256";
}
```

PR #1342, opened 2022-04-25
Updated Innr SP110 documentation

Arkoxs:
Forgot to update the link to the new image name and added legacy documentation received from the manufacturer. 
Since this manual is not online available I've added it under ./public/manuals/devices/SP_110.pdf


Koenkk:
Thanks!

PR #1346, opened 2022-04-26
Fix Inconsistencies in container setup (name & restart policy)

Secretowo:
The docker run commands didnt include a name and restart policy, while the docker compose file did.


Koenkk:
Thanks!

PR #1348, opened 2022-04-30
Update TS011F_din_smart_relay.md

henkiejan1:
Added isntruction for reset sum of consumed energy. Tested by myself with this relay and works. Think it works for all TS011F devices.


henkiejan1:
Closed i disover the relay doesn't work right now

PR #1349, opened 2022-05-01
Philips 929003055501 support

marblecoma:
…ee model number for the same device. I have adapted the existing support for the new model number.


Koenkk:
Thanks!

PR #1351, opened 2022-05-01
Update E1757.md

martinmunk:
Added a little more detail on how to set the end position of the blind, so others won't waste time trying to do it with the remote as I did.


martinmunk:
Sorry, I should probably have changed the default headline to include a little more detail here.

Koenkk:
Thanks!

PR #1353, opened 2022-05-02
Update deconz adapter information

hhromic:
* Mention ConBee and RaspBee as they are also supported.
  * Source: I own a RaspBee and works perfectly with Zigbee2MQTT.
* Update firmware link, buying links and general wording.

Note: While I don't own a ConBee, it uses the same chipset as the RaspBee and therefore it _should_ also be compatible.

### DIFF
--- a/docs/guide/adapters/README.md
+++ b/docs/guide/adapters/README.md
@@ -280,7 +280,7 @@ _(in order of first appearance)_
   * <details>
     <summary>ConBee / ConBee II / RaspBee / RaspBee II</summary>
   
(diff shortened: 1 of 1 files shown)


Koenkk:
@chrishae can you confirm the first version of the raspbee and conbee are also supported?

hhromic:
@Koenkk for what is worth, zigbee2mqtt actually reports my RaspBee as both "ConBee/RaspBee" :)

image

ChrisHae:
ConBee (1) and RaspBee (1) are also supported. They use the same chip. And the firmware version which they expose to zigbee2mqtt has the same byte for that chip. So they can't be distinguished by the firmware version.

Koenkk:
Thanks guys! 👍

PR #1354, opened 2022-05-02
Add documentation for MiBoxer FUT089Z

lorenz:
This is quite an unusual device, add documentation on how to use it.


Koenkk:
Thanks!

PR #1363, opened 2022-05-04
Update README.md

Hedda:
Change ZiGate and Silicon Labs comments to be consistent with that about experimental (alpha stage) support.


Koenkk:
Thanks!

PR #1368, opened 2022-05-06
Update KMPCIL_RES005.md

schasj:
Through experimentation I found that the state commands must be lowercase to be effective.

The device's Exposes page has the state switch commands in uppercase and fails because of it. That needs to be changed too.


Koenkk:
Fixed this now (so uppercase state will also work now).

Changes will be available in the dev branch in a few hours from now. (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html)

PR #1371, opened 2022-05-06
Update WRS-R02.md

Hugo-Persson:
Added notes for the Aqara double rocker remote. Info added about pairing and how to fix issues with changing click mode with the device.


Koenkk:
Thanks!

PR #1373, opened 2022-05-08
Add adapter type to documentation

codewise-nicolas:
Add adapter type to the sample code to be Zstack.
(I kept switching from a ezsp and a zstack router for testing and without the adapter being set to zstack it would default to the last value)


Koenkk:
Thanks!

PR #1379, opened 2022-05-13
Add possible solution for Xiami device dropping from the network.

MichaelMure:
Personally tested with WXKG11LM and MCCGQ11LM that are now stable for 24h while they were dropping after a few minutes before. I updated all the devices that mentionned the same issue.


Koenkk:
Thanks!

PR #1380, opened 2022-05-13
Update CCTFR6700.md

esseki:
Clarify how one should update external temperature measurement for this device, from coordinator through an automation


Koenkk:
Thanks!

PR #1390, opened 2022-05-19
Interference from other 2.4 GHz device

antoinelibert:
Hello,

Small addition to the doc about interferences from 2.4 GHz devices. I have just experienced such interferences from my newly bought wireless gaming mouse from Razer (ZB devices down around the room where the mouse is located).


Koenkk:
> newly bought wireless gaming mouse from Razer (ZB devices down around the room where the mouse is located).

Wow! Well found, thanks for the addition.

PR #1392, opened 2022-05-21
Rename Bticino to Legrand with Model Code 

SilentT-FR:
Hi 

I do not know how it should be done to make it cleaner and that it works with BTicino or Legrand

At the same time, it would also be necessary to make the modifications in the converters


Koenkk:
Please also make a pr for https://github.com/koenkk/zigbee-herdsman-converters, the docs you changed are automatically generated from it.

Koenkk:
Could you update the pr to rename both files, also keep the old ones otherwise the tests will fail. I will clean this up with the next release.

Koenkk:
Thanks!

PR #1393, opened 2022-05-21
Update 01_linux.md - user

deeja:


docs/guide/installation/01_linux.md
```diff
@@ -11,6 +11,9 @@ Before starting make sure you have an MQTT broker installed on your system.
 There are many tutorials available on how to do this, [example](https://randomnerdtutorials.com/how-to-install-mosquitto-broker-on-raspberry-pi/).
 Mosquitto is the recommended MQTT broker but others should also work fine.
 
+## Executing user 
+The user `pi` is used the following examples, but the user may differ between dists. e.g. `openhabian` should be used on Openhabian.
```
Koenkk, line 15: It think it would be better to move this line after:

> For the sake of simplicity this guide assumes running on a Raspberry Pi 4 with Raspbian Stretch Lite, but it should work on any Linux machine.

> Therefore the user `pi` is used the following examples, but the user may differ between distributions e.g. `openhabian` should be used on Openhabian.

deeja:
@Koenkk  updated

PR #1399, opened 2022-05-26
Typo in Config Doc

somtooo:
"It's possible specify a custom data directory by setting the ZIGBEE2MQTT_DATA environment variable." seems like It's meant to be if.


Impact123:
> It's possible **to** specify

would make more sense.

PR #1403, opened 2022-05-26
Update npm version in python installation

watercrossing:
The instructions are still on 10.15.1, which is ancient. The [docker builds](https://github.com/Koenkk/zigbee2mqtt/blob/master/docker/Dockerfile) are based on 16.15.0, which is also old... but less old. This PR brings those two versions in line. At least for now.


Koenkk:
Thanks!

PR #1407, opened 2022-05-29
add clarification to ubisys C4 raw config docs

patmalcolm91:
Add a clarification to the raw configuration documentation for the Ubisys C4. Specifically, spelling out the fact that each array element corresponds to a single byte, rather than a field (which can be multiple bytes long, and around which the technical reference from Ubisys is oriented).

In my opinion, also adding an explanation of each element of one of the arrays from the example configuration would be very helpful, though I didn't add this yet, since I wasn't sure how verbose the documentation should be, or if it's preferred that users rely more heavily on the manufacturer's technical reference and use the z2m documentation more as a hint. If such an example is desired, I'd be happy to add one.


Koenkk:
Thanks!

PR #1410, opened 2022-05-30
Perenio PEHPL0X and PEHWE20 images added

annar-dev:
The Perenio Power Link PEHPL0X and Perenio Mini-relay PEHWE20 images added


Koenkk:
Thanks!

PR #1421, opened 2022-06-06
Update home_assistant.md to new MQTT definition

andrew-codechimp:
From HA release 2022.6, MQTT sensors and switches should be updated to the new format and the old method is deprecated, due to be removed in 2022.9
See https://www.home-assistant.io/blog/2022/06/01/release-20226/#breaking-changes


Koenkk:
Thanks!

PR #1427, opened 2022-06-08
Remove information from docs; integration

MeanderingWanderer:
Why should this be removed?
integration broken for HA >= 2021.12

Problem detailed here
https://github.com/rgruebel/ha_zigbee2mqtt_networkmap/issues/42
and here,  https://github.com/hacs/integration/issues/2517


Koenkk:
Thanks!

PR #1456, opened 2022-06-28
Update of sniffer software

PietPetrus:
Previous link was not referring to ZBOSS for Windows.
Corrected it. The default path to the hex file was also changed.


Koenkk:
Thanks!

PR #1467, opened 2022-07-02
Add picture for TuYa Temperature and humidity sensor TS0601

alxbator:
Temperature and humidity sensor TS0601  _TZE200_qyflbnbj


alxbator:
I'll cancel this PR due to reusing an already existing converter

PR #1494, opened 2022-07-25
add Plugwise Lisa

vhdirk:


Koenkk:
Thanks, could you also make a pr for zigbee-herdsman-converters?

vhdirk:
Yes it is :) They (Plugwise) implemented embarrassingly little, so there's not a lot to be done here either.

Koenkk:
Thanks!

PR #1503, opened 2022-08-01
Add Ubisys H1

sjorge:


sjorge:
Also fixed the shadows on the image, looks like we both grabbed the same one but the shadow on yours got messed up while adding the white background.

sjorge:
I don't think the test failure is related to this change?

Koenkk:
Thanks!

PR #1505, opened 2022-08-07
The bridge state is not json, so don't treat it as such

nateww:
The data for zigbee2mqtt/bridge/state is one of `online/offline`, and not JSON.  With the example setup, if you include the value_template, the logs will be filled with errors such as:
```
ERROR (MainThread) [homeassistant.helpers.template] Template variable error: 'value_json' is undefined when rendering '{{ value_json.state }}'
```

Removing value_template causes the error to go away AND the sensor contains the proper (text) value.


Koenkk:
Thanks!

PR #1510, opened 2022-08-12
Add TubeZB Silabs EFR32 network adapter to list

Hedda:
Update supported adapters README.md adding add TubeZB Silabs EFR32 Hybrid (network + USB) adapter by @tube0013 to list in preparation for when Silicon Labs EZSP will no longer be experimental in Zigbee2MQTT, as discussed here -> https://github.com/Koenkk/zigbee2mqtt/discussions/13373

In addition, added the CoolKit ZB-GW04 USB dongle (a.k.a. easyiot stick) with a warning as becoming more common because it is currently the cheapest Silabs EFR32 USB adapter available but there is also a reason for that so users should note that before buying.

Also added a comment about ITead now selling both a “ZBDongle-E” (EFR32MG) adapter and a “ZBDongle-P” (CC2652)  adapter.


Koenkk:
Thanks!

PR #1523, opened 2022-08-24
Updated device icons

T-IOOO:
Philips Hue Go: Official device icon

Philips Hue Ensis: Correct device icon. The one that is displaying now is wrong


Koenkk:
Thanks!

PR #1526, opened 2022-08-25
Update WXKG01LM.md

Marcossi:
The current instructions in the documentation for pairing did not work for me. I don't know if they are wrong or if the method changes depending on the firmware.

I have seen a second option described, but I think it is incomplete (or poorly explained). I have tried to complete it as it was the method that worked for me.


Koenkk:
Thanks!

PR #1533, opened 2022-08-27
Add device iH-F001

kapabel:
Hello. The TuYa iH-F001 door sensor isn’t listed in the supported devices section. It gets recognised as TS0203 which is similar (or is probably the same anyway). Everything works just fine.


Koenkk:
Added it as a whitelabel, thanks!

PR #1542, opened 2022-09-01
Add pairing information to Philips 929003017102

guedressel:
Since I struggled some time to find a way to pair my new Philips Hue Wall switch modules I was happy once I learned about its mechanism.  
Please add this missing information to the documentation.


Koenkk:
Thanks!

PR #1543, opened 2022-09-02
STOFTMOLN ceiling/wall lamp WW37

mastertape:
My first try to support the community - it the new Ikea STOFTMOLN ceiling/wall lamp WW37 bouhgt in germany


Koenkk:
Thanks!

PR #1545, opened 2022-09-03
Update TS0601_din.md

evlo:
adden note about how total energy consumed is reported to avoid potential confussion


Koenkk:
I've moved it to the notes section since the exposes part is automatically generated, thanks!

PR #1551, opened 2022-09-05
Update README.md

wizbowes:
Changed all instances of repair to re-pair for clarity except for one instance which looks like a link so I left it alone.


Koenkk:
Thanks!

PR #1552, opened 2022-09-06
Name corrected 

dhanukapulkit:
The model number was wrong; 
CLA 60 DIM was changed to A60 DIM T
Thanks @Koenkk for pointing it out!


Koenkk:
Thanks!

PR #1558, opened 2022-09-07
Update 02_docker.md

DubhAd:
Guide people to use the `/dev/serial/by-id/` path, to reduce the problems that arise when paths change due to having multiple USB sticks (whether more than one Zigbee coordinator, or a Z-Wave radio, or a USB GPS, or ...)


Koenkk:
Thanks!

PR #1562, opened 2022-09-11
Update PTM_215Z.md

pow4all:
Added Feller push-botton (Switzerland) as white-label product


Koenkk:
What is the model number of this device?

pow4all:
Thanks for checking. Inside the Feller push-button the EnOcean PTM_215Z (GreenPower) module is used. So there is no specific model number (an order number is not really what we are looking for...).
If you search for "Feller" on zigbee2mqtt you will never find this information. That was my intention to add this information.

pow4all:
https://online-katalog.feller.ch/kat_details.php?fnr=4120.2.S.FMI.61

pow4all:
OK, I found a model number :)

Koenkk:
Thanks!

PR #1569, opened 2022-09-12
Added page with image of IKEA TRADFRI LED2009C3

ima-quack:
IKEA retails an additional model to their E12 candle light 250 lumen, however it is labeled as model LED2009C3 rather than the already supported LED1935C3.


Koenkk:
Thanks!

ima-quack:
Hi! Just wondering why the markdown file was deleted, and if I should rewrite the description so that the page can be added?

Koenkk:
The markdown file will be automatically generated on the next release.

PR #1572, opened 2022-09-13
43113.jpg

majorsl:
Added 43113.jpg which is the updated version of the 43080 (same pic, nothing changed, no model number shown.)


Koenkk:
Thanks!

PR #1576, opened 2022-09-15
Notes for TS0207_water_leak_detector

JohnMcLear:


Koenkk:
Is there already a z2m issue open about this? It may be helpful to catch a sniff when the device leaves: https://www.zigbee2mqtt.io/advanced/zigbee/04_sniff_zigbee_traffic.html#with-cc2531

PR #1577, opened 2022-09-15
Update TS011F_plug_1.md

mkotek:
Added Modemix models (Polish local seller branded items) available at MediaExpert stores, to find them more easily adn confirm they are working as expected.


Koenkk:
Thanks!

PR #1580, opened 2022-09-16
Add warning to Aqara Plug (SP-EUC01) documentation

0ip:
Until a solution to the firmware bug is found, users of this device should be warned of the problem and offered a solution.


Koenkk:
Thanks!

PR #1582, opened 2022-09-16
Update 14153806L.md

VincentSC:
Added pairing info. Tested myself. I used text from https://www.zigbee2mqtt.io/devices/HG06104A.html but made a textual change by switching `on` and `off`, as that would be more correct.


Koenkk:
Thanks!

PR #1589, opened 2022-09-19
:books: add pairing notes

rediculum:
- Added pairing notes for Moes MS-105Z and ZS-EUD
- Replace Image for ZS-EUD


Koenkk:
Thanks!

PR #1596, opened 2022-09-22
Update Philips Hue wall switch actions

Lakerfield:
When pressing and holding the switch buttons there are 4 more `hold` -actions send


Koenkk:
Thanks!

PR #1601, opened 2022-09-23
Remove unsupported actions

adorobis:
The device does not support hold and release actions therefore should be removed from the description


Koenkk:
Thanks!

PR #1606, opened 2022-09-25
Update 701721.md

TomTonic:
Viewing direction documentation is wrong. There is no horizontal mode on this device, both modes are vertical (I own the device and I tryed). ‚False‘ is normal and ‚true‘ is upside-down.


Koenkk:
Thanks!

PR #1617, opened 2022-10-01
Update DDS238-2.md

gericho:
Here the "beta" external converter, it shows also power factor, reactive energy, switch control, and AC frequency, need to update documentation too, I made some changes on the image and model version. Unfortunately it still show also voltage V and current A in the tuya app, but no datapoints, the only one that seems to be encrypted is Phase A, someone knows how to unencrypt it? Currently no firmware updates are present on tuya app.

[[](url)](https://gist.github.com/Koenkk/34e60391471f8a6333176594dd80445d)

### DIFF
--- a/docs/devices/DDS238-2.md
+++ b/docs/devices/DDS238-2.md
@@ -15,11 +15,11 @@ pageClass: device-page
 
 |     |     |
 |-----|-----|
-| Model | DDS238-2  |
+| Model | DDS238-2 Ver.IVAP |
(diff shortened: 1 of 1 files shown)


Koenkk:
See my latest comment for the voltage https://gist.github.com/Koenkk/34e60391471f8a6333176594dd80445d?permalink_comment_id=4321384#gistcomment-4321384

There is no need to update just the md file, it will be updated automatically with the next z2m release.

PR #1619, opened 2022-10-03
Update scenes.md

WhimsySpoon:
Details on how to create a scene with a transition were not immediately clear to me.


Koenkk:
Good addition, thanks!

WhimsySpoon:
Thanks, @Koenkk. Is my assumption correct that a transition cannot be supplied in `scene_store`? Because this would make things a lot easier, especially if we could add this to the front end.

Koenkk:
It is, from the spec:

image

WhimsySpoon:
Great! Is that honoured in z2m? So could I send:

```
{
    "scene_store": {
        "ID": 0,
        "name": "Scene Name",
        "transition": 5
    }
}
```

... and it would work? I'm sure I tried this in the past without any luck.

PR #1621, opened 2022-10-05
Update HR-C99C-Z-C045.md

seb49:
Add some detail on the utility of the center of circular button. Its the ways to access some action


Koenkk:
Thanks!

PR #1623, opened 2022-10-07
Small correction.

vladi1234:


Koenkk:
There is no need to manually update this, it is generated if the `setAccess` is set correctly in the exposes section of the definition.

PR #1624, opened 2022-10-07
Update E2001_E2002.md

MasterOfBytes:
Added an detailed Action Table that describes which action is triggered on which button.


Koenkk:
Thanks!

PR #1637, opened 2022-10-15
Documentation: ICPSHC24-30EU-IL-1 - troubleshooting tips for pairing problems

MarvAmBass:
It seems to affect multiple users, I've checkt all suggestions in the bug and this was the only working way. https://github.com/Koenkk/zigbee2mqtt/issues/611#issuecomment-1206404155

Tested with latest models straight out of ikea, and with devices I had at home for almost 2 years.

I was't able to pair any of those devices without this workaround/tips.

It was frustrating, that the wiki didn't have more tips than hold the reset button - which seems to to more of an reset instead of helping with pairing

### DIFF
--- a/docs/devices/ICPSHC24-30EU-IL-1.md
+++ b/docs/devices/ICPSHC24-30EU-IL-1.md
@@ -30,6 +30,16 @@ pageClass: device-page
 To factory reset the TRADFRI drivers use a
 small pin or paperclip to push the reset button once.
 NB: you HAVE to put the device VERY close to the zigbee device, otherwise it won't be able to detect.
+
+If you struggle to pair this device, the following procedure should help getting it paired:
+
+1. Have the driver disconnected from power
+2. Enable join in your zigbee2mqtt server
+3. Connect the driver to the power, pairing begins immediately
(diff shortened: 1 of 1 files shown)


MarvAmBass:
still not consistently pairable...

MarvAmBass:
Hi Koenkk,

I've added a some kind of troubleshooting guide within the `pairing` section - not sure if this is the best place for it.
Maybe you want to refactor it into some kind of it's own `troubleshooting section`

I don't think the issue https://github.com/Koenkk/zigbee2mqtt/issues/611 is closed - might not be a problem for zigbee2mqtt, but for users it should be still hard to pair - even with latest devices from ikea
I've tried it with 3 from 2020 and 3 from 2022 (bought yesterday) and all had the same problems

If there were some infos in the wiki I wouldn't have bought the last 3 devices and could have just continued with the old one I had at home

would be nice if you could merge this, or enhance the documentation.

thanks and kind regrads

Marvin

Koenkk:
Could you dive a bit deeper here and provide a sniff when putting the device into pairing mode? https://www.zigbee2mqtt.io/advanced/zigbee/04_sniff_zigbee_traffic.html#with-cc2531

PR #1663, opened 2022-10-28
Update TS0203.md to include pairing instructions.

brewston:
Taken from https://zigbee.blakadder.com/Tuya_iH-F001.html but I verified that they are correct.


Koenkk:
Thanks!

PR #1664, opened 2022-10-28
Update IH-K009.md

brewston:
Taken from https://zigbee.blakadder.com/Tuya_IH-K009.html but I verified that they are correct.


Koenkk:
Thanks!

PR #1665, opened 2022-10-28
Update ID-UK21FW09.md

realroywalker:
Add instructions on how to reset the device so it can re-paired to a new network, as these are missing from the user manual for the device, so makes re-pairing a bit of a guessing game!


Koenkk:
Thanks!

PR #1671, opened 2022-11-02
Update systemd instructions

Nardol:
* Add NODE_ENV environment variable to run in production mode
* Add instruction to set a custom data directory


Koenkk:
Thanks!

PR #1676, opened 2022-11-03
Added pairing instructions

mdehaas:
Almost re-paired all devices after changing channel and network key... new network is stable. Hope someone reads this page one day ;-)


Koenkk:
Thanks!

PR #1698, opened 2022-11-12
Update remote adapter guide to include Conbee

domenix:
Conbee II indeed requires a bit different ser2net configuration, both according to an closed issue in the other [repo](https://github.com/Koenkk/zigbee2mqtt/issues/10533) and my tests as well. Z2M fails to even start as an addon in Home Assistant when `nobreak` is left out of the configration.


Koenkk:
Thanks!

PR #1701, opened 2022-11-14
Make it clear that Hue 9290012573A has no BT

r41d:
Source (german): https://www.digitalzimmer.de/artikel/wissen/hue-generation-1-2-3-4-5-6-unterschiede/

You can actually see the exact Model number in this picture really well: 
image


docs/devices/9290012573A.md
```diff
@@ -17,7 +17,7 @@ pageClass: device-page
 |-----|-----|
 | Model | 9290012573A  |
 | Vendor  | Philips  |
-| Description | Hue white and color ambiance E26/E27/E14 |
+| Description | Hue white and color, E27 socket, 9.5 watt power draw, 806 lumen |
```
Koenkk, line 20: I believe not all models use the E27 socket so I propose to leave as is

docs/devices/9290012573A.md
```diff
@@ -48,10 +48,9 @@ With [one](./324131092621.md) of the [two](./929002398602.md) Hue Dimmer switche
 
 See also the [VIDEO: Factory reset a Hue bulb with Hue dimmer switch](https://www.youtube.com/watch?v=qvlEAELiJKs).
 
-#### Bluetooth (if supported by device)
-Install the Philips Hue Bluetooth app for [Android](https://play.google.com/store/apps/details?id=com.signify.hue.blue)
-or [iOS](https://apps.apple.com/us/app/philips-hue-bluetooth/id1456604186). You can use the app to trigger a factory reset on a paired light. (Note: The light will only be in bluetooth pairing
-mode for a couple of minutes after power-on)
+#### Bluetooth (unsupported here)
```
Koenkk, line 51: I think it would be better to remove the bluetooth section entirely.

r41d: Maybe put it in a different place but the hint that BT is definitively not available could be handy so that people don't keep trying

Koenkk: Then I suggest to remove it and put something like: "This device does not support Bluetooth right under" "## Notes"

docs/devices/9290012573A.md
r41d: You may be right, unfortunately this 9290012573A bulb with E27 is the only Hue bulb that I possess.
I primarily thought think this ID is E27-only because 90%+ of all images when searching for it in google image search are of E27 bulbs.

docs/devices/9290012573A.md
r41d: ✓

docs/devices/9290012573A.md
Koenkk: What is the model ID of yours? Can be found in the data/database.db

r41d: I can unfortunately not pair it because zigbee2mqtt crashes every time, I think the culprit here is my Conbee2, but I'm not sure. This MR was just to share my findings that Gen4 Hue bulbs don't have Bluetooth which I still think should be included in the doc

Koenkk:
Thanks!

PR #1706, opened 2022-11-19
Change documentation/supported adapters to add warning about incompatible conbee2 firmware versions

OrbitingOcelot:
Added warning regarding incompatibility with newer firmware for Dresden Electronics Conbee 2 stick


Koenkk:
Thanks!

PR #1711, opened 2022-11-20
Update ZigStar adapters

mercenaruss:
Update ZigStar adapters:
1. Replace old PoE with new version
2. Adding PI HAT shield with POE and RTC


Koenkk:
Thanks!

PR #1718, opened 2022-11-24
Add pairing instructions for the Xiaomi Radiator Thermostat E1

ronaldvdmeer:
Add pairing instructions for the Xiaomi Radiator Thermostat E1
Tested this myself and I can confirm this is the way it works.


Koenkk:
Thanks!

PR #1720, opened 2022-11-24
Update ZNCZ04LM.md

SmartPhoneLover:
I have added an issue that seems to be common, even with the latest FW available, which makes the plug to automatically turn if off by itself when it looses the connection from the coordinator.

Notes: I have 2 of them, and integrated using Sonoff Zigbee 3.0 Dongle Plus in Home Assistant.


Koenkk:
Thanks!

PR #1722, opened 2022-11-25
Added support for the HG08673-EU model

MartinAirN:
Changed FR model to generic supporting both.

This PR is paired with https://github.com/Koenkk/zigbee-herdsman-converters/pull/5020


MartinAirN:
@Koenkk Seeing as the test fails on the -FR image no longer existing, should we duplicate the page and image instead?

Koenkk:
@MartinAirN yes please duplicate it, will cleaup later.

Koenkk:
Thanks!

PR #1723, opened 2022-11-25
Failure for the Current_summation

marcog1973:
From my point of view ther is a failure for the Current_summation. In the technical documentation, which i found here: http://docplayer.net/135845279-Technical-manual-for-zhemi101-zigbee-external-meter-interface-ha.html the range for the CurrentSummation is defined as 0x000000 - 0xFFFFFFF. This should be 0-268435455.
And this makes much more sense as it is more realistic to set initial counter from the electricity meter. So the value of 1000 = 1,0 kWh
10.000 = 10,0 kWH and so on.


Koenkk:
Thanks!

PR #1724, opened 2022-11-26
Updating Reset instructions with timing

stefancrain:
Source : https://byjasco.com/mwdownloads/download/link/id/1858

> Note: If the button is held down more than 4 seconds after being plugged in, the smart switch will NOT reset and will remain with the current network.


Impact123:
I'd add the note, and maybe the link, as well

Koenkk:
Thanks!

PR #1730, opened 2022-11-28
Update README.md

algoz098:
Hey guys, so, i want to share a problem i had for a couple of months, and manage to solve.

It prob has happens to other ppl, and prob, at least a couple of those, still have, or never manage to solve.

If the wrong place, sorry, but i think this should be in the FAQ, even it's not the most common problem.

So, i have a setup with a good router stick, home assistant, and zigbee2mqtt addon.

The problem: some devices was not updating it's status, or desyncing from the network, also, was unable to add new devices. Map, was not being generate prob and timeout was happening without a clue or reason.

I started asking for help, changing the router stick, and without luck, it just ignored because it was kinda working almost all the times.

One day i decide to modify a door sensor, and for such, i needed it to connect to the network to test, it was unable to do so.

I started zigbee2mqtt in a new computer, though docker, and it worked FLAWLESS, after a while, i try to move all the devices to such new device.

Everything working perfect. Doubled the size of the network, no problem. Add more automations, no problem.

The reason of such strange behaiviour was my raspberry pi model 3+, because it just have 1gb memory or maybe the processor, but, the problem was the limitations of the platform.

The reason for using such a limited board, is, the model 4 is expensive where i live, and i never had such problems. It started without warning.

So, i would like to add such information to the FAQ of Zigbee2mqtt, for more people now that you need a little room of memory and processor to avoid timeouts.

### DIFF
--- a/docs/guide/faq/README.md
+++ b/docs/guide/faq/README.md
@@ -106,3 +106,11 @@ sed -i 's/GRUB_CMDLINE_LINUX_DEFAULT="/&usbcore.autosuspend=-1 /' /etc/default/g
 update-grub
 systemctl reboot
 ```
+
+## Strange timeouts and desyncs from the network
+If you have experienced strange timeouts, unable to add new devices, or just strange unexplicable behaiviours, the problem maybe limitation of hardware.
+- Such as running too much in the single computer
+- Raspberry pi model 3+ as a homeassistant addon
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks! I rephrased it a bit and added some more points.

PR #1733, opened 2022-11-29
Add Philips Hue Festavia

Sabesto:
Add documentation for Philips Hue Festavia gradient string light (250 lights). Documentation for gradient scene is missing but supported


Koenkk:
Thanks!

PR #1734, opened 2022-11-30
TS0601 (AQ): Remove PM2.5 from Options

RubenKelevra:
This device does not expose PM2.5 and is not capable of measuring PM2.5, so this option makes no sense.


Koenkk:
Thanks!

PR #1737, opened 2022-12-03
Update PTM_215Z.md

miksen:
Make page easier to read by combining two tables (currently you have to cross-reference them when trying to determine the location of a button for a certain channel)


Koenkk:
thanks!

PR #1740, opened 2022-12-03
Add position to channel table for PTM_215Z

Alxandr:
Add position to channel table for PTM_215Z to make it easier for users so they don't have to consult two tables at once.


Koenkk:
this was just done in https://github.com/Koenkk/zigbee2mqtt.io/pull/1737

Alxandr:
Ah - that's even better :)

PR #1744, opened 2022-12-05
Adjust a typo to position

bruvv:
add more clearer that it is about the valve position and not about something else


Koenkk:
Thanks!

PR #1746, opened 2022-12-05
Added Lidl/Silvercrest/Livarno 14158704L pairing notes

Rene-Sackers:
These Lidl lights are very picky about how to go into pairing mode. Best way I found is to say "Mississippi" between each on/off toggle so I mention it as a rule of thumb, but this might not be a good recommendation for international audiences, please do comment.


Koenkk:
Thanks!

PR #1747, opened 2022-12-06
Changed fz and tz references in example converter

danieledwardgeorgehitchcock:
Resolves Koenkk/zigbee2mqtt#15383


Koenkk:
See https://github.com/Koenkk/zigbee2mqtt/issues/15383#issuecomment-1339674386

PR #1761, opened 2022-12-13
Update 01_support_new_devices.md

aliijm:
Removed duplicate text and link from converters section, and revised grammar in a few places.


Koenkk:
Thanks!

PR #1762, opened 2022-12-14
PoPP 701721 thermostat: add note reg. feature availability depending on firmware version

ernstkl:
This is based on my personal experience with two devices, one with each of the firmware versions mentioned.


Koenkk:
Thanks!

PR #1766, opened 2022-12-17
Update pairing notes for ikea led driver

Alxandr:
We found we had to hold the button for 5 seconds, not just click it to pair to Zigbee.


Koenkk:
Thanks!

PR #1774, opened 2022-12-20
More explaination on External_measured_room_sensor

TomTonic:
Based on the Danfoss documentation and some experimentation

### DIFF
--- a/docs/devices/701721.md
+++ b/docs/devices/701721.md
@@ -126,7 +126,14 @@ The minimal value is `5` and the maximum value is `35`.
 The unit of this value is `°C`.
 
 ### External_measured_room_sensor (numeric)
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

docs/devices/701721.md
```diff
@@ -126,7 +126,14 @@ The minimal value is `5` and the maximum value is `35`.
...
+
+The way the TRV operates on the `External_measured_room_sensor` depends on the setting of the `Radiator_covered` property:
+
+If `Radiator_covered` is `true` (Auto Offset Mode): You *must* set the `External_measured_room_sensor` property *at least* every 3 hours. After 3 hours the TRV disables this function and resets the value of the `External_measured_room_sensor` property to -8000 (disabled). You *should* set the `External_measured_room_sensor` property *at most* every 30 minutes or every 0.1K change in measured room temperature.
```
sjorge, line 133: 0.1K … i doubt kelvin here is correct.

TomTonic: Well Kelvin is what the Danfoss documents say. In the end a difference in Kelvin it’s the same as the difference in Celsius (same scale, different ‚offset‘). But if for the sake of clarity and understandability, it should probably be changed to 0.1C.

PR #1775, opened 2022-12-20
improves IKEA E2001_E2002 pairing instructions

tadly:
For me pairing would only work when "spamming" the pair button multiple (more than 4) times. Based on [this](https://www.reddit.com/r/tradfri/comments/qjp3id/comment/hje4uoj/?utm_source=share&utm_medium=web2x&context=3) reddit comment, 6-7 should be enough but the remote enters pairing mode regardless of how many times you press past the "correct" amount, hence me recommanding 10.

I don't know if this is model or firmware dependent which is why I left the original instructions as there might be devices where only those work.

### DIFF
--- a/docs/devices/E2001_E2002.md
+++ b/docs/devices/E2001_E2002.md
@@ -28,7 +28,7 @@ pageClass: device-page
 
 ### Pairing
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1794, opened 2023-01-02
Added new hue effects to Philips 929002966501

merlinschumacher:
This adds the new hue effects to the page of the Hue Surimu 929002966501


Koenkk:
Thanks, note that this part is automagically generated so no need to update manually next time 😄

PR #1795, opened 2023-01-03
Updated cc2531 flashing macos instructions

bruvv:
boost got updated and v1.8 does not work anymore, so you have to specify the older boost version.


Koenkk:
Thanks!

PR #1797, opened 2023-01-03
Added IKEA PILSKOTT LED floor lamp

TimoRemon:


Koenkk:
Thanks!

N3rdix:
@Koenkk has Support for this device been added already? Couldn‘t find any PR related to this model and would be happy if it can be supported

PR #1798, opened 2023-01-03
Add legrand 067766 device

AntorFr:
Add image for Legrand 067766 Centralized ventilation switch  ([here](https://www.legrand.fr/pro/catalogue/44198-micromodule-pour-vmc-connectee/micromodule-pour-rendre-une-vmc-2-vitesses-connectee-250w-max))


Koenkk:
Thanks!

PR #1800, opened 2023-01-04
Added gradient and hue effects documentation to some Hue lights

marblecoma:
Adding gradient documentation for the Hue gradient lightstrip (this is already supported in the dev branch of zigbee2mqtt). Also, adding hue effects documentation in line with https://github.com/Koenkk/zigbee-herdsman-converters/pull/5267.


Koenkk:
Thanks for the pr, but it is not needed to manually update this. It will be automatically updated with the next release.

PR #1802, opened 2023-01-05
Update ota_updates.md

ReneTode:
added a tip to let people know about the risks from firmware updates.


Koenkk:
I agree, thanks!

PR #1804, opened 2023-01-05
Adding a note about a current todo in the base converter

DennisGaida:
It took me a while to figure out why binding didn't work / why I always received `toggle_1` when clicking any of the rockers in `command` mode. The code comment actually made it clear that it is a current *TODO/BUG* that should be reflected in the docs to make it easier for other users..


Koenkk:
Thanks!

PR #1807, opened 2023-01-07
Added support for Schneider Wiser Smoke Alarm

dennisgranasen:


docs/devices/W599001.md
```diff
@@ -0,0 +1,5 @@
+<!-- Notes BEGIN -->
+## Notes
+Cluster 64516 is not implemented.
```
Koenkk, line 3: It is not clear, what this means for the user, can you explain it?

dennisgranasen: Well, it is not clear for me either... but I suppose I could make that more explicit...

Koenkk:
Thanks!

PR #1809, opened 2023-01-08
Fix undefined 'value_json' error

andrenarchy:
With zigbee2mqtt 1.29.1 and Home Assistant 2023.1.1 the bridge state sensor spammed the Home Assistant logs with the following error:
```
ERROR (MainThread) [homeassistant.helpers.template] Template variable error: 'value_json' is undefined when rendering '{{ value_json.state }}'
```

The `zigbee2mqtt/bridge/state` topic returns a plain string instead of a JSON payload. Removing the `value_template` fixed the errors and the sensor value shows up as `online` correctly.


Koenkk:
Thanks!

PR #1812, opened 2023-01-10
Added device image for Namron ZigBee Termostatplugg 16A

slippern1:
Added device image for Namron ZigBee Termostatplugg 16A.
https://www.namron.com/products/namron-zigbee-termostatplugg-16a/


Koenkk:
Thanks!

PR #1814, opened 2023-01-10
Update IH-K009.md

martindell:
Short note that battery value takes 30 minutes to show up (I thought there was a fault - but they just take a while).


Koenkk:
This part of the docs is generated from the code. I've adapted the code so this page will be automatically updated with the next release. Thanks!

PR #1816, opened 2023-01-11
Added and updated images for the following Mercator Ikuü devices - SM…

kiwizznz:
Added and updated images for the following Mercator Ikuü devices - SMA02P SMA03P SMA04P SMCL01-ZB SMD4109W-RGB-ZB SPBS01G SPP02G SPP02GIP SPP04G SPPUSB02G SSW01G SSW02G SSW03G SSW04G SSWF01G


Koenkk:
Thanks!

PR #1828, opened 2023-01-16
Add note that L1530 is known as "PWID012V24P-EU"

eikowagenknecht:
On all of my L1530 devices, the type "PWID012V24P-EU" is written instead. Added it to the docs to make clear that it's the right device.


Koenkk:
Thanks!

PR #1829, opened 2023-01-17
Improve wording for reset with Hue dimmer switch

eikowagenknecht:
It was a bit unclear if both dimmer switches work and the buttons are labeled differently on v2 switches.

Maybe this can be changed on other device pages as well.


Koenkk:
Thanks!

PR #1834, opened 2023-01-19
Replace docker-compose with docker compose

Impact123:
[`docker-compose` is no longer supported](https://docs.docker.com/compose/install/#scenario-three-install-the-compose-standalone)
image


Koenkk:
Thanks!

PR #1839, opened 2023-01-22
add PILSKOTT pendent as a seperate device

GsakuL:
Koenkk/zigbee2mqtt#16288

sadly, the product images (without other stuff) are a case of "white object on white background". The other image including the top part/anchor/mount/driver case/etc. would have been even worse to make out, as it would "degrade" into "a thin black line"

Edit: the device definition is here: Koenkk/zigbee-herdsman-converters#5379


Koenkk:
Thanks!

PR #1844, opened 2023-01-25
Improve basic development experience

mdziekon:
Changelog:
- Added support for switching dev server port
- Added support for including one single device to the dev build, even in `EXCLUDE_DEVICES=yes` mode
- Added `.nvmrc` file for nvm or nvm-compatible tool users, with node version conforming to the README.md docs


.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
Koenkk, line 1: I don't expect any problem, but could you check if node 18 also works fine (node 16 will go EOL in +- 7 months from now)

Koenkk: Preferably in both places, or maybe even better, can `.nvmrc` be removed such that you are forced to specify it?

mdziekon: I'm not sure if I follow - how removing the `.nvmrc` file is better? `.nvmrc`'s purpose is to specify that version for you, similar to how `.editorconfig`'s file is to specify IDE's / editor's configuration.

Koenkk:
Looks good now, thanks!

PR #1849, opened 2023-01-26
fixed broken image refs and image locations

TomTonic:
On ZBDongle-E device page


TomTonic:
Directory only allows 150x150 pixel images

PR #1851, opened 2023-01-27
Update docker-compose syntax to docker compose plugin

psi-4ward:


Impact123:
https://github.com/Koenkk/zigbee2mqtt.io/pull/1834

psi-4ward:
😄👍🏻

Probably we should add the Tipp to clarify the difference? I think many ppl didn't understand why it won't work on their (old) systems

Koenkk:
@psi-4ward could you target this pr against `dev`?

psi-4ward:
> @psi-4ward could you target this pr against `dev`?

#1834 from @Impact123 has already solved 90% of my PR. I've only added 
```
::: tip
`docker-compose` (v1) is deprecated and you should use the docker compose **plugin** instead (`docker compose` without the dash).
If your system relays on the old Python-based `docker-compose` you've to add the dash to the docker compose command.
:::
```

to give ppl with outdated docker compose a hint why they probably are getting an error. Do you think we need this _tip_?   Usually I would say the one who use docker should know this or easily google it by themselves but Z2M users are often everything but good SysOps ;)

PR #1856, opened 2023-01-28
add command to store sniffed data for later analysis

Entepotenz:
This might decrease the amount of research people need to do to provide proper zigbee captures for analysis


Koenkk:
Good addition, thanks!

PR #1857, opened 2023-01-29
Fix device specific configuration link

Impact123:
As the title says. It previously linked to the main configuration's docs.


Koenkk:
Thanks!

PR #1858, opened 2023-01-29
added more details to the routing section

TomTonic:


docs/advanced/zigbee/02_improve_network_range_and_stability.md
```diff
@@ -33,8 +33,24 @@ To change the Zigbee channel Zigbee2MQTT uses you have to set the [`channel` in
...
+
+Introducing a router ([read more about this](./01_zigbee_network.md)) can improve the forward path as well as the return path:
+
+![Routers can stabilize the complete message roundtrip.](https://www.zigbee2mqtt.io/images/routing3.jpg)
+
+You might choose a dedicated router (for example, a [SONOFF ZBDongle-E based router](https://www.zigbee2mqtt.io/devices/ZBDongle-E.html)) or a mains-powered Zigbee device (for example, a [Hue lamp](https://www.zigbee2mqtt.io/devices/8719514301481.html#philips-8719514301481)) to stabilize your network. Almost all AC powered devices will serve as a router.
+
+Please note that there are routers of mediocre quality that might not harmonize well with your network (for example, the [SONOFF Smart Plug S26R2ZB](https://www.zigbee2mqtt.io/devices/S26R2ZB.html) is known to be limited). This may yield in message routing errors. In case you have such devices in your network, it might help to add additional routers of better quality and bind your devices to these routers (re-pairing devices with “Permit join” restricted to the new/better router) to improve the overall network performance.
```
Koenkk, line 51: > Pease note that there are routers of mediocre quality that might not harmonize well with your network 

I know some firmware versions of the S26R2ZB are known to lock the device after some time, but I don't think they are bad routers as long as they are online (since they use the TI stack)

TomTonic: Hmm, sorry, I don't understand the intend of your comment. Would you want
1. the reference to this specific device to be removed? 
2. recommend something else to the user?
3. the whole section to be removed?

I'm not sure if I understand this correctly, but if there are some firmware versions of the S26R2ZB known to lock the device after some time and if there is no way for the user to update the firmware (I don't see any), those users will end up with an unreliable network. One way would be to throw the devices in the trash and buy new ones. Another way would be to add stable routers to the network and keep the devices in use (only as smart plugs). Is there a third alternative?

docs/advanced/zigbee/02_improve_network_range_and_stability.md
```diff
@@ -33,8 +33,24 @@ To change the Zigbee channel Zigbee2MQTT uses you have to set the [`channel` in
 ## Interference from other 2.4 GHz devices
 Any device using the open 2.4 GHz spectrum could interfere with Zigbee such as Bluetooth or gaming devices like Logitech “Unifying” or “Lightspeed” or Razer “Hyperspeed Wireless”.
 
-## Adding more routers to your network
-In a Zigbee network, each router will extend the range of the network ([read more about this](./01_zigbee_network.md)). Almost all AC powered devices will serve as a router.
+## Adding routers to your network
+"Zigbee is a low-power wireless mesh network standard targeted at battery-powered devices" (see https://en.wikipedia.org/wiki/Zigbee). Yet, low transmission power can be the cause of an unstable or unreliable network:
+
+![The signal could be to weak fpr the message to reach its target.](https://www.zigbee2mqtt.io/images/routing1.jpg)
```
Impact123, line 39: > be ~~to~~ weak ~~fpr~~

`too`, `for`

docs/advanced/zigbee/02_improve_network_range_and_stability.md
Koenkk: `for example, some versions of the [SONOFF Smart Plug S26R2ZB]` sounds better.

TomTonic: fixed wording

Koenkk:
Thanks!

PR #1860, opened 2023-01-30
add image for bankamp grazia

ahemwe:
This refers to [pr #5369](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5369)

I've asked the vendor for permission to use this [image](https://www.bankamp.de/thumbnail/f3/fc/b3/1669132239/fca4eeb685b3f889a32c53e7970e9spw.jpg). Unfortunately they refused this request.

Furthermore, they confirmed that the zigbee controller is also used in other models. I personally observed another device which also works with the same config. Should we make the naming more generic "bankamp light" instead of a single model?


Koenkk:
> Should we make the naming more generic "bankamp light" instead of a single model?

Yes please do so. You can make the picture a generic light.

ahemwe:
I've raised this PR to make the description more generic: [pr 5424](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5424)

Concerning the image, the on in this PR is quite representative, as it reflects the style of the vendor quite well and is self made, not from above mentioned link.

Koenkk:
I see, could you rebase this pr?

ahemwe:
done after some hurdles :-(

Koenkk:
Thanks!

PR #1861, opened 2023-02-01
Adding image for Hue LTV003

mattkrins:
Adding image for LTV003/8719514301528/Philips Hue white Ambiance filament ST72 B22 LED


Koenkk:
Thanks!

PR #1868, opened 2023-02-03
Update LDSENK08 picture and add HG07310 one

Chartreusito:
Update LDSENK08 picture and add HG07310 before device is added in supported devices


Koenkk:
`HG07310.jpg` should be `HG06335-HG07310.jpg`

Chartreusito:
@Koenkk still getting the hangs of it, i thought i would be able to make 2 different devices out of the HG06335-HG07310 which are 2 different hardware versions. But there's indeed no way for z2m to know what hardware version it is as they share the same modelID and _TZ1800_fcdjzz3s i'll do another PR to update LDSENK08 picture only, thanks.

PR #1874, opened 2023-02-06
Update connect_to_a_remote_adapter.md

matti125:
Spent a day figuring out why my rpi did not work with slesh adapter. Perhaps it will help someone in the future to know that really, the current raspbian cannot be used as such with the stick.


Koenkk:
Thanks!

PR #1880, opened 2023-02-08
Add notes about control_sequence_of_operation for SMaBiT AV2010/32

sjorge:
Now that we have this change https://github.com/Koenkk/zigbee-herdsman-converters/commit/f7e27c2dd0326db676f5ac844e1da0bb011bf3f0, update the docs to reflect that we need a restart of z2m.


Koenkk:
Thanks!

PR #1883, opened 2023-02-08
Update ICPSHC24-10EU-IL-1.md

aso824:
Copy line from ICPSHC24-30EU-IL-1 about pairing.

I observed same behaviour (long press was required), single press didn't worked.
Also, I paired successfully -30 version in ~5m distance, but -10 needed to be very close to the ZB adapter.


Koenkk:
Thanks!

PR #1884, opened 2023-02-09
Update R7051.md

elninosi:
Correcting topic to publish payload.

### DIFF
--- a/docs/devices/R7051.md
+++ b/docs/devices/R7051.md
@@ -49,7 +49,7 @@ It's not possible to read (`/get`) or write (`/set`) this value.
 The unit of this value is `mV`.
 
 ### Warning (composite)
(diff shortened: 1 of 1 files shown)


Koenkk:
I don't see how this could work, `set_message` is not part of the Z2M api. Could you explain this further?

elninosi:
It did not work with "set" payload. After a few more digging I found out it is working with "set_message", here is log example:
slika

Koenkk:
Can you provide the debug log when using `/set`?

Koenkk:
Could you provide the herdsman debug log of both `set` and `set_message`?

See https://www.zigbee2mqtt.io/guide/usage/debug.html on how to enable the herdsman debug logging. Note that this is only logged to STDOUT and not to log files.

PR #1899, opened 2023-02-16
Update 8719514440937_8719514440999.md

savdagod:
When pairing a brand new tap, the setup button doesn’t work. The manual states you must hold button 1 for 3 sec. After that the device will pair. If you unpair the device, the setup button can then be used to reset/pair the device again.


Koenkk:
Thanks!

PR #1906, opened 2023-02-19
Update BSEN-C2.md

bonjour81:
after pairing issues, I discovered that same pairing procedure as https://www.zigbee2mqtt.io/devices/8750001213.html  is working. I updated note with copy-paste of the working one + minor adaptation ( the door sensor has only one battery and does not emit sound).


Koenkk:
Thanks!

PR #1910, opened 2023-02-20
Update BSEN-C2.md

bonjour81:
I spent/waited quite some time (years!) to find a door sensor,Z2M compatible and suitable for outdoor use and finally found this one!  This is great for whoever want to use such sensor on a garden gate, letter box or whatever outdoor use. IP45 level is as well as "indoor/outdoor use" information is from Bosch documentation. I'm not sure if it's against any rules or best pratice to mention technical characteristics on a sensor page, but I thought it might be interesting for other people that might seek outdoor door sensor too. Feel free to cancel if not compliant.

### DIFF
--- a/docs/devices/BSEN-C2.md
+++ b/docs/devices/BSEN-C2.md
@@ -25,7 +25,7 @@ pageClass: device-page
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
 ## Notes
 
-### Indoor / Outdoor use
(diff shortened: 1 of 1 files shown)


bonjour81:
Hello, would it be possible and make sense to add "indoor/outdoor" after "Door/window contact II" in the supported device list. IMHO, It's pretty useful when searching/filtering device with keyword "outdoor".

Koenkk:
Thanks!

PR #1914, opened 2023-02-22
Create 12502.jpg

Wiese-TKW:
Image file for Scan Products Zigbee Switch 3.0 item no. 12502


Koenkk:
It looks like the image is not valid, could you fix this?

Screenshot 2023-02-22 at 20 44 49

PR #1915, opened 2023-02-22
Add factory reset sequence for AwoX 33957/EGLO 12239

fdemmer:
Text adapted from IKEA device docs and tested with an EGLO 12239.


Koenkk:
Thanks!

PR #1916, opened 2023-02-24
Fix example tuya datapoint

afharo:
I noticed a small mismatch in the example: line 86 talks about _... datapoint `106`..._, so I thought it might be worth updating the following references to the datapoint.


Koenkk:
Thanks!

PR #1920, opened 2023-02-25
Update devices-groups.md

baze-:
Clarified how Debounce combines messages and when messages are sent out. Added also example.


Koenkk:
Thanks!

PR #1924, opened 2023-02-27
Added Apache2 reverse proxy configuration

dajuly20:
Credit: Florian Metzger-Noel https://stackoverflow.com/a/60506715 Configuration is tested & running exactly like that. Took quite a while to figure out, so maybe this will help someone


Koenkk:
Thanks!

PR #1928, opened 2023-03-01
Update E2103.md

thenakuls:
Adding blurb about importance of using the repeater as I just went through the procedure


Koenkk:
Thanks!

PR #1931, opened 2023-03-04
to replace with the correct picture 

fred-c1:
as discussed here:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/5469


Koenkk:
Thanks!

PR #1937, opened 2023-03-07
Update RTCZCGQ11LM.md with conbee2 firmware version

MansM:
works with the stable firmware for conbee2 0x26580700 as well


Koenkk:
Thanks!

PR #1939, opened 2023-03-08
Move legacy_entity_attributes to advanced key

basnijholt:
I just installed the addon and noticed this key moved.

Additionally, the default for me seems to be `false`.

### DIFF
--- a/docs/guide/configuration/homeassistant.md
+++ b/docs/guide/configuration/homeassistant.md
@@ -19,15 +19,17 @@ homeassistant:
   # Optional: Home Assistant status topic (default: shown below)
   # Note: in addition to the `status_topic`, 'homeassistant/status' will also be used
   status_topic: 'hass/status'
+  # Optional: Home Assistant legacy triggers (default: shown below), when enabled:
+  # - Zigbee2mqt will send an empty 'action' or 'click' after one has been send
+  # - A 'sensor_action' and 'sensor_click' will be discoverd
+  legacy_triggers: true
+
+advanced:
   # Optional: Home Assistant legacy entity attributes, (default: shown below), when enabled:
   # Zigbee2MQTT will send additional states as attributes with each entity. For example,
   # A temperature & humidity sensor will have 2 entities for the temperature and
   # humidity, with this setting enabled both entities will also have
   # an temperature and humidity attribute.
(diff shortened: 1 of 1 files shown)


docs/guide/configuration/homeassistant.md
```diff
@@ -19,15 +19,17 @@ homeassistant:
   # Optional: Home Assistant status topic (default: shown below)
   # Note: in addition to the `status_topic`, 'homeassistant/status' will also be used
   status_topic: 'hass/status'
+  # Optional: Home Assistant legacy triggers (default: shown below), when enabled:
+  # - Zigbee2mqt will send an empty 'action' or 'click' after one has been send
+  # - A 'sensor_action' and 'sensor_click' will be discoverd
+  legacy_triggers: true
```
Koenkk, line 25: Default is `true` indeed, but I propose to list `false` here (since we don't want users to use this anymore) and change ` (default: shown below)` to ` (default: true)`

basnijholt: Oh upon re-reading your message, I understand you only propose to change the text.

docs/guide/configuration/homeassistant.md
```diff
@@ -19,15 +19,17 @@ homeassistant:
...
   # A temperature & humidity sensor will have 2 entities for the temperature and
   # humidity, with this setting enabled both entities will also have
   # an temperature and humidity attribute.
   # Note: Disabling this option, requires a Home Assistant restart
-  legacy_entity_attributes: true
-  # Optional: Home Assistant legacy triggers (default: shown below), when enabled:
-  # - Zigbee2mqt will send an empty 'action' or 'click' after one has been send
-  # - A 'sensor_action' and 'sensor_click' will be discoverd
-  legacy_triggers: true
+  homeassistant_legacy_entity_attributes: false
```
Koenkk, line 34: Sorry for checking just now, but this can stay under `homeassistant` as `legacy_entity_attributes`. Both this and `advanced`.`homeassistant_legacy_entity_attributes` will still work but `homeassitant`.`legacy_entity_attributes` is preferred. Code: https://github.com/Koenkk/zigbee2mqtt/blob/cdf62ea5601f74af19b1626b72afc88edbd8e4ab/lib/util/settings.ts#L128

PR #1948, opened 2023-03-12
Fix AC03641 pairing instructions

crabmanX:
The process differs from all other lamps as described here: https://www.osram.com/cb/lightify/lightify-home/lightify-home-faq/lightify_home_faqs.jsp


Koenkk:
Thanks!

PR #1949, opened 2023-03-13
instructions for router firmware flash with ubuntu terminal only

niemsoen:
this adds instructions on how to flash the router firmware without 3rd party tools, only using ubuntu terminal tools


niemsoen:
i added the images to `docs/images` :)

Koenkk:
Thanks!

PR #1951, opened 2023-03-13
Bosch `BSP-FZ2`, `BSP-EZ2` and `BTH-RA`: Add notes on how to pair and factory reset

DerDreschner:
The Bosch `BSP-FZ2` and `BSP-EZ2` plugs as well as `BTH-RA` thermostat are missing notes on how to pair and factory reset them. This PR is adding them. I based my text on the already-existing notes on the Bosch `8750001213` device.


Koenkk:
Thanks!

PR #1954, opened 2023-03-14
Added Note for a firmware bug in Osram AC03642 bulb

JsBergbau:
Added Note for a firmware bug that bulb will be brighter than saved value, when using a classic switch to turn off and on. Still the bulb will report in both cases the same brightness value.


Koenkk:
Thanks!

PR #1960, opened 2023-03-16
Add reference to groups in mqtt topic guide

junosuarez:
Z2M treats Devices and Groups mostly the same (as "entities" in the code). This PR updates the reference docs to make this more explicit and clear for people developing MQTT integrations.


Koenkk:
Thanks!

PR #1961, opened 2023-03-16
Philips 929002398602: update reset instructions

junosuarez:
The current instructions say "push the reset button", but it is hidden under the battery cover. This makes the documentation clearer for setup.


Koenkk:
Thanks!

PR #1964, opened 2023-03-19
Add Garza ST-64 Smart Bulb

jccf07:
Hello, I have been able to successfully use the Garza ST-64 light bulbs, with temperature and brightness regulation. 
I would like to share it and be able to have the image of the devices

I attch garzast64.js in rar file, that it is working properly and the image 150x150.jpg

What should I do?

thank you 

[GarzaST64.zip](https://github.com/Koenkk/zigbee2mqtt.io/files/11010818/GarzaST64.zip)

GarzaST64 150x150


Koenkk:
This device should already be detected as a TS0502B (this is a tuya device). No need to add additional support for it.

jccf07:
I tried to pair them in november 2022. but it was impossible, zigbee2mqtt was detecting them as incompatible devices.  try it several times.  
Since it was not possible, I did it manually, it has already been working properly for 5 months.  That's why I want to share it.  I don't know if they will be supported in new versions.

PR #1969, opened 2023-03-23
Document use of unix sockets for Koenkk/zigbee2mqtt#17123

oddlama:
Documentation for https://github.com/Koenkk/zigbee2mqtt/pull/17123 as requested. Let me know if that suffices or if you want a whole category explaining it. IMO anyone specifically looking for unix sockets should know what to do given that description.


Koenkk:
Thanks!

PR #1971, opened 2023-03-24
Update Silabs flashing tips

Hedda:
Update Silabs flashing tips, moving "Universal Silicon Labs Flasher (by Nabu Casa)" to the top of the list now that its developers are trying to make it more compatible out-of-the-box with Silicon Labs based hardware radio dongles from other manufacturers than the Home Assistant SkyConnect and Home Assistant Yellow.


Koenkk:
Thanks!

PR #1980, opened 2023-03-31
New RGBW lamps added

dhanukapulkit:


Koenkk:
- Could you resize the pictures to be 150x150?
- Do you have a 4058075729025.jpg without the box on it (just the lamp)?

PR #1983, opened 2023-04-03
Add pairing note for BSD-2 regarding install codes

N3rdix:
This might help users not aware of the importance of the install code before pairing the device

### DIFF
--- a/docs/devices/BSD-2.md
+++ b/docs/devices/BSD-2.md
@@ -24,6 +24,11 @@ pageClass: device-page
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
 
+## Notes
+### Pairing
+- Before starting the pairing process the device's install code needs to be added to Zigbee2MQTT. Note down the install code e.g. by scanning the QR code with the camera of your mobile phone and copy the received value
+- The install code can be found inside the device before attaching it to the mounting plate
+- Add this install code under Settings -> Tools -> Add install code
 
 <!-- Notes END: Do not edit below this line -->
 


Koenkk:
Thanks!

docs/devices/BSD-2.md
```diff
@@ -24,6 +24,11 @@ pageClass: device-page
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
 
+## Notes
+### Pairing
+- Before starting the pairing process the device's install code needs to be added to Zigbee2MQTT. Note down the install code e.g. by scanning the QR code with the camera of your mobile phone and copy the received value
+- The install code can be found inside the device before attaching it to the mounting plate
+- Add this install code under Settings -> Tools -> Add install code
```
larsxschneider, line 31: @N3rdix My BSD-2 has an install code of 9 blocks each with 4 hex characters printed on the back. That's 36 characters (excluding the `-` characters). 

If I paste that install code then I get a `Unsupported install code, got 36 chars, expected 95 or 91` error. My dongle is the `SONOFF Zigbee 3.0 USB Dongle Plus,TI CC2652P + CP2102(N)` and I am running zigbee2mqtt [1.36.1](https://github.com/Koenkk/zigbee2mqtt/releases/tag/1.36.1) commit: [ffc2ff1](https://github.com/Koenkk/zigbee2mqtt/commit/ffc2ff1).

Any idea why the pairing instructions don't work for me? (I've also asked about it [here](https://github.com/Koenkk/zigbee2mqtt/discussions/15966#discussioncomment-9019980))

N3rdix: good catch, indeed the 9 blocks code is not the complete code you need (there are some other parts in front (like the ieee code between the QR code and the install code).
Easiest way I do this is scanning the QR code with my iPhone and copy that text

larsxschneider: @N3rdix I opened a PR to improve the instructions: https://github.com/Koenkk/zigbee2mqtt.io/pull/2684

Kilshaw878 (commented):
- [ ] Ggggg ****************

docs/devices/BSD-2.md
hannesroets: Hi, did you manage to add them? I'm having issues with the QR code on the back of my Frient smoke detectors.
`|0015BC0031011E9D|BBBBEF2AAD0B5DB74AAA030CF02EF2B4841D|`
(changed some characters for security reasons)
When pasting the full string it doesn't seem to accept my code.

hannesroets: Thanks for the suggestion.
I just tried android and ipad both with different apps. They all resulted in the same code.
I'm lost. They used to work fine on a conbee 2 zha on an outdatedi instance of ha. I even bougt a new zigbee stick (sonff 3 plus v2 e type stick) both woth orignal firmware and updated one with zigbee2mqtt. I can't get passed the interview state.
Feel free to try scanning. I attached the qr code.
Thanks again!
IMG20250213163943~2

N3rdix: Ok this is a totally different device, not a Bosch. Can't help you with that, but I doubt this QR code contains the install code?
My QR code scanner does not recognize it at all...

You might raise an issue in Zigbee2Mqtt and ask in context of that specific device

PR #1987, opened 2023-04-06
Add 4512728 image

KristianAndersson:
Add image for Namron Zigbee 1 channel switch K2 - Black (4512728)
https://www.namron.com/products/namron-zigbee-1-kanal-bryter-k2-sort/


Koenkk:
Thanks!

pinpoint90:
> Thanks!

Added this DEVICE to zigbee2mqtt on home assistant.  Unable to use any of the functions, can only see battery level. 
 image

PR #1993, opened 2023-04-10
Green power, broadcasts and congestion

tausen:
Highlight different types of green power device pairing and performance cost of broadcasts. Copy Green Power device intro from misc devices to "Zigbee network".

I've spent some time debugging why my green power switches performed so poorly (started a discussion with myself here: https://github.com/Koenkk/zigbee2mqtt/discussions/17263). This is an attempt to spare others from the same headache.

I'm not sure in which cases broadcasting green power device events is a good idea, perhaps someone else could shed some light on that? In my case at least it would be better if green power device events were just always re-transmitted by the proxy as unicast...


Koenkk:
> I'm not sure in which cases broadcasting green power device events is a good idea, perhaps someone else could shed some light on that? In my case at least it would be better if green power device events were just always re-transmitted by the proxy as unicast...

I'm not sure if this is possible, but I believe some devices can be setup to directly interpret the commands, then broadcast makes sense if you want to address multiple devices.

Thanks for the docs!

PR #1994, opened 2023-04-11
add docs for homeetec products

elad-bar:


Koenkk:
@elad-bar could you update this PR? (conflict with `TS0601_cover_4`)

PR #1995, opened 2023-04-11
Update HG08164.md

yasdfgr:
added reported Manufacturer and Model Identifier


yasdfgr:
Yes, i know. The benefit is that you can search for the device by this identifier, even if you don't know the "seller name" of the product, because this is reported by the device itself. So regardless under which name the product is sold - you have a unique identification of the device.

Koenkk:
But is this manufacturer name listed on the sellers page?

PR #2001, opened 2023-04-14
Add picture for Iluminize 5123.1110

saseba:
Added picture for Iluminize 5123.1110 a "Zigbee 3.0 controller with adjustable current 250-1500mA, max. 50W / 48V SELV"
I will shortly also add device description to zigbee-herdsman, which is in turn just a copy of "5120.1100"


Koenkk:
thanks!

PR #2004, opened 2023-04-15
Add WDE002960 push button dimmer

energizer91:
Adding image and docs (not sure what I should do there though) for Schneider Electric WDE002960 Push button dimmer


Koenkk:
Thanks!

PR #2012, opened 2023-04-21
Update 324131092621.md

jeeftor:
I was trying to figure out how to register multiple button presses on my hue remotes (like the hue bridge offers) - turns out Legacy mode is the way

### DIFF
--- a/docs/devices/324131092621.md
+++ b/docs/devices/324131092621.md
@@ -74,6 +74,9 @@ This device supports OTA updates, for more information see [OTA updates](../guid
 
 * `legacy`: Set to false to disable the legacy integration (highly recommended), will change structure of the published payload (default true). The value must be `true` or `false`
 
+  - `legacy` mode will allow you to track the number of button presses - which are exposed under the `counter`field. The button presses must be `.25` apart to register.
+
+
 * `simulated_brightness`: Simulate a brightness value. If this device provides a brightness_move_up or brightness_move_down action it is possible to specify the update interval and delta. The action_brightness_delta indicates the delta for each interval. Only works when legacy is false. Example:
 ```yaml
 simulated_brightness:


docs/devices/324131092621.md
```diff
@@ -74,6 +74,9 @@ This device supports OTA updates, for more information see [OTA updates](../guid
 
 * `legacy`: Set to false to disable the legacy integration (highly recommended), will change structure of the published payload (default true). The value must be `true` or `false`
 
+  - `legacy` mode will allow you to track the number of button presses - which are exposed under the `counter`field. The button presses must be `.25` apart to register.
```
Koenkk, line 77: This should be moved below `<!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->`, but I don't think we should mention any legacy features here (we don't want people to use them anymore)

jeeftor: @Koenkk  - fair enough ... however the main issue would be that the existing Hue Apps handle this in software. Additionally ZHA apparently also has it handled in software. So when you come to Z2M one would be confused as to why a feature we "thought" was part of the remote is missing..

Ive seen some issues about it as well..
https://github.com/Koenkk/zigbee2mqtt/issues/10312


Happy to close the PR -> but - I do think perhaps somewhere in the docs this should be mentioned... (in a different way if you prefer...)

jeeftor: Is there a better place to mention somewhere in the docs for the various systems that multi-click is not an actual feature of the remote but has to be implemented in software somewhere?

Thanks!

jeeftor: Maybe the FAQ?

Koenkk: You can create a feature request [here](https://github.com/Koenkk/zigbee2mqtt/issues/new?assignees=&labels=feature+request&template=feature_request.yaml&title=%5BFeature+request%5D%3A+), if more people are interested in it, it can be implemented.

PR #2018, opened 2023-04-28
Add section about install code for Bosch Motion Sensor

itkama:
According to the manual https://www.bosch-smarthome.com/de/media/content/10_bedienungsanleitungen/bewegungsmelder/8-750-000-920_v001_shmd_qsg_de_20200211_web.pdf you need to also scan an install code with this device. There is already an issue (https://github.com/Koenkk/zigbee2mqtt/issues/17315) because this is missing in the documentation.  The person can't test this anymore, since he already send back the product, but it seems kind of obvious that it was the problem in that case.


Koenkk:
thanks!

PR #2019, opened 2023-04-28
Update documentation to include the pairing process via install code for Bosch Water Sensor

itkama:
According to the manual https://www.bosch-smarthome.com/de/de/service/hilfe/hilfe-zum-produkt/hilfe-zum-wassermelder/ you need to also scan an install code with this device. There was already an issue (Koenkk/zigbee2mqtt#17315) on the motion sensor because this is missing in the documentation. I think all currently supported Bosch devices that need an Install Code should have this in their documentation now.  @Koenkk is there any way to make sure this is added for new Bosch devices in the future (as they will probably keep this going).

### DIFF
--- a/docs/devices/BWA-1.md
+++ b/docs/devices/BWA-1.md
@@ -24,6 +24,10 @@ pageClass: device-page
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
 
+## Notes
+
+### Pairing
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

>  I think all currently supported Bosch devices that need an Install Code should have this in their documentation now. @Koenkk is there any way to make sure this is added for new Bosch devices in the future (as they will probably keep this going).

Currently not, I will think about how we can handle this.

PR #2024, opened 2023-05-02
Document how to expose floor sensor in HA UI

atus:
Moving discussion conclusion into documentation so it's easier to find https://github.com/Koenkk/zigbee2mqtt/issues/15144


docs/devices/4512737_4512738.md
```diff
@@ -23,8 +23,21 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
+## Notes
+### Exposing floor sensor in Home Assistant UI
```
Koenkk, line 27: If I understand correctly, this does not expose a new sensor but updated the thermostat entity to use the `outdoor_temperature`, if yes, please explain that here.

Leissson: I would replace "Exposing" with Using. Correct, this is just populating climate entity with `outdoor_temperature` insted of `local_temperature` .

PR #2029, opened 2023-05-06
Update SPLZB-131.md

emandtf:
Update for new "Reactive Power" measurement


Koenkk:
Thanks! For next time, there is no need to manually update this, it will be automatically generated with the next release.

PR #2032, opened 2023-05-08
Update notes on SZ-ESW01-AU.md

lskeet:
Added indication that this device is a Zigbee router.


docs/devices/SZ-ESW01-AU.md
```diff
@@ -24,7 +24,7 @@ pageClass: device-page
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
 ## Notes
-
+This smart plug also acts as a Zigbee router.
```
Koenkk, line 27: I'm not sure if this is useful info, as you can expect that every AC powered Zigbee device is a router.

lskeet: Unfortunately this hasn't been my experience so far. I have purchased some mains outlet (with neutral) and it does not have routing capabilities for some unknown reason. 
I found this identical information on https://www.zigbee2mqtt.io/devices/E1603_E1702_E1708.html useful when purchasing it so I thought it'd be good to add it here as well. 
Maybe I should also, or instead, add notes to the mains device that I found did not act as a router?

Koenkk: > Maybe I should also, or instead, add notes to the mains device that I found did not act as a router?

I think we should do this instead.

lskeet: I've reverted the initial router indication and added notes for non-routers.

Koenkk: I see, how did you determine that these devices do not have routing capabilities?

lskeet: I have them added to z2m and the device type is shown as an EndDevice. Additionally, this is also consistent when viewing them in my map.

Koenkk: I see, thanks!

PR #2042, opened 2023-05-15
Update TS0207_water_leak_detector.md

kzajac83:
Add anothet name from label of carton box of my device


Koenkk:
Please add this to https://github.com/Koenkk/zigbee-herdsman-converters/blob/cd8c128b2d1e03ad148cb000fae71ad95ec40762/devices/tuya.js#L1520 (then it will be automatically updated here)

PR #2049, opened 2023-05-19
Added example for device settings & Corrected Home Assistant URL

DennisGaida:
* Added an example on how to configure "any" MQTT discovery property - the trick being to just copy the hierarchy/words underneath the `homeassistant` entry
* Corrected a HA documentation URL that was changed


Koenkk:
Thanks!

PR #2051, opened 2023-05-22
Added note to clarify lack of router support in Moes SZ-EUB wall switches

llmora:
These wall switches provide an optional neutral cable connection, but even when connected the device always reports itself as EndDevice and does not work as a router. This has been confirmed by the vendor customer support:

> **I bought three ZS-EUB 1-gang switches with optional neutral. The devices are detected as EndDevice and do not work as routers, even when Neutral is connected. Is there a firmware upgrade for the device that allows it to work as router?**
> 
> Thanks for your feedback, our zigbee products need to be connected to a zigbee gateway.
> In addition, our zigbee switch cannot be used as a repeater. If you want to expand the zigbee signal, you need a zigbee gateway.
>


Koenkk:
Thanks!

PR #2056, opened 2023-05-25
Add Shade Control MC-02 cover

adarazs:
I'm submitting a PR for adding support for this device shortly.


adarazs:
Fixed the model number to be MC-02 instead of MC-01, turns out I got an early model with a non-matching version number.

Koenkk:
Thanks!

PR #2064, opened 2023-06-01
Update SIN-4-RS-20_LEX.md

drount:


drount:
Sorry for the late reply, totally missed the response.

After successfully pairing the device, any command send via UI ends up with
a timeout. After restarting  z2m (in my case the add-on) it works as
expected.

PR #2071, opened 2023-06-05
Added pairing instructions for Kwikset 66492-001

phan505:
Added pairing instructions/picutre for Kwikset 66492-001 in the notes section.


Koenkk:
Thanks!

PR #2074, opened 2023-06-07
Add more instructions on how to bind an E1745

romixch:
According to issue  #5778 it is not possible to directly bind a device or a group. Nevertheless the device actually sends messages to a groupID. Knowing this groupID one can set up a Group to receive them. I had appreciated it if it was explained here on zigbee2mqtt.


Koenkk:
Thanks!

PR #2075, opened 2023-06-08
fix: correct hass docs yaml entry

YodaDaCoda:
`zigbee2mqtt/bridge/state` gives a state, not a json object, so template is not required and produces an error in hass logs

```Template variable error: 'value_json' is undefined when rendering '{{ value_json.state }}'```

### CODE REVIEW

**Comment by Koenkk** [comment]
In non legacy mode it is a JSON payload, add the following to your z2m config:

```yaml
advanced:
  legacy_availability_payload: false
```

YodaDaCoda:
🤦 yeah that's sorted it, thanks.

is `legacy_availability_payload: false` the default for new installations?

Koenkk:
> is legacy_availability_payload: false the default for new installations?

yes!

PR #2091, opened 2023-06-18
correct images for ZY-M100-L and ZY-M100-S

pswid:
TS0601_smart_human_presence_sensor_2.jpg can be removed because real name of the device is ZY-M100-S

ZY-M100-LvsS


Koenkk:
Thanks!

PR #2124, opened 2023-07-10
Added switch recommendation note to TS130F_dual

MacManas:
I believe this information may affect people's decision on whether go for this device or not


Koenkk:
Thanks!

PR #2130, opened 2023-07-13
Update home_assistant.md

pieterdevriesch:
Example mqtt entity configuration for the networkmap is (no longer) correct. Used the example values from @azuwis [documentation](https://github.com/azuwis/zigbee2mqtt-networkmap).

Topic changed to: `zigbee2mqtt/bridge/response/networkmap` and needed additional setting: 
`json_attributes_template: "{{ value_json.data.value | tojson }}"`


Koenkk:
Thanks!

PR #2133, opened 2023-07-17
Heiman HS1SA: Document workaround for firmware bug

klada:
Some of my Heiman HS1SA-E smoke detectors used to send a (wrong) alarm state over ZigBee regularly, without actually triggering their built-in device alarm. It looks like their firmware is somehow buggy but there is an easy workaround for the issue.

Unfortunately the issue is non-deterministic. I have 4 of these devices (same model, same firmware) and only 2 needed this workaround applied. I guess this is why a documentation fix should be sufficient.

Thanks to the deCONZ people for spotting this workaround: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5824#issuecomment-1092089211


Koenkk:
Thanks!

PR #2144, opened 2023-07-24
Update swarm instructions for cgroup v2

tedivm:
I can into issues where the current version of armbian does not enable cgroup v1, but the current instructions for docker swarm didn't work for cgroup v2.

After digging through the issue on github I found that another person already encountered the same issue and created a solution for v2. I helped them publish the docker container, and have been using it locally. I figured I'd update the instructions so the next person who ran into this issue could solve it quickly.


Koenkk:
Thanks!

PR #2156, opened 2023-08-03
Update home_assistant.md - Remove leading device name

andrew-codechimp:
Remove the leading device name from all MQTT sensors/switches to stop warnings in log for HA 2023.8


Koenkk:
thanks!

PR #2162, opened 2023-08-06
Update user_extensions.md

klebom:
stop() {
        this.eventBus.removeListenersExtension(this);
    }

Is simply wrong. Changing to 

    stop() {
        this.eventBus.removeListeners(this);
    }

seems to do the trick...


Koenkk:
Thanks!

PR #2164, opened 2023-08-07
Update Weiser SmartCode 10 docs with more details on pin codes and pairing

JasonSanDiego:
Replaces previous pull request that had updates in the wrong place on the page.


Koenkk:
Thanks!

PR #2165, opened 2023-08-10
Add note to Sengled bulb docs clarifying that they are not Zigbee routers

harrv:
According to the [Sengled FAQ](https://support.sengled.com/hc/en-us/articles/115010871308-Do-any-Sengled-Zigbee-devices-act-as-Zigbee-repeaters-) their bulbs are end devices, not routers. Since it goes against common wisdom that all mains-powered Zigbee devices are routers, exceptions to that commonly held belief should be noted. This pull request contains a single commit that adds the same note to all of the Sengled bulb doc pages.

### DIFF
--- a/docgen/device_page_notes.ts
+++ b/docgen/device_page_notes.ts
@@ -81,6 +81,14 @@ Note: if \`hue_power_on_behavior\` is set to \`off\`, then the only way to turn 
 `
         }
     },
+    sengledBulbs: (definition) => {
+        if (definition.vendor === 'Sengled' && definition.exposes.find((e) => e.type === 'light')) {
+            return `
+## Device Type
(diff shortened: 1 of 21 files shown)


Koenkk:
Could you add this to `device_page_notes.ts` like:

https://github.com/Koenkk/zigbee2mqtt.io/blob/d09d6e34cdfe144e3db92530b81389d1273f63db/docgen/device_page_notes.ts#L12

This will ensure docs get added to all current and in the future to be supported devices.

harrv:
> Could you add this to `device_page_notes.ts` like:
... 
> This will ensure docs get added to all current and in the future to be supported devices.

Yes, that seems like a good approach. This is my first contribution so I'm figuring out how to build the docs to test the change and will update the PR when ready.

harrv:
I got it figured out and committed, but my dumb editor reformatted most of the lines of `/docgen/device_page_notes.ts`. That was unintended. I've converted the PR to draft until I can figure out a good way to revert the unintended reformatting.

harrv:
OK, should be good now. Let me know if you need any more changes.

harrv:
I should clarify that since the *.md files generated by `npm run docgen` are also committed to this repo, I first made the change you requested to `docgen/device_page_notes.ts`, then ran `npm run docgen`, and then committed both `docgen/device_page_notes.ts` and the 20 generated files in `doc\devices` that were affected by the change. I hope that's what you wanted.

Koenkk:
Perfect, thanks!

PR #2166, opened 2023-08-11
Add a new picture for ME168

dyx052612:
Add a new picture for ME168


dyx052612:
Added a supported device picture

dyx052612:
Because the appearance and models of these two products are different, two different pictures and introductions are required.

Koenkk:
Can you improve the picture?

- Make the background fully white
- Increase size of the radiator

Koenkk:
Thanks!

PR #2180, opened 2023-08-20
List UZG-01 mDNS Zeroconf availability

mercenaruss:
List UZG-01 mDNS Zeroconf availability.
Was tested locally with succes.


Koenkk:
Thanks!

PR #2188, opened 2023-08-23
Add support for Namron water leak sensor 4512764

einarjh:
Adds notes and image for the new device Namron water leak sensor 4512764 - corresponds to pull request https://github.com/Koenkk/zigbee-herdsman-converters/pull/6101


Koenkk:
Thanks!

PR #2194, opened 2023-08-29
Added image for Bosch Room Thermostat II

miczab:
There are two variants of the Bosch Room Thermostat II which are basically the same. One has battery power and the other one mains power. The added image is for the latter one.


Koenkk:
This picture will not be used, merge it with `BTH-RM.jpg`

miczab:
How? I think I should add a new zigbee model (RBSH-RTH0-ZB-EU), because this variant has no 'battery_low' and 'voltage' properties. If I link the new device to an already present model (BTH-RM) the test will fail.

Koenkk:
I checked again but this image is already available: https://github.com/miczab/zigbee2mqtt.io/blob/miczab-patch-1/public/images/devices/BTH-RM.jpg

miczab:
Yes, I know. But what do you suggest in that case? If I want to add a new device to the Bosch vendor file in [zigbee-herdsman-converters](https://github.com/Koenkk/zigbee-herdsman-converters) I need an new image for it that matches the new model name. Or should I merge it into the alraedy available model BTH-RM and ignore the fact that the new device haven't got battery properties like BTH-RM?

Koenkk:
> If I want to add a new device to the Bosch vendor file in [zigbee-herdsman-converters](https://github.com/Koenkk/zigbee-herdsman-converters)

In this case a new image is needed indeed, could you link me the PR?

miczab:
OK, I just made a PR: https://github.com/Koenkk/zigbee-herdsman-converters/pull/6120

Koenkk:
Thanks!

PR #2202, opened 2023-09-02
Update E2112.md

201505409:
Added notes, to warn possible buyers that the device cannot currently report with decimal precision


Koenkk:
Thanks!

PR #2213, opened 2023-09-08
Update linux install instructions

juantxorena:
We need to build the app before being able to run it.

This will fix https://github.com/Koenkk/zigbee2mqtt/issues/18812


Koenkk:
thanks!

PR #2215, opened 2023-09-08
Update WM25L-Z.md - add reversal notes

amaisano:
Adding directions for the WM24F/Z which should also work for the WM25. Based on the included direction booklet and tested to work.

NOTE: This may or may not affect the remote and/or how other smart home hubs interpret the commands.


Koenkk:
Thanks!

PR #2220, opened 2023-09-17
Picture for Cleverio RH/Temp Sensor SS300

nightcbis:
Cleverio RH/Temp Sensor SS300


Koenkk:
Please rename the picture to `SS300.jpg` (capitals)

PR #2227, opened 2023-09-20
Update copy IEEE instructions

guillochon:
Added a note that some adapters have an autobootloader and don't require being put into bsl mode.


Koenkk:
Thanks!

PR #2234, opened 2023-10-01
Add some help for EZSP devices using conflicting USB identifiers.

tschiex:
Plus a hint that the right ZigBee channel needs to be given to the EZSP sniffer.


Koenkk:
Thanks!

PR #2235, opened 2023-10-02
fix(image): remove duplicate

Rotzbua:
## Problem

The pictures are already available with `.jpg` ending.

## Solution

Remove duplicate pictures with `.jpeg` ending.


Koenkk:
thanks!

PR #2257, opened 2023-10-08
Update LED1624G9.md

louiecaulfield:
Describes issue regarding minimal brightness when coming from OFF-state


Koenkk:
does this also happen with other tradfri bulbs?

louiecaulfield:
In the Tradfri series, I only own 2x LED1836G9 and 1x ICPSHC24-30EU-IL-1

They both don't exhibit this behaviour, i.e. when setting the brightness to 1 while they're in an OFF state, they simply turn on at brightness 1.

Koenkk:
Is there no OTA update available for this device? (check the OTA updates tab of the z2m frontend)

louiecaulfield:
> Is there no OTA update available for this device? (check the OTA updates tab of the z2m frontend)

I did - it's running firmware version 2.3.093

Koenkk:
Thanks for checking!

PR #2258, opened 2023-10-09
Update Hue Wall Switch Binding

jamesonuk:
If you add switch to group you can control the group it is controlling


Koenkk:
Nice finding, thanks!

PR #2280, opened 2023-10-21
Update Linux Setup nodejs version

ferbar:
https://github.com/Koenkk/zigbee2mqtt/blob/master/package.json#L11

should be v16-20


Koenkk:
thanks!

PR #2286, opened 2023-10-26
Add troubleshoot note for random automation exceptions

Pathen85:
Automations in restart mode will often crash when being triggered by changes of the action attribute of the remote, because of the release action that happens immediately after the button press itself.

This problem is analog to what is described in this [issue](https://github.com/home-assistant/core/issues/98664).


Koenkk:
Thanks!

PR #2289, opened 2023-10-28
Update adapter-settings.md

Multipas84:
sky connect


Koenkk:
Whether it is `ttyACM0` or `ttyUSB0` depends on the OS, this page is also not specific to the SkyConnet so I propose to not merge this.

PR #2292, opened 2023-10-28
Warning about BRT-100-TRV firmware updates

gyver:
Some revisions have been updated to incompatible firmware versions, warn users that a firmware update can brick their device.

### DIFF
--- a/docs/devices/BRT-100-TRV.md
+++ b/docs/devices/BRT-100-TRV.md
@@ -25,6 +25,16 @@ pageClass: device-page
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
 ## Notes
 
+### Firmware updates
+
+In the past there have been at least 2 revisions of this device where Zigbee2MQTT detected new firmwares that were incompatible with the device. See :
+* [BRT-100-TRV OTA broken / flash ZT5](https://github.com/Koenkk/zigbee2mqtt/issues/19173)
+* [Moes BRT-100-TRV not reporting after OTA update](https://github.com/Koenkk/zigbee2mqtt/issues/7941)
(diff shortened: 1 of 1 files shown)


docs/devices/BRT-100-TRV.md
```diff
@@ -25,6 +25,16 @@ pageClass: device-page
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
 ## Notes
 
+### Firmware updates
+
+In the past there have been at least 2 revisions of this device where Zigbee2MQTT detected new firmwares that were incompatible with the device. See :
+* [BRT-100-TRV OTA broken / flash ZT5](https://github.com/Koenkk/zigbee2mqtt/issues/19173)
+* [Moes BRT-100-TRV not reporting after OTA update](https://github.com/Koenkk/zigbee2mqtt/issues/7941)
+
+In this situation a firmware update made the device unable to connect to the Zigbee network. As of 2023/10/28 Zigbee2MQTT has been updated to support the first revision with this problem but the last one will only be correctly supported in the next release.
```
Koenkk, line 34: OTA support has been completely disabled for this device, so I'm not sure what you mean with "supported" here.

gyver: I agree "supported" isn't the right term. Explicitly stating that the OTA support has been disabled is probably better, I'll update and resubmit.

zatramanh56789: OTA (Over the Air) update is the process of uploading firmware to an ESP module using a Wi-Fi connection rather than a serial port. Such functionality becomes extremely useful in case of limited or no physical access to the module.

PR #2296, opened 2023-10-30
Add QUAD-ZIG-SW and update MULTI-ZIG-SW

mortenmoulder:
New device added (QUAD-ZIG-SW) and updated previous documentation of MULTI-ZIG-SW


Koenkk:
Thanks!

PR #2299, opened 2023-10-31
Philips / Signify 4" Recessed

thk-socal:
Image file for Zigbee model LTD016 and Philips / Signify model 9290031347


Koenkk:
Thanks!

PR #2301, opened 2023-11-02
Update homeassistant.md - HASS discovery topic note

niawag:
Added a note to HASS discovery topic to prevent errors due to using the same topic for HASS discovery and base topic. (error made by me... https://github.com/domoticz/domoticz/issues/5859)


Koenkk:
thanks!

PR #2303, opened 2023-11-05
Philips Festavia 500 Light

thk-socal:
Image for Festavia 500 lights.


Koenkk:
Could you change the file extension to `.jpg`?

PR #2314, opened 2023-11-06
Update 4523430.md

LarsOlafSimonsen:
Notes: This device does not seem to support Zigbee channels higher than channel 15. If your network uses a higher Zigbee channel, you might not be able to join this device to the network.


Koenkk:
Thanks!

PR #2339, opened 2023-11-18
Update E1525_E1745.md

dg7pc:
Add a second option to pair the device. Needed on two new bought devices with Firmware 2.0.022

I had some  problems to pair my two new devices pressing the pair button 10sec way.


Koenkk:
thanks!

PR #2343, opened 2023-11-21
Update 01_linux.md -- Add sysvinit script

StafrGotar:
For systemd-free Linux distributions, such as Devuan or equivalent, I've added instructions for how to start and stop Zigbee2MQTT as a SysV service.


Koenkk:
Please move this to a new page and link to it it under `(Optional) Running as a daemon with systemctl` with something like: 

```
NOTE: In case you want to use sysvinit, see this page
```

PR #2348, opened 2023-11-22
New 8ch switch (TS0601)

bszczepanik:
I bought new 8 channel zigbee switch in Poland. I think that is this from alibaba https://www.alibaba.com/product-detail/Tuya-smart-wifi-zigbee-Bluetooth-4CH_1600834099744.html. Of course here we have it with polish label. 

I would like to add support for this device. I'm not sure if I must upload image first or configuration.


bszczepanik:
This is picture for switch, which configuration is merged today.

Koenkk:
Thanks!

PR #2376, opened 2023-12-05
Add new device:SONOFF S60ZBTPF/S60ZBTPG

lcheng33775823:


Koenkk:
Thanks!

OJFord:
Do Zigbee variants (S60ZB) of the S60 plug actually exist? The only reference I can find is zigbee2mqtt.io, not for sale anywhere, not on Sonoff's site.

The S60TPG (TPG is Type G plug - i.e. a UK plug) I bought hoping to use with Z2M seems to be Wi-Fi only; on closer inspection of the product page I realised it didn't actually mention Zigbee at all, I got that purely from zigbee2mqtt.io.

PR #2378, opened 2023-12-05
Added image for namron_4512771

larsnygard:


Koenkk:
- rename to `4512771.jpg`
- do you have a better picture? one with a white background.

larsnygard:
Edited with a white background. 
There are device pictures on the Namron pages, but I'm not sure if it's allowed to pull images from there for copyright issues, so I snapped one with my phone. 

https://www.namron.com/products/namron-zigbee-multisensor-hvit/

Koenkk:
> There are device pictures on the Namron pages, but I'm not sure if it's allowed to pull images from there for copyright issues, so I snapped one with my phone.

We do that for all devices already, so let's just do that. Please also add one for the white version.

PR #2383, opened 2023-12-06
Update home_assistant.md for new bridge entities

mundschenk-at:
Adapts Home Assistant integration notes for new bridge entities

### DIFF
--- a/docs/guide/usage/integrations/home_assistant.md
+++ b/docs/guide/usage/integrations/home_assistant.md
@@ -168,18 +168,6 @@ The following Home Assistant configuration allows you to control Zigbee2MQTT fro
 You can add it to the appropriate section of your `configuration.yaml`, or you can add it as a [Home Assistant Package](https://www.home-assistant.io/docs/configuration/packages/) by adding the following to `zigbee2mqtt.yaml` in your packages folder.
 
 ```yaml
-# Input select for Zigbee2MQTT debug level
-input_select:
-  zigbee2mqtt_log_level:
-    name: Zigbee2MQTT Log Level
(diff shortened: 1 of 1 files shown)


waclaw66:
@mundschenk-at: Is there any particular reason why you've removed join timer with this PR? I'm configuring a new z2m instance and noticed that feature is missing now in the notes.

mundschenk-at:
> @mundschenk-at: Is there any particular reason why you've removed join timer with this PR? I'm configuring a new z2m instance and noticed that feature is missing now in the notes.

Because a sensor entity for that is now automatically discovered on the bridge.

waclaw66:
Timer was working fine before. Still don't understand why there is still `input_number.zigbee2mqtt_join_minutes` when the timer automation was removed.

waclaw66:
> Not sure what you want here. You are free to use old the automation and and template entities if you want, they are just not necessary anymore because the bridge discovery now includes those features by default.
> 
That's I'm not sure, join with timer doesn't work anymore. If that was removed intentionaly, then ok, I can still use the old automation.

mundschenk-at:
Yes, if you want custom discovery times you need to use the old automations. If there is a countdown, it will show up as a sensor value, but the HA switch entity toggles the state permanently (because it's easier to add fine-grained control via scripts/automations). 

There's been [some discussion](https://github.com/Koenkk/zigbee2mqtt/discussions/20468#discussioncomment-8005781) on having a second switch entity with the default timeout, but there didn't seem to be that much demand for that feature in the end, so it didn't get implemented (yet).

PR #2392, opened 2023-12-09
Documents on with timed off for switches and lights

ghoz:
The feature is rediscovered from time to time
* https://github.com/Koenkk/zigbee-herdsman-converters/pull/2326
* https://github.com/Koenkk/zigbee2mqtt/issues/3747
* https://github.com/Koenkk/zigbee2mqtt/discussions/19691

Unfortunately not all devices support this,
so this might need refinement to opt out devices


Koenkk:
Thanks!

PR #2394, opened 2023-12-10
Update CCTFR6100Z3.md

jj-uk:
Added notes to describe device operation and how to factory reset and bind the device.


Koenkk:
Thanks!

PR #2403, opened 2023-12-13
Add picture for 808EZ

Chartreusito:
Converter incoming


Chartreusito:
Well, this device shares the same Zigbee Manufacturer as [809WZT ](https://www.zigbee2mqtt.io/devices/809WZT.html#tuya-809wzt) : _TZ3040_bb6xaihh
I guess we can't differentiate them.

Koenkk:
It's the same indeed so unfortunately not.

PR #2404, opened 2023-12-13
Add Müller Licht - tint LED-Strip white+color, 3 m / 6W RGB

rcmcronny:
Bought this thing and as it is not supported, i will add it.
Converter File will follow shortly


Koenkk:
Thanks!

PR #2412, opened 2023-12-14
New Device TH05Z

xs400dohc:
New Device TH05Z, another temperature/humidity sensor with display and clock. Similar to JM-TRH-ZGB-V1 / SZ-T04 / TH01Z.
See pr #6705 @ Koenkk/zigbee-herdsman-converters


Koenkk:
Thanks!

PR #2433, opened 2023-12-23
Image for Hue MR16/GU10/LCG005

thk-socal:


thk-socal:
I mistyped GU10 instead of GU5.3 in the title of this PR. But all internal references are GU5.3 as they should be in the files.

Koenkk:
Thanks!

PR #2445, opened 2023-12-31
Update ERS-10TZBVB-AA documentation

chrisbraucker:
Button supports triple-click that changes the mode of operation. This is not well documented and caused me quite the head ache.


Koenkk:
Thanks!

PR #2455, opened 2024-01-03
Update A1Z.md add "reset-energy"

Marlor:
Same function as https://www.zigbee2mqtt.io/devices/TS011F_plug_1.html#reset-energy
Found here https://forum.iobroker.net/topic/65461/nous-a1z-steckdose-energiewert-auf-0-setzten/8
Testet it myself


Koenkk:
thanks!

PR #2464, opened 2024-01-05
Update Philips 929003017102 Hue Wall Switch Module docs

marazmarci:
I discovered that [this device](https://www.zigbee2mqtt.io/devices/929003017102.html) supports the `LevelCtrl` cluster for bound groups! I extended the binding instructions accordingly.

(I also did some boy scouting and put some effort into improving the documentation's phrasing and grammar, I hope I didn't make it worse :sweat_smile:)

### DIFF
--- a/docs/devices/929003017102.md
+++ b/docs/devices/929003017102.md
@@ -32,7 +32,7 @@ Out of the box, the device enters pairing mode, as soon as the pins of input 1 a
 If the device has not been paired yet, it can also be put into pairing mode by short pressing the reset button.
 
 ### Directly control Zigbee groups
-It's possible to configure this switch to directly control Zigbee groups. In this way, the switch can control e.g. a group of light bulbs even when Zigbee2MQTT is not running. To accomplish this in the Zigbee2MQTT frontend:
+It's possible to configure this switch to directly control Zigbee groups. In this way, the switch can control e.g. a group of light bulbs even when Zigbee2MQTT is not running, and it can also make controlling the lights feel a bit more responsive. To accomplish this in the Zigbee2MQTT frontend:
 1. Go to the *Groups* tab and create a new group if it does not already exist. Add devices to the groups that you want to be controlled by this switch.
 1. Add this device to the group. Note that this device will not show up in the group, but this step is important.
 1. Click on this device and go to the *Bind* tab. Modify the checkboxes so that only `manuSpecificPhilips`, `OnOff`, and `LevelCtrl` (optional ¹) are checked, and click *Unbind* in the row where the destination is `Coordinator`.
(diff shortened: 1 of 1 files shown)


docs/devices/929003017102.md
```diff
@@ -28,22 +28,23 @@ pageClass: device-page
...
 
 ### Directly control Zigbee groups
-It's possible to configure this switch to directly control Zigbee groups. In this way the switch can control e.g. a group of light bulbs even when Zigbee2MQTT is not running. To accomplish this in the Zigbee2MQTT frontend:
-1. Go to the *Groups* tab and create a new group if it does not already exist. Add devices to the groups which you want to be controlled by this switch.
+It's possible to configure this switch to directly control Zigbee groups. In this way, the switch can control e.g. a group of light bulbs even when Zigbee2MQTT is not running, and it can also make controlling the lights feel a bit more responsive. This switch can also act as a dimmer: for each successive long press of the switch, it will alternate between increasing and decreasing brightness. To set this all up, follow these steps in the Zigbee2MQTT frontend:
+1. Go to the *Groups* tab and create a new group if it does not already exist. Add devices to the groups that you want to be controlled by this switch.
 1. Add this device to the group. Note that this device will not show up in the group, but this step is important.
```
Koenkk, line 37: Since you are at it, can you make this change? I believe the wall switch shows up in the group (at least does in my case)

```suggestion
1. Add this Hue wall switch to the group.
```

marazmarci: I just checked my group which is controlled by the Hue Wall Switch Module, and I can't see the device being in the group. Maybe it only shows up just after being added? And probably disappears after a restart? I'll validate this theory soon.

marazmarci: I tried it, and I couldn't get the device to show up in the group.

Seemingly the device doesn't respond to the group add request, but handles it silently. I got an error every time I tried. I only copied the last error log, but I think it was the same every time:

```
Failed to add from group (Command 0x001788010cc233b8/1 genGroups.add({"groupid":111,"groupname":""}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Timeout - 54334 - 1 - 70 - 4 - 0 after 10000ms))
```

marazmarci: And I don't remember it ever showing up in the group.

Koenkk: In my case it does, so I suggest to use:

```
1. Add this device to the group. Note that this device might not show up in the group, but this step is important.
```

marazmarci: I applied your latest suggestion and the first part of your original suggestion.

Koenkk:
Thanks!

PR #2465, opened 2024-01-05
Sonoff ZBDongle Rotuer Firmware: fixes putty font error, firmware don…

niemsoen:
…wload command

Adds Router Firmware download command and fix for putty font error


Koenkk:
Thanks!

PR #2468, opened 2024-01-06
added info for other OS

oszczech:
commands provided look different for most non debian bases systems. Added what worked for me on arch based system


Koenkk:
Thanks!

PR #2482, opened 2024-01-10
Add Schneier electric 41ECSFWMZ-VW image

lucasteligioridis:


Koenkk:
Please resize the picture to 150x150

lucasteligioridis:
@Koenkk done, thanks!

I also tried to generate docs but I believe the code needs to be merged into the herdsmen repo for it to generate correctly?

If thats the case, the documentation on the site needs to be updated? If its not possible to have docs generated without the device merged into trunk then maybe that should be made clear, I was a bit confused initially anyway!

Koenkk:
Thanks!

PR #2485, opened 2024-01-12
Add Yale NF-YRD622 TSDB Image

arunbala81:
Adding new Yale supported device


Koenkk:
There is no picture added, could you fix this PR and revert changes to `exposes.md`?

PR #2486, opened 2024-01-12
DJT11LM.md: document vibration event frequency limit

strix-technica:
In experimentation after having first received this device, I could not understand why sometimes `vibration` events would register and other times not, even though `tilt` events would be emitted every time the device was reoriented.  Eventually I found from [Aqara's product FAQ](https://www.aqara.com/eu/product/vibration-sensor/faq/#accordion-3) that vibration events are rate-limited to once per minute.


Koenkk:
Thanks!

PR #2491, opened 2024-01-13
Add WSDCGQ11LM pairing tip: use nearby router to connect to

BrianHenryIE:
Add a tip to the pairing instructions.

These devices are notoriously bad for pairing.

After trying many, many times to pair one of these sensors, I was able to immediately pair it by selecting a specific Zigbee router device rather than using "All" or "coordinator". 

Screenshot 2024-01-13 at 11 14 33 AM

Screenshot 2024-01-13 at 11 14 43 AM


Koenkk:
Thanks!

PR #2497, opened 2024-01-17
Clarified token is cleartext

JsBergbau:
Added description that token is in cleartext, because I configure a lot with passwords and mainly you have to hash them. Using the word `token` instead of password here, confused me and I started to google a lot how to hash the password for the token. 

To avoid this hassle for others I've updated the docs.


Koenkk:
Thanks!

PR #2513, opened 2024-01-22
Added notes on 1.2v AAA batteries for VALLHORN

charnley:
HA forum is full of example where switching to 1.2v AAA batteries for VALLHORN fixes unexpected behavior (myself included).


Koenkk:
Thanks!

PR #2515, opened 2024-01-23
Add image for Linkoze LKWSZ211 2-button switch

kiler129:
### PR contents
Add image for `Linkoze LKWSZ211` Zigbee scene switcher with 2 buttons. I adjusted the curves of the image a bit so it's visible on the pure white background.

### Also sold as....

 - Generic Aliexpress specials (e.g. [`Tuya Smart WiFi/Zigbee Switch Push Button Switch 2Gang 6 Scene Wireless Smart Home Remote Controller Automation Scenario Switch`](https://a.aliexpress.com/_mM3HCf6))
 - `Adaprox` brand as [`Wireless Switch (2-Key)`](https://www.adaprox.io/products/linkoze-wireless-switch-2-key?variant=41178716700877)
 - `Adaprox` version seems to be particularly popular in Europe, e.g. large Polish e-retailer [Kamami](https://kamami.pl/automatyka-domowa/1184385-lkwsz211-bezprzewodowy-przelacznik-zigbee.html) stocks plenty of these; Polish equivalent of eBay (Allegro) is also full of `Adaprox` versions.
 
The device is also sometimes advertised under `LKWSW201`, but the `...201` model is actually the WiFi version of the same switch.

### OEM original packaging 
image
image
image image

### Docs/quirks

- I didn't know how to/if I should add a note about peculiar behavior of the button hold. This device only sends a single "button was held" event on release. The button has to be held for 4-5s until LED turns off.
- To reset or pair the device, one needs to press and hold the button with an indentation for >10s, past the point of LED turning off, and waiting until it blinks.
- Official manual: https://www.meshhub.com/manuals/lkwsz211-en

### Related PRs
As per [Zigbee2MQTT instruction](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html) I also added PR with the actual device definition. See https://github.com/Koenkk/zigbee-herdsman-converters/pull/6944


Koenkk:
Thanks

k1kuu:
All my buttons stopped working. I think this is the latest update for them? 

No converter available for 'LKWSZ211' with cluster 'genOnOff' and type 'raw' and data '{"data":[1,8,253,0],"type":"Buffer"}'

PR #2517, opened 2024-01-26
Update 324131092621.md with battery type

juanrodgil:
Added battery type in description of Philips Hue dimmer switch 324131092621


Koenkk:
Thanks!

PR #2527, opened 2024-01-31
added SSW06G device image

SeanoNET:


Koenkk:
Could you remove the Zigbee logo from it?

PR #2530, opened 2024-01-31
Update 57004000.md

aw1604:
added notes on the device internal debounce in switch mode which causes delays,
and showed an alternative to use sensor mode for quicker action.


Koenkk:
Thanks!

PR #2538, opened 2024-02-03
Update Linux installation guide to add libsystemd-dev when installing required packages

Nardol:
To use new SystemD unit, libsystemd-dev must be installed.
Since it is not part of the apt-get command to run, SystemD service fails.


Koenkk:
Thanks!

PR #2539, opened 2024-02-03
Fix device card description size and text overflow problems

mrskycriper:
Small fix that turns this

Снимок экрана 2024-02-03 в 21 51 53

into this

Снимок экрана 2024-02-03 в 21 52 33

for all devices with long list of exposes.


Koenkk:
Thanks!

PR #2544, opened 2024-02-04
update aqara z1 switches images

waterlou:


Koenkk:
Could you make this against `develop` with 512x512 `.png` with a transparent background?

PR #2546, opened 2024-02-05
Update states template

TheGroundZero:
> Avoid using `states.sensor.temperature.state`, instead use `states('sensor.temperature')`. It is strongly advised to use the `states()`, `is_state()`, `state_attr()` and `is_state_attr()` as much as possible, to avoid errors and error message when the entity isn’t ready yet (e.g., during Home Assistant startup).

https://www.home-assistant.io/docs/configuration/templating/


Koenkk:
Thanks!

PR #2547, opened 2024-02-05
Update NLG-TW_light.md with pairing information

Gulpman:
Pairing information for Paul Neuhaus Q-lights added. 
Two possibilities: Either via a Q-remote or via a switch sequence with a switch connected to the light.


Koenkk:
Thanks!

PR #2549, opened 2024-02-06
Update Sinopé ECO Mode

MivraMe:
Add to the docs the way to activate the ECO mode on most of the Sinopé Thermostat


Koenkk:
Thanks!

PR #2551, opened 2024-02-06
Update PCI_E.md with pairing instructions

BenJamesAndo:
Added pairing instructions.
I also removed `ON means no buzzer noise.` Turns out I was wrong and ON in fact does mean buzzer noise is on.

### DIFF
--- a/docs/devices/PCI_E.md
+++ b/docs/devices/PCI_E.md
@@ -44,6 +44,7 @@ To write (`/set`) a value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/
 If value equals `ON` state is ON, if `OFF` OFF.
 
 ### Buzzer feedback (binary)
+Enable buzzer feedback. It sounds on device actions like power state changes, child lock activation, etc.
 Value can be found in the published state on the `buzzer_feedback` property.
 It's not possible to read (`/get`) this value.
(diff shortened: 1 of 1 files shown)


docs/devices/PCI_E.md
```diff
@@ -40,7 +44,6 @@ To write (`/set`) a value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/
 If value equals `ON` state is ON, if `OFF` OFF.
 
 ### Buzzer feedback (binary)
-ON means no buzzer noise.
```
Koenkk, line 43: What's the correct description for buzzer feedback?

BenJamesAndo: When it's enabled it basically just makes a noise whenever a control is changed. Perhaps something like the following would do?

Enable buzzer feedback. The buzzer will make a noise whenever an action is carried out via this device such as the power state changing, child lock being enabled etc.

Koenkk: Sounds good, then also please make a PR to update it [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/41c2143627699db8cf35e1524ee120c7c7f1f825/src/devices/weten.ts#L28)

BenJamesAndo: Done. Linked PR at https://github.com/Koenkk/zigbee-herdsman-converters/pull/7058

Koenkk:
Thanks!

PR #2555, opened 2024-02-07
fix: transparancy issues for a few Niko, Ikea, and Develco devices

sjorge:
Since the latest release a look of icons looked really bad, they were switched to transparent ones.

I spend a few minutes manually tracing the ones I saw broken in my setup.


Koenkk:
Thanks!

PR #2558, opened 2024-02-08
Add missing documentation to Linux installation to install requirements of sd-notify

IgnacioHR:
If you just follow the instructions. sd-notify is not installed because UNMMET REQUIREMENTS. As it is an optional package this sould be OK. But if you want to tun systemctl with Type=notify you need to install the requirements.


docs/guide/installation/01_linux.md
```diff
@@ -180,6 +180,11 @@ WantedBy=multi-user.target
 
 Save the file and exit.
 
+You need some __systemd__ development files, on __Ubuntu__ these can be installed via:
+```
+$ sudo apt install libsystemd-dev
```
Koenkk, line 185: There is already an `apt-get install` on line 39, can you add it there?

Koenkk: Agree, thanks!

PR #2566, opened 2024-02-12
fix: illegal character `:` in image filenames

pano9000:
There were a handful of "leftover" images, that contained a colon `:` in their file name.
This is no problem for Linux environments, but for Windows Environment the colon is not an allowed character for a filename.

And that actually seems to be causing the repository to not be cloneable, when using Windows, where git complains "error: invalid path 'public/images/devices/AU-A1GSZ9B:27.png'"

This commit fixes this issue, by deleting these files, simply to the fact that they are *unused*.

I have double-checked all of the files/filenames:
* none of them are currently linked in any of the other files (according to a grep search)
* the images were always a duplicate anyways, with the alternative files using a "safe" hyphen `-` instead of the `:`


Koenkk:
Thanks!

PR #2570, opened 2024-02-13
Update 01_linux.md with curl note

Impact123:
Someone in the discord server just ran into a problem installing it due to curl missing on their system. This simply adds a note on how to install curl which might help a little bit or we can refer to it.


Koenkk:
Thanks!

PR #2571, opened 2024-02-14
Update MFKZQ01LM.md

Gulpman:
Added notes for battery type and pairing process. Just copied that part from the Aqara MFCZQ12LM notes section as battery and pairing process are identically for these devices.


Koenkk:
Thanks!

PR #2557, opened 2024-02-08
fix: Update and replace Aqara device images

mrskycriper:
This PR is all about cleaning up the mess that is recent png images implementation. A lot of them turned into blobs of color. I've replacing them with the combination of original png froma Aqara web sites, Photoshop background cuts and some images found on the internet.

# Convert

- delete jpgs
- replace corrupted and blurry png
- resize to 512x512

# Backlog

- [x] QBKG11LM QBKG12LM
- [x] QBKG03LM QBKG04LM 
- [x] WXKG02LM
- [x] WXKG03LM
- [x] WXKG04LM
- [x] WXKG05LM
- [x] QBKG17LM QBKG19LM 
- [x] QBKG18LM QBKG20LM
- [x] QBKG33LM QBKG34LM
- [x] WXCJKG11LM
- [x] WXCJKG12LM
- [x] WXCJKG13LM

### H1

- [x] WS-EUK01 WS-EUK03
- [x] WS-EUK02 WS-EUK04
- [x] WXKG14LM
- [x] WXKG15LM WRS-R02
- [x] QBKG27LM
- [x] QBKG29LM QBKG32LM
- [x] ZNXNKG02LM
- [x] ZNQBKG24LM
- [x] ZNQBKG25LM WXKG22LM
- [x] ZNQBKG26LM

### D1

- [x] WXKG06LM
- [x] WXKG07LM
- [x] QBKG21LM QBKG23LM
- [x] QBKG22LM QBKG24LM
- [x] QBKG25LM QBKG26LM

### E1

- [x] WXKG16LM
- [x] WXKG17LM
- [x] QBKG38LM QBKG40LM
- [x] QBKG39LM QBKG41LM
- [x] ZNQBKG31LM

### US

- [x] WS-USC01 WS-USC03
- [x] WS-USC02 WS-USC04

### Buttons

- [x] WXKG11LM WXKG12LM WXKG13LM WXKG20LM

## Sensors

- [x] WSDCGQ11LM WSDCGQ12LM TH-S02D
- [x] RTCGQ11LM RTCGQ12LM RTCGQ15LM
- [x] RTCGQ13LM
- [x] RTCGQ14LM MS-S02
- [x] RTCZCGQ11LM
- [x] SJCGQ11LM SJCGQ12LM WL-S02D SJCGQ13LM
- [x] DJT11LM DJT12LM
- [x] MCCGQ11LM MCCGQ12LM MCCGQ14LM
- [x] VOCKQJK11LM AAQS-S01
[...]

PR #2573, opened 2024-02-14
fix: add Aqara jpgs back

mrskycriper:
Added back JPG pictures of Aqara devices for backwards compatibility.

They were previously removed in #2557


Koenkk:
Thanks!

PR #2576, opened 2024-02-15
Update frontend.md

DubhAd:
Added a warning about not setting `host` for Docker based installs - seen this trip up too many people


Koenkk:
Thanks!

PR #2579, opened 2024-02-17
Adding images for [TS0601 /  _TZE200_ntcy3xu1]  and [TS0205 / _TZ3210_up3pngle],

marcelhoogantink:
Both devices have identical cases but different electronics and logic


Koenkk:
Bedankt!

PR #2580, opened 2024-02-17
fix: reverse ZNQBKG24LM png image regression

mrskycriper:
Fixes problem with ZNQBKG24LM png image introduced in #2574.


Koenkk:
Thanks!

PR #2581, opened 2024-02-18
Device image for samotech SM325

alexeiw123:
added device image for samotech SM325


Koenkk:
Could you make the background transparent using https://www.adobe.com/express/feature/image/remove-background

alexeiw123:
> Could you make the background transparent using https://www.adobe.com/express/feature/image/remove-background

Yes - my mistake. I removed the background then reduced the resolution, not realising it lost the transparent background. on it.

Koenkk:
Thanks!

PR #2584, opened 2024-02-20
Add files via upload

alexeiw123:
Change SM325 device image from black to white - it looks better in the dashboard and probably more commonly used.


Koenkk:
thanks!

PR #2588, opened 2024-02-22
Add picture and .md notes for Paulmann 500.46

einFreak:
New device info for new Paulmann 500.46 device

500.46 was reported as 929.63 before with wrong description etc.
upcomming herdsmann-converters pull request will reference this pull request


Koenkk:
Thanks!

PR #2590, opened 2024-02-23
Update 100.462.31.md

Gulpman:
Extended description for pairing process: My previous description was valid for a brand new remote. However, if the remote already is connected to a network, trying to pair the remote with my previous description will fail. Now the description should be bullet-proof for new devices and those already connected to a Zigbee network. :)


Koenkk:
thanks!

PR #2594, opened 2024-02-26
feat: add endpoints description for wb multisensors

mrskycriper:
Since changing endpoint names will introduce breaking changes, adding endpoints description is the best thing in terms of quicker endpoint identification. Especially with the loud buzzer.


Koenkk:
Thanks!

PR #2597, opened 2024-02-27
fix: update Opple dimmers

mrskycriper:
This PR clears up, fixes and spellchecks Opple dimmer descriptions. It also adds battery replacement instruction, because it's very complicated and absolutely not intuitive.


Koenkk:
Thanks!

PR #2598, opened 2024-02-27
Adding Luminea (Pearl) ZX-5232-675 as white-label of TuYa/TS0203

Zasari:
The Luminea ZX-5232-675 (sold for example by Pearl) is a white-label of the TuYa TS0203


Koenkk:
Please add to https://github.com/Koenkk/zigbee-herdsman-converters/blob/919166ad8f8948be51b20dff32a0986c331ed022/src/devices/tuya.ts#L589 with `tuya.whiteLabel`, you can take the `manufacturerName` from the `data/database.db`

PR #2603, opened 2024-02-29
fix(notes): binding clarifications for hue dimmers

mrskycriper:
Continuing my work on deep analysis of binding commands


Koenkk:
Thanks!

PR #2610, opened 2024-03-03
fix: HESZB-120 has broken transparancy

sjorge:
Fixed broken transparancy for HESZB-120, I had to switch the image to one with slightly more contrast to make tracing the object easier.


Koenkk:
could you fix the merge conflict?

sjorge:
OK that took way to much effort, I basically deleted my commit, rebased on develop, replaced the files and made a new commit -_-

Koenkk:
thanks!

PR #2614, opened 2024-03-06
Add: Images for Z1 Pro (ZNQBKG42LM to ZNQBKG45LM)

LZXCorp:
The images are per spec with transparent background and 512x512 sizing.


Koenkk:
Thanks!

PR #2617, opened 2024-03-07
new adapter added

gam-st:


sergge1:
confirmed

Koenkk:
I believe we don't have a stable FW for this device yet?

gam-st:
> I believe we don't have a stable FW for this device yet?

Already have. Even more, the OTA fw update works on SLZB-06p10...

Koenkk:
Where can I find the FW including the source? It should be added to https://github.com/Koenkk/Z-Stack-firmware

PR #2618, opened 2024-03-08
fix: change image for the SZR07U to the correct one

AntoineGS:
Image can be confirmed here: https://expo.tuya.com/product/1105354


Koenkk:
Thanks!

PR #2621, opened 2024-03-10
Adding high res image for Hue Iris silver limited edition (generation 4)

folta:
Updated existing image and added alternate device id: 929002401101


Koenkk:
Thanks!

PR #2629, opened 2024-03-13
Fix image device TS0012

igo-r:
After the latest optimizations, the image of the TS0012 device has been corrupted. I took the old image and removed the background myself.


Koenkk:
Thanks!

PR #2630, opened 2024-03-14
Add (JPG): JPG Version of Images for Z1 Pro (ZNQBKG42LM to ZNQBKG45LM)

LZXCorp:
I saw that there were jpg fallback imgs so I decided to do it for this as well.

All JPG images are in spec of size 150x150.


mrskycriper:
They are probably not necessary since those are for older versions of Z2M. Those don't have any support for Z1 Pro anyway.

PR #2636, opened 2024-03-16
Bankamp: Add note for reset and pairing

ahemwe:
Source: https://www.bankamp.de/media/3b/ef/3a/1640615074/MAL_2215.3_Grace.pdf


Koenkk:
Thanks!

PR #2655, opened 2024-03-28
Update 404049D.md

kaiseracm:
Hi,

I've both devices MLI-404011 and 404049D from Müller Licht. Changing groups binding with 404011 is no problem, but 404049D is always bound to the default tint groups regardless my binding requests. 
In addition with 404049D it is possible to use the scene buttons with activated group0 and it has a long press feature for some scene buttons.

This information should be added to the documentation.

- Added information about groups
- Added long press scene
- Added pairing

Best regards

Achim


Koenkk:
Thanks!

PR #2665, opened 2024-03-31
Build devcontainer

taasko:
For example GH Codespaces up and running with correct Node.js version.
Install Node.js dependencies automatically on container build as `postCreateCommand`.


Koenkk:
Thanks!

PR #2667, opened 2024-04-01
feat: mention remove background tool alternative

taasko:
Adobe Express requires user account. Recommend https://remove.bg/ tool for quicker image optimization for non-Adobe users.


Koenkk:
It seems this tool is limited to 500x500 (while z2m uses 512x512), to get an image bigger than this you also need an account.

Koenkk:
And it will reduce image quality since the original image is only 500x500 (although diff is small)

PR #2669, opened 2024-04-01
Update E2123.md

LaurentiuGrama:
Update the firmware date for Ikea E2123 and additional info if the device doesn't work after the firmware update


Koenkk:
Thanks!

PR #2701, opened 2024-04-12
Update Secure your Zigbee network page for the now automatically random encryption key

j-vaughn:
Updated network encryption key guidance for updates in https://github.com/Koenkk/zigbee2mqtt/pull/18357

https://github.com/Koenkk/zigbee2mqtt/releases/tag/1.33.0

Release 1.33.0 implements https://github.com/Koenkk/zigbee2mqtt/issues/14868 which automatically generates random key generation on startup.  Guidance that Zigbee2MQTT uses an insecure default key is no longer accurate.


Koenkk:
Thanks!

PR #2704, opened 2024-04-14
Update `E2001_E2002.md` to suggest needed setting

HoroTW:
Add a note that `legacy entity attributes` *must* (I'm not entirely sure if it is always the case but for me and a friend it was the case [Devices from 2022 and 2024]) be active for some attributes.

Because I'm not entirely sure I worded it as a suggestion.

Note: Of course we tried to set the device specific setting `Legacy` to false but we also needed to change the setting in Z2M.

### DIFF
--- a/docs/devices/E2001_E2002.md
+++ b/docs/devices/E2001_E2002.md
@@ -72,6 +72,7 @@ simulated_brightness:
   delta: 20 # delta per interval, default = 20
   interval: 200 # interval in milliseconds, default = 200
 ```
+**Note:** If the `action` entry does not report all the attributes (brightness, action_brightness_delta, ...) you may need to activate the `Home Assistant legacy entity attributes` setting in `Zigbee2Mqtt Settings` > `Home Assistant integration`.
 
 * `legacy`: Set to false to disable the legacy integration (highly recommended), will change structure of the published payload (default true). The value must be `true` or `false`
 


docs/devices/E2001_E2002.md
```diff
@@ -72,6 +72,7 @@ simulated_brightness:
   delta: 20 # delta per interval, default = 20
   interval: 200 # interval in milliseconds, default = 200
 ```
+**Note:** If the `action` entry does not report all the attributes (brightness, action_brightness_delta, ...) you may need to activate the `Home Assistant legacy entity attributes` setting in `Zigbee2Mqtt Settings` > `Home Assistant integration`.
```
Koenkk, line 75: I don't see how setting this option influences the `action` published. Can you check this again?

HoroTW: Hi, thanks for your reply,
Sure, I will also take screenshots with the option on and off, but I'm not sure if I'll be able to do that before the end of the week... since I'm not at home right now.

I also should have been clearer in the pull request here, the observed difference is:
In Home Assistant there is an entry called: `sensor.<device_name>_action`, this entry has a status and attributes and depending on the option I can see some attributes (or not if disabled).

I will also find out if this might only happen on a specific version of the E2002 I have.
I will ping you again when this is done :wink:

Koenkk: It's not recommended to enable this as it will increase the load on HA a lot

HoroTW: I see - that is a pity, I did not notice any more load on HA - but I am running it on a pretty powerful machine...
What would be the best course of action here - as far as I understand the option, it *should* not even be necessary to publish the attributes... I was actually suprised that it changed the attributes like that.... could this be a misconfiguration of the device specification or a bug somewhere else... ?
If it is something like that, maybe we can "fix" it so that the attributes are published without the option?

Also, is it already documented that this option significantly increases the load on HA (I think it is worth documenting as I enabled it without thinking it would have a significant impact - I also understood this guide as expecting this option to be enabled by default: https://www.zigbee2mqtt.io/guide/configuration/homeassistant.html#advanced-configuration "default: shown below" ..... and below it is enabled - maybe that needs to be updated as well...)?

I'm happy to do some investigation and/or documentation work if appreciated.

~Horo

Koenkk: The default when nothing is specified in the `configuration.yaml` is `true` indeed, but nowadays is set to `false` for new installations. If you want to use these attributes, the following is recommended: https://www.zigbee2mqtt.io/guide/usage/integrations/home_assistant.html#via-mqtt

HoroTW: @Koenkk Thanks for your previous answers and sorry to bother you again, but I'm still not sure what to do with this issue/PR.

Today I had another friend who bought IKEA switches and wanted to use the "simulated brightness" as described in the documentation (https://www.zigbee2mqtt.io/devices/E2001_E2002.html).

They thought their switch was broken until I pointed out that they need to turn on the `legacy entity attributes` to use the Simulated Brightness in Home Assistant....

I personally have an automation setup that uses a helper state that represents the "desired state" of a light, and the button only manipulates that state ... So I personally don't need this anymore, but I see people stumbling over this "Simulated Brightness" and don't knowing that they need to enable `legacy entity attributes` to see it in Home Assistant ....

Maybe I'm not good at describing it... see the links where people are finding out that this setting is needed:
- https://github.com/Koenkk/zigbee2mqtt/issues/13981
- https://github.com/Koenkk/zigbee2mqtt/discussions/6091

But in the end, if this is something that is not wanted, I think we should close this PR, keeping it open is not really helping the project!

Thank you for your time and this project! 😊

Koenkk: I think we should add some docs on the [home assistant integration](https://www.zigbee2mqtt.io/guide/usage/integrations/home_assistant.html) page on how to use simulated brightness by using [mqtt](https://www.zigbee2mqtt.io/guide/usage/integrations/home_assistant.html#via-mqtt). Then in the simulated brightness option for each device link to this page.

s00500: hey @Koenkk , any pointers on how to do this unless setting up a manual mqtt sensor ?

PR #2730, opened 2024-04-28
Update ZB-RGBCW.md

Woody14619:
Small change to add details about XY/RGB conversion used for this bulb (and possibly others.)

### DIFF
--- a/docs/devices/ZB-RGBCW.md
+++ b/docs/devices/ZB-RGBCW.md
@@ -56,6 +56,8 @@ This light supports the following features: `state`, `brightness`, `color_temp`,
   - `{"color": {"rgb": "R,G,B"}}` e.g. `{"color":{"rgb":"46,102,150"}}`
   - `{"color": {"hex": HEX}}` e.g. `{"color":{"hex":"#547CFF"}}`
 
+**Note:** The returnd values are always in XY color, a conversion function (in perl) is available to change those back into RGB space for personal verification if desired.
+
 #### On with timed off
 When setting the state to ON, it might be possible to specify an automatic shutoff after a certain amount of time. To do this add an additional property `on_time` to the payload which is the time in seconds the state should remain on.
 Additionnaly an `off_wait_time` property can be added to the payload to specify the cooldown time in seconds when the light will not answer to other on with timed off commands.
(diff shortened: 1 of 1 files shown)


Woody14619:
Small add to make it easier to use RGB settings for this bulb (and maybe others).

docs/devices/ZB-RGBCW.md
```diff
@@ -93,3 +95,36 @@ It's not possible to read (`/get`) or write (`/set`) this value.
 The minimal value is `0` and the maximum value is `255`.
 The unit of this value is `lqi`.
 
+### Conversion from returned XY to RGB
```
Koenkk, line 98: I don't think this should be documented here.

Woody14619: If you knew how hard it was to find how to do this conversion, I think you'd change your mind on that.  There are well over 8 different ways to convert between X/Y and sRGB, and the way Z2M has chosen is not one of the "standard" methods.

Koenkk: But this is generic for all bulbs, right?

Woody14619: The formula in Z2M code for RGB to XY is treating them all the same.  But the response curves for even Singled vs Hue bulbs are quite different. 

The point being, when you put RGB into this bulb, Z2M creates an XY to send.  The bulb replies with a reflection of the XY it achieved.  If you're looking to get RGB back (to validate against what you sent), you never get it, which some home control systems treat as a failure.

If you don't think this should be documented here, I'm fine with placing it elsewhere. 
Do you have any suggestions on where should it go? Minimally though I'd like to link to it from here as reference, since the bulb does exhibit the "takes RGB, but replies in XY" behavior. And this is where people are going to look first for hints on odd behavior.

Koenkk: If I understand correctly, this formula is just for this device?

Woody14619: It is for this and a few other generics (and many LED strips).  It's explicitly _not_ for a number of other bulbs that have their own specific color space (Hue, Linkind, and a couple others).  Most of the latter will reply back with RGB if the user asks for RGB settings.  This one replies to all requests (RGB, raw, etc) with the XY setting only.   So the only way to validate that the RGB/raw value you sent was honored is by converting the returned XY to RGB.

Koenkk: Then let's include it, can you move it above ` Notes END: Do not edit below this line -->`?

PR #2734, opened 2024-05-03
add note to PJ-1203A

schauveau:
Add a note to explain the purpose of the new `late_energy_flow_a|b` options


Koenkk:
Thanks!

schauveau:
@Koenkk  Not sure if you will ever read this since the pull request is closed but there are still issues with the HA integration because PJ-1203A uses attribute names that are not yet supported in `zigbee2mqtt/lib/extension/homeassistant.ts`. I could create a PR to had them there but I would rather have a more generic solution like the one I proposed in https://github.com/Koenkk/zigbee2mqtt/issues/22098 . Did you look at it?

PR #2741, opened 2024-05-07
update SDMO1 imgae

script0803:


Koenkk:
Please make the background transparent, you can use e.g. https://www.adobe.com/express/feature/image/remove-background

PR #2745, opened 2024-05-08
Add a better looking picture

piwi3910:
Added a better looking picture for the TS0013.
The existing one, was a white switch and when removing the background, they also removed the color of the switch itselve.
As such i now added a newer, better looking picture for the documentation


Koenkk:
Could you make the background transparent with https://www.adobe.com/express/feature/image/remove-background ?

PR #2746, opened 2024-05-08
Link directly to fix notes for E2001/2002 remotes

rjhenry:
There have been many responses to the issue since it's closure making it harder to find the actual resolution steps. Linking directly to the comment simplifies this.


Koenkk:
Thanks!

PR #2755, opened 2024-05-16
Update ZBMINIL2.md

dierochade:
Added information for reset via external switch/power cycle. CAn be useful since normally device is installed in a wall mount.


Koenkk:
Thanks!

PR #2760, opened 2024-05-20
Create MLI-404051.png

Chrischan-git:


Chrischan-git:
Hope that's the right way.
Can you please add support for this device?

Here is the external definition:

`const {identify, light} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: ['tint-Spotlights'],
    model: 'tint-Spotlights',
    vendor: 'MLI',
    description: 'Automatically generated definition',
    extend: [identify(), light({"colorTemp":{"range":[153,555]},"color":{"modes":["xy","hs"],"enhancedHue":true}})],
    meta: {},
};

module.exports = definition;`

PR #2774, opened 2024-05-24
Update deconz.md

maxant:
it cost me an hour. i found stuff online about setting the adapter, but it was the baudrate that made it work.

### DIFF
--- a/docs/guide/adapters/deconz.md
+++ b/docs/guide/adapters/deconz.md
@@ -5,6 +5,9 @@
 ```yaml
 serial:
   adapter: deconz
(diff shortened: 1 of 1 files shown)


docs/guide/adapters/deconz.md
```diff
@@ -5,6 +5,9 @@
 ```yaml
 serial:
   adapter: deconz
+  port: /dev/xyz    <<< e.g. ttyUSB0. use `dmesg | grep tty` after plugging the conbee into the usb
+  adapter: deconz
```
Koenkk, line 9: this line is already there

PR #2778, opened 2024-05-29
Update TS0201-z to match other TuYa devices

baod-rate:
Tuya -> TuYa to match other TuYa devices and to avoid multiple vendors on devices page


Koenkk:
Thanks!

PR #2781, opened 2024-05-30
Updated referenced node version in python virtual env installation

mgfcf:
The previously referenced version was not sufficient to fulfill the requirements of zigbee2mqtt.


Koenkk:
thanks!

PR #2796, opened 2024-06-07
Added doc and image for device Tuya YK-S03 Pool monitoring

lauer:
This will be almost a copy of the  BLE-YL01 device


Koenkk:
Thanks!

PR #2801, opened 2024-06-11
Update TH01Z.md

Mosol80:


Koenkk:
Does it give an error when setting e.g. 100?

PR #2807, opened 2024-06-15
Update E1524_E1810.md

GrumpyMeow:
Added some hints to succesfully do an OTA of this remote.


Koenkk:
Could you move this to the ota updates troubleshooting section?

PR #2808, opened 2024-06-15
Update ota_updates.md

GrumpyMeow:
Also added the tip for battery level checking of remotes here.
I did take notice of the previous tip, but didn't expect devices to block the OTA.


Koenkk:
thanks!

PR #2810, opened 2024-06-15
 On Button note for FB20-002

Virusmater:
I added a note about the remote FB20-002. I think it should be mentioned on the page because that issue made the remote useless for me. Probably the issue is not even possible to fix on zigbee2mqtt side.


Koenkk:
thanks

PR #2811, opened 2024-06-16
Add table of known ZGP routers

burmistrzak:
This table by far doesn't include all ZGP-capable routers, only the ones that have been confirmed. 😅

Related: https://github.com/Koenkk/zigbee2mqtt/issues/23031

### DIFF
--- a/docs/advanced/zigbee/01_zigbee_network.md
+++ b/docs/advanced/zigbee/01_zigbee_network.md
@@ -44,7 +44,7 @@ When pairing a Green Power device, you must choose whether translated messages s
 
 #### GP Proxies & Sinks
 
-The following Zigbee routers have been confirmed to support (at least some) ZGP features.
+The following Zigbee routers have already been confirmed to support (at least some) ZGP features. Other devices very likely also work without issues, they just haven't been added to this table yet.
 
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

chris-1243:
@burmistrzak What kind of info should I check, while sniffing traffic, in order to add routers in your table ? As you already know, I do not use Philips Hue as ZGP router in my network. I would like to share the device used but I wish to be sure those are acting correctly.

chris-1243:
One more question:

If the ZGP device is already paired to a router, I guess there is a way to check if the packets are sent as `unicast`

chris-1243:
@burmistrzak  Here are the results of my tries. All done with a PTM215Z

[ZGP router table.txt](https://github.com/user-attachments/files/15903134/ZGP.router.table.txt)

chris-1243:
@burmistrzak There is a firmware update form some Hue devices (1.116.5) which seems to fix the issue.

How would you like to show this to the users ? e.g. [9290022166](https://www.zigbee2mqtt.io/devices/9290022166.html#philips-9290022166) is working as `unicast` firmware 1.116.5 but as `broadcast` firmware 1.116.3...

PR #2812, opened 2024-06-16
Update emberznet.md

paull27:
Add a note to disable hw flow control with the recommended firmware since it is a sw flow control firmware, otherwise Z2M does not start.


Koenkk:
rtscts is false by default, so I propose to rephrase it to not use the rtscts option at all

PR #2828, opened 2024-06-19
Add support for TYZGTH4CH-D1RF

Lacimaci78:
This pull request adds support for the Mumubiz TYZGTH4CH-D1RF device to the Zigbee2MQTT documentation.


DooMMasteR:
ohh well, we did the same within an hour :-)

DooMMasteR:
@Lacimaci78 should you not have already noted, be aware of the fact that it is not advisable to power the device via mains, only USB should be used.  
The power supply uses the classic buck-converter many Tuya modules use, but this one also exposes a mains referenced GND to the temperature and humidity sensor.

Lacimaci78:
> @Lacimaci78 should you not have already noted, be aware of the fact that it is not advisable to power the device via mains, only USB should be used. The power supply uses the classic buck-converter many Tuya modules use, but this one also exposes a mains referenced GND to the temperature and humidity sensor.

yes, but the device will be stored in a distribution box where it will be mounted on a DIN rail and there will be no space for a USB power source

Koenkk:
Already added in https://github.com/Koenkk/zigbee2mqtt.io/pull/2829

Lacimaci78:
The AC-DC non-isolated power supply solution is a streamlined low-power non-isolated switching power supply solution. This solution is developed based on a highly integrated constant-voltage controller chip as the core and a small number of peripheral circuits. This solution supports one or more of the topologies, such as the buck converter, boost converter, and buck-boost converter. The constant-voltage controller chip is embedded with a high-voltage metal–oxide–semiconductor (MOS) transistor. This structure supports multiple functions, such as overcurrent protection, undervoltage protection, and short-circuit protection.

3f7786e124f24f969713c0cb1038560e

My knowledge is that by direct measurement against the short circuit, I do not have the body of the temperature sensor connected to the neutral of the AC system. I have the AC connection directly neutral to neutral and L to L. Even if there was a phase breakdown to the frame of the temperature sensor, I still have a 230V circuit breaker in the system. And a circuit breaker for a normal short circuit.

PR #2841, opened 2024-06-24
Extend the warning about using an unstable connection

Hedda:
Extend the warning about why users should be aware before use an unstable connection to a Zigbee Coordinator serial interface.

The reason for extending and firther generalizing this general warning making it clear that it is not recommend to use a WiFi, WAN and VPN connection between the Zigbee gateway host application and a network-attached Zigbee Coordinator with Serial-Proxy-Server is that some such remote adapters now include a built-in VPN and because of that some people in the Home Assistant community have begun suggesting that it is a good idea that normal users should connect Zigbee2MQTT (or the ZHA integration) running one site directly to a Zigbee Coordinator located on a second site via VPN over WAN instead of setting up another instance of Zigbee2MQTT to run locally on that second site.


Koenkk:
Thanks!

PR #2849, opened 2024-06-26
Mention some general Zigbee imitations in FAQ

Hedda:
Update FAQ to also mention some general limitations that apply to all Zigbee implementations.


docs/guide/faq/README.md
```diff
@@ -6,6 +6,19 @@ sidebarDepth: 0
...
+
+Any Zigbee device is limited to only be paired/joined to one Zigbee Coordinator and only be part of one Zigbee network, meaning that Zigbee devices can only be connected to a single Zigbee gateway. If you want to move a Zigbee device to a different Zigbee network then you need to factory reset that Zigbee device and re-pair/re-join it to the other Zigbee Gateway. This is a limitation in the current (as well as previous) Zigbee protocol specifications, governed by the [CSA (Connectivity Standards Alliance)](https://csa-iot.org/all-solutions/zigbee/), and as such, that limitation applies to all Zigbee implementations, not just the Zigbee2MQTT implementation. This also makes the Zigbee Coordinator a SPOF (Single-Point-Of-Failure) in your Zigbee network, so it needs to be available for your Zigbee devices to stay connected.
```
Koenkk, line 14: > This also makes the Zigbee Coordinator a SPOF (Single-Point-Of-Failure) in your Zigbee network, so it needs to be available for your Zigbee devices to stay connected.

I do not fully agree with this, the Zigbee devices stay connected even when the coordinator is down (a mesh). If the coordinator goes down, you of course cannot control them via the coordinator anymore but devices can still control each other through binding.

docs/guide/faq/README.md
```diff
@@ -6,6 +6,19 @@ sidebarDepth: 0
...
+
+Support for commissioning Zigbee 3.0 devices via "Install Code" or "QR Code" has so far only been implemented for 'zstack' (Texas Instruments ZNP) and 'ember' (Silicon Labs EmberZNet) adapter type radios in Zigbee2MQTT. Other radio adapter types are either missing support in their respective adapter/driver for [zigbee-herdsman](https://github.com/Koenkk/zigbee-herdsman) or more likely missing in the manufacturer's firmware commands/APIs and documentation.
+
+Zigbee2MQTT does not currently support devices that can only use the ZSE ("Zigbee Smart Energy") profile, that is however due to the "Zigbee SE" specification not being part of the standard Zigbee 3.0 specification that includes the more common Zigbee Home Automation + Zigbee lighting and o and thus not implemented in most of the Zigbee Coordinator firmware that is commonly available Zigbee Coordinator radio adapters and modules, usually because the manufacturer offers separate Zigbee protocol stack SDK for Zigbee Smart Energy.
```
marazmarci, line 20:
```suggestion
Zigbee2MQTT does not currently support devices that can only use the ZSE ("Zigbee Smart Energy") profile, that is however due to the "Zigbee SE" specification not being part of the standard Zigbee 3.0 specification that includes the more common Zigbee Home Automation + Zigbee lighting and thus not implemented in most of the Zigbee Coordinator firmware that is commonly available for Zigbee Coordinator radio adapters and modules, usually because the manufacturer offers separate Zigbee protocol stack SDK for Zigbee Smart Energy.
```

docs/guide/faq/README.md
Hedda: @Koenkk OK removed that line now.

Koenkk:
Thanks!

PR #2851, opened 2024-06-26
Update ota_updates.md

Wireheadbe:
Missing info that loglevel needs to be set to "info" or the UI will not show messages about updates.


Koenkk:
thanks

PR #2856, opened 2024-06-28
Fix: Namron dimmer 2.0 4512750 image

geirra:
Seems like the wrong picture is used for Namron Dimmer 2.0 white (4512750).

Source: https://www.elektroimportoren.no/namron-zigbee-led-dimmer-200w-2-0-hvit/4512750/Product.html


Koenkk:
Thanks!

PR #2858, opened 2024-07-01
Add 'device_age' support for STARKVIND air purifier

indrajitr:
Doc update for https://github.com/Koenkk/zigbee-herdsman-converters/pull/7716

Edit: This is probably unnecessary if the doc is auto-generated.


Koenkk:
> Edit: This is probably unnecessary if the doc is auto-generated.

Yes, it will be automatically added with the next release, thanks anyway!

PR #2859, opened 2024-07-01
Add separate model for STARKVIND air purifier table variant

indrajitr:
This is a followup for https://github.com/Koenkk/zigbee-herdsman-converters/pull/7717.


Koenkk:
Thanks!

PR #2876, opened 2024-07-07
Update 15_watchdog.md

mrespin:
I tested it, and it only works when excluding the 'npm start' from the command.


Koenkk:
The `npm start` is a bit confusing indeed, this command should only be used when running bare metal. Thanks!

PR #2880, opened 2024-07-10
Correct image for Namron 4512772

Super1337Boy:
The image for Namron 4512772 is wrong. It shows a different product from the same manufacturer.

I added a new and correct picture.


public/4512772.png
Koenkk, line 1: Please remove this one

public/images/devices/4512772.png
Koenkk, line 1: Please leave some space at the sides of the image and make sure that the background is transparent (use e.g. https://new.express.adobe.com/tools/remove-background)

PR #2882, opened 2024-07-11
Add images for Pushok Hardware devices

zombierzn:


public/images/devices/POK002|POK007.png
Koenkk, line 1: Please name this `POK002_POK007.png`

zombierzn: Is it ok to keep the model as `POK002|POK007` or better to change it too?

Koenkk: Change that too.

PR #2897, opened 2024-07-15
Images for Gledopto GL-MC-002P and Tuya LY-1668

Ru4rc:
Device images for Gledopto GL-MC-002P and Tuya LY-1668


public/images/devices/LY-1668.png
Koenkk, line 1: I don't think this device is supported by z2m?

Ru4rc: Oh true, sorry. Went a bit too fast, still trying to figure out, how to get this device working correctly. Maybe you could take a look at #21876?

Koenkk: Done, will merge the picture already.

PR #2902, opened 2024-07-18
Namron 4512707 Added picture

tore1980:


Koenkk:
Could you make the picture 512x512?

PR #2909, opened 2024-07-23
 fixed TS130F image

aviscode:


Koenkk:
The `TS130F` comes in many forms, if you want a proper image, you can add a [whitelabel](https://github.com/Koenkk/zigbee-herdsman-converters/blob/102ed4723df2c076382dd0627827c760d1ab9b30/src/devices/tuya.ts#L2407) and use the model number for the picture.

aviscode:
not sure i understand 
can u pls explain

Koenkk:
Many different devices are detected as TS130F, so changing the picture isn't the solution. For proper detection of your device you can add it as a whitelabel

aviscode:
but i dont see how can i change the image using whitelabel

Koenkk:
The whitelabel has a `model` property (the second argument of `tuya.whitelabel`), this should be used as the picture name to show it correctly in the z2m frontend.

PR #2919, opened 2024-07-29
List not supported features of adapters

Koenkk:
See title

Regarding Deconz, @ChrisHae are you maybe interested in adding these features? Let me know, then I can help

CC: @Nerivec

### DIFF
--- a/docs/guide/adapters/deconz.md
+++ b/docs/guide/adapters/deconz.md
@@ -1,5 +1,16 @@
 # deCONZ (Dresden Elektronik)
 
+::: warning ATTENTION
+Various features are not supported by this adapter, in case you depend on these features, consider a different adapter.
+
+-   [Changing the channel](../configuration/zigbee-network.md#changing-the-zigbee-channel), changing requires re-pairing all devices.
+-   Adding [install codes](../../guide/usage/mqtt_topics_and_messages.md#zigbee2mqttbridgerequestinstall_codeadd), which is required to pair some devices.
+-   [Backups](../../guide/usage/mqtt_topics_and_messages.md#zigbee2mqttbridgerequestbackup)
+-   Inter-PAN, which is required for [touchlink](../../guide/usage/touchlink.md)
+-   Changing the [transmit power](../../guide/configuration/adapter-settings.md)
+
+:::
+
 ### Configuration
 
 ```yaml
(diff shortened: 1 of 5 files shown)


Nerivec:
How about listing supported settings too? The stuff that is generically listed in [adapter settings](https://www.zigbee2mqtt.io/guide/configuration/adapter-settings.html#configuration-of-the-zigbee-adapter), so users know which settings are actually used by their adapter.
```
disable_led
adapter_concurrent
transmit_power
adapter_delay
```

ember (applicable for current dev): 
```
disable_led: not used
adapter_concurrent: yes, default 16 (if not set or set to zero)
transmit_power: yes, accepted values depend on adapter/chip, -128<->127 range (due to type)
adapter_delay: not used
```

Note1: Seems some of these values need better validation too, number range mostly.
Note2: transmit power is currently [set by Z2M start sequence](https://github.com/Koenkk/zigbee2mqtt/blob/032692673b533a10c9e0cb6af9af4681deee6b05/lib/zigbee.ts#L152), probably should be moved internally to adapters (ember can set it during network forming for example, and it would also allow to set it at a better point in time during the adapter's start).

Koenkk:
> How about listing supported settings too? The stuff that is generically listed in [adapter settings](https://www.zigbee2mqtt.io/guide/configuration/adapter-settings.html#configuration-of-the-zigbee-adapter), so users know which settings are actually used by their adapter.

Done!

> Note1: Seems some of these values need better validation too, number range mostly.

I guess we should do this inside the adapter itself (if it varies per type)

> Note2: transmit power is currently [set by Z2M start sequence](https://github.com/Koenkk/zigbee2mqtt/blob/032692673b533a10c9e0cb6af9af4681deee6b05/lib/zigbee.ts#L152), probably should be moved internally to adapters (ember can set it during network forming for example, and it would also allow to set it at a better point in time during the adapter's start).

Then the benefit would be that the transmit power is also set during the start?

Nerivec:
I'm thinking at least some validation of "no unreal value" straight in the settings.json wouldn't hurt (prevent z2m from starting). No concurrency 0, or 30000, no tx power at 50000, that kind of thing...

> Then the benefit would be that the transmit power is also set during the start?

Yeah, instead of setting it when devices are likely rejoining/reconnecting (which looks to be somewhat the case right now), we could set it much earlier (depending on the adapter), and for stacks that support it, right away on network forming too.

PR #2920, opened 2024-07-30
add supprot for TS0049 flow water timer

MScorpioLee:
add supprot for TS0049 flow water timer


Koenkk:
Could you also make a PR to add support for this device to https://github.com/Koenkk/zigbee-herdsman-converters?

PR #2923, opened 2024-08-01
RPi Zero Update guide

templehasfallen:
Created a guide detailing a workaround for updating Zigbee2MQTT out-of-place when using a RPi Zero, 2, W using chroot and qemu.
This is in the likely case of users running into OOM issues on their Pi.

### DIFF
--- a/docs/advanced/more/pi-zero-update.md
+++ b/docs/advanced/more/pi-zero-update.md
@@ -49,8 +49,8 @@ umount /media/ubuntu/bootfs
 Mount the microsd in your mountpoint:
 ```bash
 # Use drive letter from before
-sudo mount -o rw /dev/sde1 /mnt
-sudo mount -o rw /dev/sde2 /mnt/boot
+sudo mount -o rw /dev/sdc1 /mnt
+sudo mount -o rw /dev/sdc2 /mnt/boot
 ```
 
 Mount the binds:
@@ -85,6 +85,11 @@ chroot /mnt /bin/bash
 ```
 
 ### Updating
+
+Login as your user:
+```bash
+su USER
+```
 
 Now you should be able to follow the same update steps as normally.
 
--- a/docs/advanced/more/pi-zero-update.md
+++ b/docs/advanced/more/pi-zero-update.md
@@ -1,0 +1,114 @@
+---
+sidebar: auto
+---
+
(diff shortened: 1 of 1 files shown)


Koenkk:
I'm wondering if `pnpm` will fix this, can you try:

```
npm install -g pnpm
pnpm install
```

templehasfallen:
> I'm wondering if `pnpm` will fix this, can you try:
> 
> ```
> npm install -g pnpm
> pnpm install
> ```

Sadly this also errors out due to OOM related errors. 

```
 WARN  GET https://registry.npmjs.org/is-stream error (EAI_AGAIN). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/async error (EAI_AGAIN). Will retry in 10 seconds. 2 retries left.
Progress: resolved 520, reused 0, downloaded 484, added 0

<--- Last few GCs --->

[2901:0x47058a0]   198301 ms: Mark-Compact (reduce) 124.2 (129.2) -> 122.7 (128.6) MB, 317.22 / 0.00 ms  (average mu = 0.222, current mu = 0.162) allocation failure; scavenge might not succeed
[2901:0x47058a0]   199194 ms: Mark-Compact (reduce) 123.6 (128.9) -> 123.2 (128.4) MB, 745.23 / 0.17 ms  (average mu = 0.189, current mu = 0.166) allocation failure; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

Aborted
```

I also tried setting `export NODE_OPTIONS="--max-old-space-size=256"` or 512, in which case it completed, but the service just kept starting/restarting without any specific errors.

Koenkk:
What if you configure the Pi to have a swap file? https://forums.raspberrypi.com/viewtopic.php?t=369538

templehasfallen:
> What if you configure the Pi to have a swap file? https://forums.raspberrypi.com/viewtopic.php?t=369538

I went to do that but apparently already did that a while ago (I have 2G swap). This may also explain why my `npm ci` didn't directly fail and was still running after ~12 hours without any change.

Koenkk:
I wonder if it would be easier to grab to dependencies from the Docker image. Since this is quite complicated and needs to be done after every update of z2m.

PR #2924, opened 2024-08-01
Update home_assistant.md

github-vdabbeel:
Add condition to only loop over devices containing the requested attribute.
This avoids the error "No first item, sequence was empty."


Koenkk:
Thanks!

PR #2936, opened 2024-08-08
Update QS-Zigbee-C01.md

SalsaGael:
Calibration Time change solution


rdorys:
Hi @SalsaGael, do you think there is no mean of doing it other than via dev console ? For some reasons,I would like to change it via HA with a MQTT message.

SalsaGael:
> Hi @SalsaGael, do you think there is no mean of doing it other than via dev console ? For some reasons,I would like to change it via HA with a MQTT message.

Yes it will, @Koenkk add it, it's on HA with Z2M release 1.40.0

mremedi2023:
Hi, apply for QS-Zigbee-C03?

mremedi2023:
Hello, does it work for QS-Zigbee-C03?

PR #2939, opened 2024-08-09
Update TS0601_dimmer_4.md

electronicm:
Whitelabel MS-105B-M removed because it is not a Zigbee device. Ref: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7835#discussion_r1709140620


Koenkk:
Thanks!

PR #2943, opened 2024-08-11
Replaced incorrect image

MyungjinSong:
The previous image in the TS0601_dimmer_5.md file was incorrect. It depicted the "2-gang" version, but the correct image should show the "1-gang" version. I have replaced it with the correct image to ensure the documentation accurately reflects the intended product.


Koenkk:
Please update the image itself instead of changing the link

PR #2953, opened 2024-08-17
Remove warning that input 2 is not supported for Philips wall switch

juanpescador:
Is the warning still relevant? Actions are reported for both inputs out of the box for me with Z2M 1.39.1 and firmware version 1.0.3 (build date 20210115). I did click on "dual_rocker" before recording the GIF, not sure if that had an impact (when going back to the "Exposes" tab in the recording, it no longer displays which device mode is enabled).

2022-08-17 Philips Hue wall switch both inputs work

Also I've edited the GIF, pausing on the frames where the action is visible. In real life the transitions are much faster.

### DIFF
--- a/docs/devices/929003017102.md
+++ b/docs/devices/929003017102.md
@@ -44,7 +44,6 @@ Notes:
 - This device does not support direct binding (to a device instead of a group).
 - After doing this, the power events are still sent and can hence be used for automation.
 - Pressing the yellow reconfigure button in the *About* tab will reset the bindings of the coordinator, but it will not unbind any groups - they have to be unbound explicitly.
-- Actions are only sent for input 1 of the device. For input 2 no actions will be sent (so a double rocker is **not** supported).
 <!-- Notes END: Do not edit below this line -->
 
 


juanpescador:
Looking through the edit history of this file, it looks like the disclaimer applies (or applied) to group binding. Should I edit the disclaimer rather than remove it then?

For reference, the disclaimer was added in https://github.com/Koenkk/zigbee2mqtt.io/commit/7f05e31574f293638c19b67945fa399dbe333dd8, which referred to zigbee2mqtt [issue 10740](https://github.com/Koenkk/zigbee2mqtt/issues/10740):

>> As written earlier in this thread, it is possible to make this device send toggle commands to Zigbee groups. To enable this some fixes where required. These are available in the latest-dev branch now. I documented how to accomplish the group binding: https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/devices/929003017102.md#directly-control-zigbee-groups
>
>To clarify.. this only works for one of the input/side of switch, right?

_Originally posted by @martijn-brant in https://github.com/Koenkk/zigbee2mqtt/issues/10740#issuecomment-1629753783_

martijn-brant:
This device always had two working inputs. The warning was for when using direct binding to groups (allowing the wall switch to control the lights even when Z2m / controller software is down). So please don't remove this warning as it applies to the binding scenario, not the normal usage scenario.

juanpescador:
> The warning was for when using direct binding to groups

Indeed, that's what I found looking through the file history. After playing around with it and trying to bind it to a group things came together. For first-time readers I think it's worth rewording; it seemed like the limitation applied in all cases. I'll open another PR with a proposal, see what everyone thinks.

markuswebert:
@martijn-brant Sorry for writing on a closed issue.. I wondered if there are any information about the direct binding issue with the second input available? What is the restricting part? The wall switch firmware? Read about this restriction multiple times but I was not able find more details. Thanks!

PR #2956, opened 2024-08-18
Update A1Z.md reintroduce Reset energy chapter

krusta4711:
Reset energy chapter was removed with commit "c9e8b4810dbf3458a48946739049a8082328292a". But resetting the A1Z plug still works. So this is valuable input for users.


Koenkk:
Thanks!

PR #2967, opened 2024-08-24
Show how to pair PTM_215ZE to a specific channel

kalsan:
The user manual for this device was updated and thus, more detailled information is now available. This change clarifies the pairing of PTM_215E with regard to the enhanced manual.


Koenkk:
Thanks!

PR #2976, opened 2024-08-28
Update 01_zigbee_network.md

chris-1243:
Add new device which support ZGP translation.

By the way, all my tests are done with PTM215Z/ZE devices (greenPower_2).


Koenkk:
Thanks!

PR #2992, opened 2024-09-02
Added Profalux MAI-ZTS cover remote

Mecrean79:
MAI-ZTS cover remote seems to be a clone of Profalux MAI-ZTM20C


Koenkk:
Thanks!

PR #2993, opened 2024-09-02
Update home_assistant.md 

maniac-semal:
Additional condition to avoid having "Non-Zigbee2MQTT" devices listed.


Koenkk:
Thanks!

PR #2999, opened 2024-09-04
support Hue Datura

coendaar:
Add image
Add note that it shows up as 2 devices.


coendaar:
resolves Koenkk/zigbee2mqtt#23831

PR #3000, opened 2024-09-04
Add pair info to the notes of TRÅDFRI LED drivers

adriaandotcom:
I tried the sequence for so many times. This note would have saved my quite some time.


Koenkk:
Thanks!

PR #3012, opened 2024-09-10
Add images for Perifo light tubes

LaurentvdBos:
See https://github.com/Koenkk/zigbee-herdsman-converters/pull/7968


LaurentvdBos:
Small sidenote: z2m added `identify()` to the list of extends, but since that was missing for many of the Philips bulbs, I did not add it (it also adds it for Hue bulbs that I already have, which do not have such a block in converters).

I rarely use the identify button, so I won't miss it, but is this something to add to the bulbs that support it?

PR #3017, opened 2024-09-12
Document unit of external temperature sensor

magicus:
Added documentation of the value of the external sensor.


docs/devices/014G2461.md
```diff
@@ -146,7 +146,7 @@ The temperature sensor of the TRV is — due to its design — relatively close
 Value can be found in the published state on the `external_measured_room_sensor` property.
 To read (`/get`) the value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/get` with payload `{"external_measured_room_sensor": ""}`.
 To write (`/set`) a value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/set` with payload `{"external_measured_room_sensor": NEW_VALUE}`.
-The minimal value is `-8000` and the maximum value is `3500`.
+The minimal value is `-8000` and the maximum value is `3500`. The unit of this value is 0.01 `°C` (so e.g. 21 °C would be represented as 2100).
```
Koenkk, line 149: This part of the doc is generated, update https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/danfoss.ts#L122 instead

I'm wondering if it makes sense to divide the exposed value by 100 such that they are in celsius?

magicus: It would have made sense if it were done initially, but now there is a lot of Home Assistant blueprints and rules out there that assume they need to multiply their sensors by 100 before feeding it in here.

PR #3027, opened 2024-09-17
Add hint to required firmware for auto close switch

N3rdix:
It was not mentioned in the PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/7891 and already caused some confusion on different channels.
This hint should help...


Koenkk:
This part of the doc is generated, please update https://github.com/Koenkk/zigbee-herdsman-converters/blob/08073e93322b2ee04c802bfe111030a6febf3a89/src/devices/sonoff.ts#L1155

PR #3042, opened 2024-09-25
Update versions for `ember`.

Nerivec:
_Also removed recommended firmware links. Several manufacturers have their own system for that, so don't duplicate information that's likely to get stale and/or mismatch._

### DIFF
--- a/docs/guide/adapters/emberznet.md
+++ b/docs/guide/adapters/emberznet.md
@@ -35,6 +35,11 @@ The use of `adapter: ezsp` is now deprecated. See [https://github.com/Koenkk/zig
 
 ### Firmware flashing
 
+-   Firmware repositories:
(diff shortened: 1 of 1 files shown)


Koenkk:
For each adapter, can we point to the location on where to find the firmware? It may be confusing for users now where to find the correct firmware.

Nerivec:
darkxst repo has yet to migrate to the new builder, so, I included the link to the base repo, not to "releases", will have to change that as soon as it's updated (more "direct" link).

I also keep a curated list here:
https://github.com/Nerivec/ember-zli/blob/main/src/utils/firmware-links.ts
It's used internally by ember-zli to avoid users having to provide a file/link.

Koenkk:
Thanks!

PR #3044, opened 2024-09-26
added img for two new devices EFEKTA PST DUO V1 LR and EFEKTA PST DUO V1

smartboxchannel:
added img for two new devices EFEKTA PST DUO V1 LR and EFEKTA PST DUO V1


Koenkk:
Thanks!

PR #3054, opened 2024-09-30
delete On with timed off with ZBMINIL2

liangjia2019:
Confirm that the attribute and implementation are not included in the internal code. Please remove them to avoid misunderstandings
image


Koenkk:
Thanks!

PR #3057, opened 2024-10-01
throttle option description

ivanfmartinez:
document the changes on https://github.com/Koenkk/zigbee2mqtt/pull/24122

### DIFF
--- a/docs/guide/configuration/devices-groups.md
+++ b/docs/guide/configuration/devices-groups.md
@@ -105,7 +105,7 @@ Throttle processing of messages from this device. When setting e.g. `throttle: 1
 
 Setting this option reduces the number of MQTT messages sent for a particular device. This is directly linked to how high the option is set. When used on misbehaving devices with proper values, it can drastically reduce the size of external databases that store history (like Home Assistant), and yet have little to no impact on the quality of said history.
 
-Some ambient sensors like `TS0601_air_quality_sensor` and some water level sensors at least are know to get improved behaviour with this setting.
+Some ambient sensors like `TS0601_air_quality_sensor` and some water level sensors, among others, are known to benefit from this option
 
 **`retrieve_state`**  
 (DEPRECATED) Retrieves the state after setting it. Should only be enabled when the reporting feature
--- a/docs/guide/configuration/devices-groups.md
+++ b/docs/guide/configuration/devices-groups.md
@@ -99,9 +99,9 @@ hand `debounce: 1` with `debounce_ignore: - action` will publish all unique acti
 e.g. `action: rotate_left` and `action: rotate_stop`)
 
 **`throttle`**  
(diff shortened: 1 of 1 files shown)


docs/guide/configuration/devices-groups.md
```diff
@@ -98,6 +98,15 @@ and `debounce` option without `debounce_ignore` publishes only last payload with
 hand `debounce: 1` with `debounce_ignore: - action` will publish all unique action messages, at least two (
 e.g. `action: rotate_left` and `action: rotate_stop`)
 
+**`throttle`**  
+Throttle messages of this device. When setting e.g. `throttle: 10` the first message from device is accepted but all other messages in next 10 seconds will be ignored. Be carefull when using this option, unlike debounce that can ignore some attributes this will drop entire message.
```
Nerivec, line 102:
```markdown
Throttle processing of messages from this device. When setting e.g. `throttle: 10` the first message from the device is processed but all other messages within the next 10 seconds are ignored. Be careful when using this option, unlike `debounce` that can ignore some attributes, this will drop the entire message.
```

docs/guide/configuration/devices-groups.md
```diff
@@ -98,6 +98,15 @@ and `debounce` option without `debounce_ignore` publishes only last payload with
...
+
+For now this option will only work when debounce is disabled.
```
Nerivec, line 104:
```markdown
`debounce` option has priority over `throttle`; if both are present for one device, only `debounce` will have any effect.
```

docs/guide/configuration/devices-groups.md
```diff
@@ -98,6 +98,15 @@ and `debounce` option without `debounce_ignore` publishes only last payload with
...
+
+This is handy for some sensor devices that keep sending messages each second, e.g. 'TS0601_air_quality_sensor', even without considerable changes of values, like some water level sensors and ambient sensors.
+
+This option will reduce the MQTT messages sent for this device and also reduce storage if the states are stored in homeassistant.
```
Nerivec, line 108:
```markdown
Setting this option reduces the number of MQTT messages sent for a particular device. This is directly linked to how high the option is set. When used on misbehaving devices with proper values, it can drastically reduce the size of external databases that store history (like Home Assistant), and yet have little to no impact on the quality of said history.
```

ivanfmartinez: @Nerivec here you think that I should replace both paragraphs with your suggestion or only the second one ?

Nerivec: I'd say both, you can put some example in parenthesis
```markdown
misbehaving devices (some water level, presence, etc. sensors, e.g. `TS0601_air_quality_sensor`)
```

ivanfmartinez: @Nerivec for me reading the example in separate became easier to understand. I'm not an english native. 
can you check my last commit ?

Koenkk:
Thanks!

PR #3060, opened 2024-10-02
Müller Licht Malea Image + Pairing info

itkama:
https://github.com/Koenkk/zigbee2mqtt/issues/24183


Koenkk:
Could you also make a PR for the device by clicking [here](https://github.com/Koenkk/zigbee-herdsman-converters/edit/master/src/devices/muller_licht.ts)?

itkama:
Will do that tomorrow. I noticed that they use a slightly different converter though that has some mullerLicht specific stuff (that does not seem to work on my device. Also want to add a note about pairing here, but that somehow seems to have gotten los.

PR #3067, opened 2024-10-04
Update E2206.md

Floo33R:
Added reference for updating the firmware manually.


Koenkk:
I've added the ota to zigbeeOTA now, so this should not be needed anymore. Thanks anyway!

PR #3070, opened 2024-10-06
add TS0726_switch_4g_2s image

imammedo:
src: https://www.aliexpress.com/item/1005004888723698.html 4-2Key Zigbee Switch


Koenkk:
Thanks!

PR #3072, opened 2024-10-06
Add info about power meter sum reset

mp3-10:
Added info about resetting power consumption sum (from TS011F_plug_1)


Koenkk:
Can you confirm that sending `zigbee2mqtt/FRIENDLY_NAME/set/factory_reset` with an empty payload also works?

mp3-10:
Checked and:
`z2m: No converter available for 'factory_reset' ({})`

Koenkk:
Whoops, had to be `reset`, not `factory_reset`

mp3-10:
@Koenkk i tested: `zigbee2mqtt/FRIENDLY_NAME/set/reset` works too

Koenkk:
Great, I'll add this later via expose. for now this is ok, thanks!

PR #3073, opened 2024-10-06
Add Hue Twilight images

mundschenk-at:
Adds images for the white and black version of Hue Twilight, `LGT001` and `LGT002`. (However, in reality both images depict LGT001, as I don't own the black one.)


Koenkk:
Thanks!

PR #3085, opened 2024-10-10
Add Philips Hue LCZ002 image

tholland15:


Koenkk:
shouldn't be needed as it will be added to 8719514419278

PR #3091, opened 2024-10-11
Added HK-SL-DIM-AU-R-A

russr999:
Added HK-SL-DIM-AU-R-A


Koenkk:
Could you make the picture 512x512?

russr999:
> Could you make the picture 512x512?

Imaged updated to be 512 x 512

Koenkk:
Thanks!

PR #3103, opened 2024-10-14
Add picture for TOMZN TOB9Z-VAP

yannpub:


Koenkk:
Thanks! Please ignore my earlier comment about the picture PR

yannpub:
Hi Koen,

I hope I won’t create some mess with the doc update.

My assumptions are that :
1) update in the converter from -M to -VAP will regenerate the device page with TOB9Z-VAP name. (PR #8134)
2) It will then pick the proper picture that I have put in the PR, named TOB9Z-VAP.png (PR #3103)
3) It will create a new page for the TOB9Z-63M, as I added it in the white label for the Tongu TO-Q-SY1-JZT. (PR #8134)

4) It will not pick-up the 'old and wrong' image of the TOB9Z-VAP, which was called TOB9Z-M.png --> Can this file be renamed TOB9Z-63M.png or should I create a new upload?  I have the file already ready, and a bit sharper than the old one, so which whichever is the more convenient for you.

Best regards,
Yann

Koenkk:
Thats correct, for the picture please create a new one (leave the old one as is)

PR #3104, opened 2024-10-14
Remove configuring group members through `configuration.yaml`

Koenkk:
See https://github.com/Koenkk/zigbee2mqtt/pull/24338

### DIFF
--- a/docs/guide/configuration/devices-groups.md
+++ b/docs/guide/configuration/devices-groups.md
@@ -4,7 +4,7 @@ sidebarDepth: 1
 
 # Devices and Groups
 
-Zigbee2MQTT also stores the definitions of Devices and Groups in the `configuration.yml`.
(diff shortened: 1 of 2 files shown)


mremedi2023:
Hi, I have some devices that flood the data bus, so the documentation suggests adding a line to the configuration.yaml. Ex:

Stop message flooding
This unit has a bug that makes it send multiple messages when updating. To stop this from flooding your MQTT Queues, please add the following to your configuration.yaml file:

devices:
'0x12345678':
friendly_name: thermostat
debounce: 1

It doesn't matter if I add the line to the file, when I restart the added line is no longer there, how can I fix this? the studio code server saves the changes but then they are not there. Can this be fixed?

Koenkk:
I don't think that has anything to do with this PR, for support use the [Forum](https://github.com/Koenkk/zigbee2mqtt/discussions) instead

PR #3105, opened 2024-10-15
Candeo images added and updated

dhc25:
Add images for Candeo LED strip controllers.
Added image for Candeo C202.1
Updated image for Candeo 202


Koenkk:
Thanks!

PR #3111, opened 2024-10-18
Update ZS-EUB_2gang.md

Shariner:
Added instructions to get ZS-EUB_2gang fully working in Z2M. There are some issues with the 2gang wall switch.

Without this fix, the configuration parameters for backlight mode, poweronbehaviour are not supported/working
Without this fix, when one of the switches is executed via Homeassistant or Z2m, both switches are turned on/off.

### DIFF
--- a/docs/devices/ZS-EUB_2gang.md
+++ b/docs/devices/ZS-EUB_2gang.md
@@ -23,8 +23,81 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
+# Issues with Moes ZS-EUB_2gang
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Could you make a PR for the device definition instead by clicking [here](https://github.com/Koenkk/zigbee-herdsman-converters/edit/master/src/devices/tuya.ts)?

Shariner:
> Could you make a PR for the device definition instead by clicking [here](https://github.com/Koenkk/zigbee-herdsman-converters/edit/master/src/devices/tuya.ts)?

Sure, looks way more better to improve the converter itself than writing a hint on the Z2M page :D

I will examine the converter files in the next day. I cancel this PR.

PR #3119, opened 2024-10-20
Add link to ESP32 adapter firmware

andryblack:
I write [basic implementation](https://github.com/andryblack/esp-coordinator) for ZBOSS NCP protocol for esp32 with coordinator role supported almost all command used by zigbee2mqtt project.


Koenkk:
thanks!

PR #3130, opened 2024-10-23
Improve documentation

roelzadh:
Couldn't get the dimmer reset to work unless I did a power cycle. For my bulbs this was not needed.


Koenkk:
Updated it a bit, thanks!

PR #3142, opened 2024-10-26
Phillips Hue Centris (4 spots white) - Added 929003808901.png

frest-de:
Phillips Hue Centris (4 spots white) - Added 929003808901.png


Koenkk:
Thanks!

PR #3149, opened 2024-10-27
Add picture for Philips Hue Go White and Color 1390780

oajns:
Add picture for Philips Hue Go White and Color 1390780


Koenkk:
Thanks!

PR #3164, opened 2024-10-30
Add Hue Datura Big

markuswebert:
Extend existing Datura description to include size information. Add description for new Datura Big. Copy over picture from Datura Small to Big.


Koenkk:
Thanks!

PR #3177, opened 2024-11-01
Hue Tento color 29,1cm and 54,2cm

hetisstephan:
Added pictures for Hue Tento color 29,1cm and 54,2cm
Model 929003823901: 54,2cm
Model 929003823501: 29,1cm


Koenkk:
Thanks!

PR #3181, opened 2024-11-02
Add device image for Hue Cher Pendant 929003054201

Basanites:
The image is just a duplicate of the already existing image for  4076131P6. 
As such it is not 500x500 but rather 150x150. 
Let me know if I should upload a 500x500 image instead. I did it this way to not have lots of different images for the "same" product.


Koenkk:
512x512 would be better indeed, could you also update the one of 4076131P6 ?

Koenkk:
jpg can stay, thanks!

PR #3215, opened 2024-11-13
Update Rodret documentation

LaurentvdBos:
Contrary to what the documentation states, I was able to bind my Rodret to a Zigbee group without problems. I also removed the superfluous sentence how to bind (if you add it back, please fix the `unbind` -> `bind` typo in the mqtt topic).

I suspect it is a firmware issue that hindered @lorenzspenger from binding it to groups in the past? I did not have to do a firmware update though, mine came with 1.0.47 out-of-the-box.


Koenkk:
Super! 😄

PR #3235, opened 2024-11-20
Update IH012-RT01.md

srett:
Add alternate method for how to fix device not updating occupancy anymore after battery swap.


Koenkk:
thanks!

PR #3241, opened 2024-11-22
Update BTH-RA.md

ondrejbaranek:
Added information how to create install code when the QR code is for Matter only, and the Zigbee Install code needs to be created manually.


Koenkk:
Thanks!

PR #3258, opened 2024-11-25
Added image for E2223 variant of IKEA INSPELNING smart plug

matban666:
This appears to be the UK version of the plug.


Koenkk:
Thanks!

PR #3259, opened 2024-11-25
Improving Gledopto Pro Controller images

pjxturner:
Better quality 512x512 png images for Gledopto **Pro** LED controllers - GL-C-003P (RGB), GL-C-006P (CCT), GL-C-007P (RGBW), GL-C-008P (RGBCCT) and GL-C-009P (Dimmer).
**Note**: A single-mode LED controller was never released by Gledopto for the GL-C-003P RGB Pro LED controller.  However, it is available as an option on their Multi-mode controllers, including GL-C-001P, GL-C-002P,GL-C-011P, GL-C-201P, GL-C-202P, GL-C-301P, which is why it is a valid Z2M device.  This image is just to represent what it would've looked like, consistent with the other single-mode Gledopto Pro controllers, rather than having a blank image.  The current image for a different model, so this is a better placeholder.


Koenkk:
Thanks!

PR #3260, opened 2024-11-25
Update Busch-Jaeger 6735/6736/6737 docs for Z2M 2.0

klada:
* Don't mention legacy payload anymore
* Remove note regarding switch entity (dropped in Z2M 2.0)


Koenkk:
Thanks!

PR #3267, opened 2024-11-28
Add read/write attribute format

tim-vk:
So I was running into some trouble when wanting to test custom attributes. 
The frontend does allow reading/writing some of them, however i wanted to test some non-standard ones.  
It was quite a challenge to find out how the API works & what is does (& doesn't) allow. 

If you search for "how to set custom attributes" or "zigbee2mqtt dev console payload" shows i'm not the only one.

Ideally I'd also like to see a change in the UI, but this is step 1.

### DIFF
--- a/docs/guide/usage/mqtt_topics_and_messages.md
+++ b/docs/guide/usage/mqtt_topics_and_messages.md
@@ -85,7 +85,7 @@ Publishing messages depends on the MQTT client you use. For example to publish a
 #### Directly reading/writing ZCL attributes
 For development/debug purposes it's possible to directly send read/write attribute requests to a device through this endpoint.
 Clusters and attributes can be referred to either by name or ID. 
-For all available names within zigbee2mqtt please refer to [cluster.ts](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/definition/cluster.ts).
+For all available names within Zigbee2MQTT please refer to [cluster.ts](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/definition/cluster.ts).
 
 ##### Read requests:
 
--- a/docs/guide/usage/mqtt_topics_and_messages.md
+++ b/docs/guide/usage/mqtt_topics_and_messages.md
@@ -82,6 +82,44 @@ Publishing messages depends on the MQTT client you use. For example to publish a
  mosquitto_pub -t 'zigbee2mqtt/0x0fffffffffffffff/set' -m '{ "state": "ON" }'
 ```
 
+#### Directly reading/writing ZCL attributes
+For development/debug purposes it's possible to directly send read/write attribute requests to a device through this endpoint.
+Clusters and attributes can be referred to either by name or ID. 
+For all available names within zigbee2mqtt please refer to [cluster.ts](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/definition/cluster.ts).
+
+##### Read requests:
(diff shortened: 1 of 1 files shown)


docs/guide/usage/mqtt_topics_and_messages.md
```diff
@@ -82,6 +82,48 @@ Publishing messages depends on the MQTT client you use. For example to publish a
  mosquitto_pub -t 'zigbee2mqtt/0x0fffffffffffffff/set' -m '{ "state": "ON" }'
 ```
 
+### directly reading/writing attributes
```
Koenkk, line 85:
```suggestion
#### Directly reading/writing ZCL attributes
```

docs/guide/usage/mqtt_topics_and_messages.md
```diff
@@ -82,6 +82,48 @@ Publishing messages depends on the MQTT client you use. For example to publish a
...
+For development/debug purposes it's possitlbe to directly send read/write attribute requests to a device through this endpoint.
```
Koenkk, line 86:
```suggestion
For development/debug purposes it's possible to directly send read/write attribute requests to a device through this endpoint.
```

docs/guide/usage/mqtt_topics_and_messages.md
```diff
@@ -82,6 +82,42 @@ Publishing messages depends on the MQTT client you use. For example to publish a
...
+```js
+{
+  "read": {
+    "cluster": genBasic, // Either name (if defined in standard) or ID of cluster can be used
+    "attributes": [1]    // Either name (if defined in standard) or ID of attribute can be used
+  }
+}
+```
+##### Write requests:
+
```
Koenkk, line 99: I think it would be good to link to the clusters; https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/definition/cluster.ts

docs/guide/usage/mqtt_topics_and_messages.md
tim-vk: So I'm not entirely sure about the terminology but isn't the ZCL only the standard attributes that every device has? (as in, this is "the" spec).

Are any other attributes that manufacturers added themselves also still "ZCL attributes"?

docs/guide/usage/mqtt_topics_and_messages.md
tim-vk: added!

docs/guide/usage/mqtt_topics_and_messages.md
```diff
@@ -82,6 +82,44 @@ Publishing messages depends on the MQTT client you use. For example to publish a
  mosquitto_pub -t 'zigbee2mqtt/0x0fffffffffffffff/set' -m '{ "state": "ON" }'
 ```
 
+#### Directly reading/writing ZCL attributes
+For development/debug purposes it's possible to directly send read/write attribute requests to a device through this endpoint.
+Clusters and attributes can be referred to either by name or ID. 
+For all available names within zigbee2mqtt please refer to [cluster.ts](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/definition/cluster.ts).
```
Koenkk, line 88:
```suggestion
For all available names within Zigbee2MQTT please refer to [cluster.ts](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/definition/cluster.ts).
```

Koenkk:
Thanks!

PR #3277, opened 2024-12-01
Add sniffer instructions for nRF52 adapter

jcblomquist:
Nordic nRF52 dongle is a $10 adapter that can be used for sniffing Zigbee traffic. Nordic maintains their own firmware + docs to set the sniffer up on any OS, so I though it'd be helpful to reference it here in the z2m sniffer guide.


Koenkk:
thanks!

PR #3287, opened 2024-12-03
Added notes on how to pair the Aqara Water Leak sensor.

kkevindev:
Added some missing notes about how to pair the device.


Koenkk:
Thanks!

PR #3291, opened 2024-12-03
Fix Mazda MZVT1Z model number

Nopraz:
I confused the model number and the reference. TR-M2Z is the model number


Koenkk:
Thanks!

PR #3292, opened 2024-12-05
Update README.md

ThePatricide:
If you change from zStack 1.4.0 to 3.0, as is my case, you need to re-pair your routers most likely. It is mentioned sometimes in the past, but it is not in the FAQ. For me, as a CC2531 user that finally decided to upgrade, I ran into the issue that none of my routers want to connect, only the end devices. Therefore, this breaks my network and I need to re-pair at least my routers, thereby hopefully automatically restoring access to all end nodes as well.

### DIFF
--- a/docs/guide/faq/README.md
+++ b/docs/guide/faq/README.md
@@ -85,9 +85,10 @@ You need to re-pair all you devices when:
 
 - Changing the network key (`network_key`) or panID (`pan_id`) in `configuration.yaml`.
 - Changing the Zigbee channel (`channel`) in the `configuration.yaml` might require repairing **some** devices, read the [documentation for more info](../configuration/zigbee-network.md#changing-the-zigbee-channel).
+- Changing from zStack 1.x to 3.x.
 - Switching between adapters requires repairing, **except when**:
     - When adapters are the same repairing is **not** required (e.g. CC2531 -> CC2531)
-    - Switching from a CC2530 or CC2531 based adapter to a CC2538/CC2652/CC1352 based adapter does **not** require repairing (e.g. CC2531 -> CC2652)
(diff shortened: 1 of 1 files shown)


docs/guide/faq/README.md
```diff
@@ -85,9 +85,10 @@ You need to re-pair all you devices when:
 
 - Changing the network key (`network_key`) or panID (`pan_id`) in `configuration.yaml`.
 - Changing the Zigbee channel (`channel`) in the `configuration.yaml` might require repairing **some** devices, read the [documentation for more info](../configuration/zigbee-network.md#changing-the-zigbee-channel).
+- Changing from zStack 1.x to 3.x.
```
Koenkk, line 88: I think this only happens with the CC2531 (legacy adapter), it simply doesn't have enough memory available to keep all the childs (3.x is NOT recommended for the CC2531)

ThePatricide: That is correct. That is exactly what I'd like to have added to the README if possible, to make it clear for people like me. In addition, is it correct only routing devices need to be re-paired, as end nodes are less strict and will keep connected? I should have added that too perhaps (I didn't do the migration yet but am planning to, might add that information from experience by then).

Koenkk: Then I would make it a bit more clear by

> - Changing from zStack 1.x to 3.x. (only applies to CC2530 and CC2531)

What happens is that zstack 3 has smaller routing tables (due to the stack being larger), I guess this is what causes the disconnects.

ThePatricide: Yeah that would be perfect. Thanks for your review and comments!

Koenkk: Updated it, thanks!

PR #3295, opened 2024-12-05
Update home_assistant.md

DubhAd:
The problem is that the event trigger as written won't work if the button is pushed a second time. This is because the attribute won't change, so HA won't trigger.

By using the fact that events have a timestamp of the last event, you can trigger when the state changes, and then check the event type attribute.

[This HA forum thread](https://community.home-assistant.io/t/wth-are-event-entities-not-triggered-a-second-time-when-using-the-attribute-event-type-in-a-state-trigger/804428) explains this problem in more detail.


Koenkk:
Thanks!

PR #3296, opened 2024-12-05
Added support TRV603-WZ

Cosik:
Hi,

I'm not sure it's all here what I should add. PR for converter is on the way.


Koenkk:
Thanks!

PR #3297, opened 2024-12-05
Add pairing notes for TS0601_cover_1

silberjan:
Added notes for pairing of bead curtain motors which are detected as TS0601_cover_1 
Pairing on these works different than the already documented methods


Koenkk:
Thanks!

PR #3301, opened 2024-12-06
[HG06104A] Correct description

RubenKelevra:
- It's 2 meters, not 2.5 meters
- The branding is "Livarno Home", not "Livarno Lux"
- Added that it's an "RGBCCT" LED strip

Packaging:

IMG_20241206_170157


Koenkk:
This part of the page is generation, please update the `description` in zigbee-herdsman-converters instead.

PR #3310, opened 2024-12-08
Add image for silabs router

Nerivec:
For https://github.com/Koenkk/zigbee-herdsman-converters/pull/8456
Best to keep it generic, since the converter is generic for several brands/adapters I figure...
Unless we can tag the filename by fingerprint instead?


Koenkk:
> Unless we can tag the filename by fingerprint instead?

Yes, this is possible by adding the whiteLabels (although not sure if you want to do this)

Nerivec:
I think it's fine like this. No reason to bloat with more images. The silabs logo even makes it easier to quickly identify the routers from other devices.

PR #3319, opened 2024-12-12
Note that Enbrighten devices sometimes need to be power cycled

deviantintegral:
I had two different Enbrighten devices stop responding over the past two days. In the past, I've just power cycled them but not figured out why they were not working.

Today, I sniffed the traffic, and discovered that the coordinator had no route to the device such as:

```
debug: 	zh:zstack: Discovering route to 7418
```

While I could see `Route Request` commands being sent, there were no `Route Record` responses coming from the devices. This was even though I could see occasional link status broadcasts coming from those devices.

As soon as I power cycled the devices, I saw them sending `Route Record` responses and all was good again.

I don't see any bug in herdsman for this and think its' an issue with the Enbrighten devices. I've seen this behaviour across a few different models, so I wouldn't be surprised if they all had this issue. Enbrighten has never issued a firmware update for Zigbee devices, so I think we should put this warning in the docs. I know between this and https://github.com/jascoproducts/firmware/issues/127, I wish I had used a different vendor!

### DIFF
--- a/docs/devices/43076.md
+++ b/docs/devices/43076.md
@@ -23,7 +23,11 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
+## Notes
 
+# Device Not Responding
+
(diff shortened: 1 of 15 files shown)


Koenkk:
Thanks!

PR #3321, opened 2024-12-12
Fix some formatting in IKEA E1525_E1745.md

vbfox:
* The occupancy_timeout text was in the middle of the previous section
* Use links that can render on the website


Koenkk:
Thanks!

PR #3329, opened 2024-12-15
Update example configuration location in docker setup docs

louis-lau:
Without it, keys would not be generated and result in an insecure setup.


Koenkk:
Thanks!

PR #3330, opened 2024-12-15
How to unstuck RTCZCGQ11LM.md

VincentSC:
Added info how to unstuck the Aqara FP1. Source is https://github.com/Koenkk/zigbee2mqtt/issues/14130#issuecomment-1304349158 and trying multiple methods myself on two misbehaving devices. One got unstuck by pressing the button after 10 seconds, while the second I only could fix by pressing the button continuously.


Koenkk:
Thanks!

PR #3337, opened 2024-12-17
add image of Hue Lightstrip Plus V4

lhaussknecht:


Koenkk:
Thanks for the PR but this image is not used (it uses the `model` which is `8718699703424`)

PR #3351, opened 2024-12-24
Schneider Electric CCT5010-0001 was also sold under the name 550B1012 in Denmark

poizan42:
Not sure exactly how we usually document multiple model numbers for the same hardware when it's not quite a white label situation. It was sold with both Schneider Electric and "LK" (Lauritz Knudsen) branding on the device (see https://www.lk.dk/produkter/550B1012/wiser-tradlos-lysdaemper-for-indbygning-puck)

They exists as two different products in Schneider Electric's catalog

* https://www.se.com/dk/da/product/CCT5010-0001/connected-dimmer-wiser-micro-module/
* https://www.se.com/dk/da/product/550B1012/wiser-tr%C3%A5dl%C3%B8s-lysd%C3%A6mper-for-indbygning-puck/

\- but their ["dimmer guide"](https://www1.lk.dk/apps/dimmertool/dimmerGuide.ph7) makes it clear that they are in fact the same product with two different SKUs (or CRs whatever that means):
 
image

### DIFF
--- a/docs/devices/CCT5010-0001.md
+++ b/docs/devices/CCT5010-0001.md
@@ -15,7 +15,7 @@ pageClass: device-page
 
 |     |     |
 |-----|-----|
-| Model | CCT5010-0001  |
+| Model | CCT5010-0001, 550B1012  |
 | Vendor  | [Schneider Electric](/supported-devices/#v=Schneider%20Electric)  |
 | Description | Micro module dimmer |
 | Exposes | ballast_minimum_level, ballast_maximum_level, dimmer_mode, light (state, brightness, level_config), effect, power_on_behavior, linkquality |


Koenkk:
This part of the docs is generated, please add it as a whitelabel to https://github.com/Koenkk/zigbee-herdsman-converters/blob/9bb99f0c972c5a339c846ac966bf2739d591fe42/src/devices/schneider_electric.ts#L663

Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/9bb99f0c972c5a339c846ac966bf2739d591fe42/src/devices/schneider_electric.ts#L694

poizan42:
Closing in favor of https://github.com/Koenkk/zigbee-herdsman-converters/pull/8569

PR #3353, opened 2024-12-24
Updated troubleshooting page

Ricc68:
Made following changes as per https://github.com/Koenkk/zigbee2mqtt/issues/23282#issuecomment-2560242661:

- Added HASH_ERROR_TIMEOUTS and HOST_FATAL_ERROR cases to troubleshooting page
- Changed 'fails to start' to 'fails to start/crashes runtime'
- Added requirement to first time run npm install to avoid dependancy issues with 'npm ci'

@Koenkk please review this commit and let me know if I can do something better.

Sorry for too long commit description: it's first time in years that I use git so had to re-learn and I did not remember that is not a good idea to use commit description as changelog.


Koenkk:
Many thanks!

docs/guide/installation/20_zigbee2mqtt-fails-to-start_crashes-runtime.md
```diff
@@ -208,3 +208,34 @@ This happens when you edit one or more of the `pan_id`, `network_key` or `ext_pa
 ```
 ext_pan_id: [0x39,0xaf,0x4d,0x83,0xh2,0xdc,0xb3,0x89]
 ```
+
+## Error: runtime crash with `status=HASH_ERROR_TIMEOUTS` or fail to start with `status=HOST_FATAL_ERROR`
```
Nerivec, line 212: The error is `ASH_ERROR_TIMEOUTS` not `HASH_ERROR_TIMEOUTS`.
`HOST_FATAL_ERROR` is also too generic to put in this context (it can have many, many causes).

I think it's best to remove the references to ember-specific errors since I'm sure the same applies to other stacks. Maybe a header like this: `Regular crashes with timeout or failure to start after serial opened`.

Ricc68: My bad for the typo.
I agree that a generic header will help more than references to ember-specific errors.
I will change the header as per your suggestion.

PR #3360, opened 2024-12-27
Error header more generic to cover all stacks, not just ember

Ricc68:
As title says, following the suggestion from @nerivec in https://github.com/Koenkk/zigbee2mqtt.io/pull/3353, I have changed the error header to be more generic and cover all stacks rather then refer to ember-only specific errors.


Koenkk:
Thanks!

Nerivec:
@Ricc68 Any chance you could add a paragraph with some explanations for [this issue](https://github.com/Koenkk/zigbee2mqtt/issues/23151#issuecomment-2497693754) since you know your way around proxmox? 😜

Ricc68:
> @Ricc68 Any chance you could add a paragraph with some explanations for [this issue](https://github.com/Koenkk/zigbee2mqtt/issues/23151#issuecomment-2497693754) since you know your way around proxmox? 😜

I'm trying to get more info about the problematic NIC. Stalls and disconnects are symptom of a weak hardware, like for example an USB NIC (which typically overheats) or even an onboard NIC which is overheating.
I can write a paragraph about general quality of the NICs because of course if you have a networked router you want a stable network connection.
Let's see if we can get more info on that specific case.

PR #3368, opened 2024-12-29
Update 01_zigbee_network.md

chris-1243:
Add new device


Koenkk:
Thanks!

Koenkk:
`LED2112` is not a supported z2m device (therefore the build failed after merging). Could you make a new PR with the correct model id?

chris-1243:
Correction done in #3379

PR #3372, opened 2024-12-30
Podman with SELinux needs :z option

jburgess777:
When running on Fedora, where SELinux is enabled by default, the `:z` flag is needed on the volume mount to prevent permission errors like:
```
cp: can't stat '/app/data/configuration.yaml': Permission denied
```


Koenkk:
Thanks!

PR #3380, opened 2025-01-01
Add Hue Globe E27 smart bulb

LaurentvdBos:
New device, PR in herdsman-converts will follow in a couple of minutes.


LaurentvdBos:
Looks like support was added over Christmas, but without a picture, so this PR may be nice to merge.

PR #3396, opened 2025-01-05
Update home_assistant.md

mundschenk-at:
Remove sentence about events becoming "recommended" with 2.0.0 as the implementation is still marked "experimental" at this time.


mundschenk-at:
@Koenkk I see that you have already removed the "recommended" sentence, but maybe we should reword the following sentence as well.

Koenkk:
thanks!

PR #3397, opened 2025-01-05
Update TS0601_soil_3.md

dverhaert:
Added pairing instructions.

### DIFF
--- a/docs/devices/TS0601_soil_3.md
+++ b/docs/devices/TS0601_soil_3.md
@@ -23,6 +23,14 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
+## Notes
+
+
+### Pairing
+Pair the sensor to Zigbee2MQTT by holding it close to the coordinator and
+holding the inside button, next to the two AAA batteries, for about 5-10 seconds.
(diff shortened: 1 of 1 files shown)


joseruisa:
hi there

Seems some issues with battery level on device, since from some HA updates.
Also I have a rechargeable batteries FULL 1.2V and device shows 20% batteries

Also the reading data (temp/moisture) not showing. Comparing a similar device I have
I have the device on my desk near to Router (and communicating to RPI)

Captura de ecrã 2025-01-25, às 11 43 04

Daniel-WW:
I would like to add my experience to this issue:

In my setup, battery levels in Zigbee2MQTT are also reported incorrectly. After inserting freshly charged NiMH batteries, several devices still show very low battery levels — typically 11%, 16%, or 21%. I verified the battery voltage using a multimeter, and each cell measures around 1.35 V, indicating a healthy charge level.

Despite this, Zigbee2MQTT reports the battery state as low, sometimes accompanied by a warning. Importantly, the reported battery level does not update, even after waiting for more than a day.

It appears that the battery level calculation may not be well-suited for rechargeable NiMH cells, or that incorrect thresholds are being applied. It would be helpful to have an improved detection method or a way to configure custom calibration values for different battery types.

PR #3406, opened 2025-01-08
Add note regarding click_mode change for WXKG16LM

guntiss:
I'm my case I was unable to change click mode for WXKG16LM switch. Found somewhere that in order to be able to change click mode need to first press button 5 times and it worked. Therefore adding this note so other people are informed as well.


docs/devices/WXKG16LM.md
```diff
@@ -60,6 +60,8 @@ To read (`/get`) the value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME
 To write (`/set`) a value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/set` with payload `{"click_mode": NEW_VALUE}`.
 The possible values are: `fast`, `multi`.
 
+Note: You may need to press button 5 times in order to change click mode.
```
Koenkk, line 63: Thanks but this part of the docs is generated, could you update: https://github.com/Koenkk/zigbee-herdsman-converters/blob/5009796f1c240de3ac2d59ff9302cb0577a5280b/src/devices/lumi.ts#L3670 instead?

PR #3409, opened 2025-01-08
Adding pairing instructions for Lidl lamp floor 14153905L

TO19:
Adding pairing instructions that are the same as 14158704L.md


Koenkk:
Thanks!

PR #3410, opened 2025-01-08
Update ZG-101ZL.md

Chartreusito:
I did not find any generic documentation concerning "operation_mode", i guess it would be useful somewhere, even if the behaviour changes from device to device.

For what it's worth, i got a [Moes button](https://www.zigbee2mqtt.io/devices/ERS-10TZBVB-AA.html) that is recognized as this device. Both behaves in the same way.


Koenkk:
Thanks!

PR #3413, opened 2025-01-09
feat: Configurator to generate basic `configuration.yaml`

Nerivec:
Can be expanded at will, but might want to keep it for "simple config" scenario.

configurator

@Koenkk might want to push this to master? (also fixed a permit_join leftover)


Koenkk:
That's really cool, thanks!

> @Koenkk might want to push this to master? (also fixed a permit_join leftover)

Changes which also apply to the current release can always go straight to master

PR #3418, opened 2025-01-11
Update external_converters.md documentation

Timieh:
- Added Breaking change warning note.
- Changed to correct new location of external converters
- Added a small debug tip how to check the loading of your newly made converter

### DIFF
--- a/docs/advanced/more/external_converters.md
+++ b/docs/advanced/more/external_converters.md
@@ -4,11 +4,17 @@ sidebar: auto
 
 # External converters
 
+:::tip WARNING V1.x > V2.x
+Breaking change:
+External converters and extensions
+    The external_converters setting is no longer used. Instead all external converters inside /external_converters directory are now automatically loaded. Make sure to move all your external converters to this directory. And remove the external_converters in your configuration file.
+:::
+
 Zigbee2MQTT uses [zigbee-herdsman-converters](https://github.com/Koenkk/zigbee-herdsman-converters) to parse messages to and from devices.
 
 External converters provide a way to test support for new devices, they work identically to internal converters.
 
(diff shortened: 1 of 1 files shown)


docs/advanced/more/external_converters.md
```diff
@@ -72,3 +78,10 @@ To save a converter at runtime, send a message to `zigbee2mqtt/bridge/request/co
 ## Remove converter
 
 To remove a converter at runtime, send a message to `zigbee2mqtt/bridge/request/converter/remove` with payload `{"name": "my-first-converter.js"}`. The file will be deleted from `data/external_converters/`.
+
+:::tip Wan't to check if your freshly made converter loaded correctly?
```
Koenkk, line 82: I think this fits better under `Once your converter is ready`?

docs/advanced/more/external_converters.md
```diff
@@ -4,19 +4,25 @@ sidebar: auto
 
 # External converters
 
+:::tip WARNING V1.x > V2.x
```
Nerivec, line 7: All breaking changes were added to a single [discussion](https://github.com/Koenkk/zigbee2mqtt/discussions/24198), and not in the docs. Shouldn't break this pattern here, it will only cause confusion.

docs/advanced/more/external_converters.md
Nerivec: Also, this is only valid for converters loaded via file system, not via MQTT (no need to restart there, just pair or re-interview the targeted device).

fred-c1:
Suggestion:   if possible ?????     also update/correct "Zigbee2MQTT 2.0.0 breaking changes" discussion:
https://github.com/Koenkk/zigbee2mqtt/discussions/24198
(which was my "guide" today)

PR #3425, opened 2025-01-12
Added pairing info to PRO-900Z

Nick-Sills:
The "Heating in process" section needs to be removed as it's no longer relevant with the new `running_staus`, and the Climate section needs a `running_status` descrition. As that part is auto generated I'm not sure how this is done

Thanks.


Koenkk:
Thanks!

PR #3433, opened 2025-01-15
Correct the Tuya IH012-RT02 Image

ellnic:
The image for the Tuya IH012-RT02 displays incorrectly when using the Z2M web UI. In the image folder, the jpg is correct, but the png is incorrect.


Koenkk:
Thanks!

PR #3434, opened 2025-01-15
Update LSPA9.md

yannpub:
Adding note regarding behaviour of this device

/|\ Important Information /|\
Some of these devices (if not all) randomly switch off. Recommendation is not to use these for critical appliances as freezers or fridges, nor appliances which would be stopped during a programs like washing machines. Only solution found so far is to monitor them and setup some automations to restart these when their status is off for a certain duration. As of Jan 2025, no OTA update is available.


Koenkk:
Thanks!

PR #3439, opened 2025-01-18
Added paragraph on network adapters to troubleshooting guide

Ricc68:
As per @Nerivec request in https://github.com/Koenkk/zigbee2mqtt.io/pull/3360#issuecomment-2565168831, I have written some guidance about network adapters in the troubleshooting guide.

### DIFF
--- a/docs/guide/installation/20_zigbee2mqtt-fails-to-start_crashes-runtime.md
+++ b/docs/guide/installation/20_zigbee2mqtt-fails-to-start_crashes-runtime.md
@@ -223,6 +223,16 @@ This happens when you edit one or more of the `pan_id`, `network_key` or `ext_pa
 ext_pan_id: [0x39,0xaf,0x4d,0x83,0xh2,0xdc,0xb3,0x89]
 ```
 
+## Zigbee adapters over the network: use good network adapters on Zigbee2MQTT server
+
+If you have a WiFi or ethernet-connected Zigbee adapter, Zigbee2MQTT is communicating with the Zigbee adapter over the LAN through serial-over-IP protocol.  
+The best setup for this situation is to use the ethernet port embedded into the Zigbee2MQTT server motherboard which guarantees reliability of communications in all load conditions.  
+As a second choice you can use the onboard WiFi adapter which should as well be designed for reliability, but also consider the stability of your WiFi network.  
+If all the onboard adapters are in use and you need to add another network adapter, the best choice is to install an internal network card on the PCIe bus, with proper cooling design.
+
+The use of USB-WiFi or USB-ethernet adapters is discouraged because despite the apparent equivalence to the onboard adapters in terms of specifications, they are designed in small enclosures, often not well ventilated and tend to overheat.  
+These adapters are known to stall or stop working in case of high loads or overheating.
+
 ## Error: regular crashes with timeout errors or failure to start after the serial port is opened
 
 These errors may occur when the serial communication between the ZigBee dongle and Zigbee2MQTT unexpectedly stops working.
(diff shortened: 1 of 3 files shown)


Nerivec:
Wondering if that paragraph should be moved to https://www.zigbee2mqtt.io/advanced/zigbee/02_improve_network_range_and_stability.html#usb-based-adapter
to be more in-line with the rest of the page?

Ricc68:
Hmmm, that page refers to **Zigbee** adapters, routers and network while my explanation refers to **LAN** adapters and network.
I think that moving it to that page may cause confusion, in the reader, between LAN and Zigbee networks which are different and not interchangeable.

Koenkk:
Not having this correct makes z2m crash right? So I guess the page is correct. Maybe it would be good to describe what errors can be seen if this is not correct (like the Ember timeout errors)

Ricc68:
The last commit addresses @Koenkk observation.

docs/guide/installation/20_zigbee2mqtt-fails-to-start_crashes-runtime.md
```diff
@@ -223,6 +223,43 @@ This happens when you edit one or more of the `pan_id`, `network_key` or `ext_pa
...
+```
+[2024-06-24 03:37:22] error: zh:ember:uart:ash: Received ERROR from NCP while connecting, with code=ERROR_EXCEEDED_MAXIMUM_ACK_TIMEOUT_COUNT.
+[2024-06-24 03:37:22] error: zh:ember:uart:ash: ASH disconnected | NCP status: ASH_NCP_FATAL_ERROR
+[2024-06-24 03:37:22] error: zh:ember:uart:ash: Error while parsing received frame, status=ASH_NCP_FATAL_ERROR.
+```
+
+where a timeout occurred on the serial-over-IP protocol, or:
+
+```
+[2024-06-24 03:37:24] warning: zh:ember:uart:ash: Frame(s) in progress cancelled in [1ac1020b0a527e]
```
Nerivec, line 242: It should not produce this one.
But then again, weird network behaviors can... see https://github.com/Koenkk/zigbee2mqtt/issues/23120#issuecomment-2585908365 😨

Ricc68:
What I see is an ASH ACK followed by an ASH NAK from 251 (SLZB) to 112 (how's that the firmware sends two answers like that, where are the requests from 112?), then, after the NAK, 112 closes the connection (112 is Zigbee2MQTT). Then after 2 seconds 112 reopens the connection and issues an ASH Reset, but 251 answers again ASH NAK.
No wonder this communication is garbled, 251 insisting to reply ASH NAK: probably the firmware ASH parser got somehow out of sync in the SLZB.

This may very well be caused by weird network behaviour, especially if some serial message is corrupted or lost in the stream (packet loss, etc).

This may have many causes, loops on the network as said in that message, but also by unstable connection caused by hardware instabilities or failures like I have described in my PR.

Anyway hats off to @sorenfalch for the nice analysis, including not only the TCP dump but also the dissector that helped understand how much the SLZB firmware was confused.

Koenkk:
@Nerivec ready for merge?

Ricc68:
Just to give a little more context, typically an embedded network stack running on an MCU like the SLZB is inherently limited by the very low resources of the MCU.
Retransmissions are usually a hint that packets are being discarded: see https://docs.espressif.com/projects/esp-faq/en/latest/software-framework/protocols/lwip.html#what-happens-when-esp8266-receives-a-tcp-out-of-order-message
If the network stack is discarding packets, the serial stream breaks and things start to go wrong with the parser of the serial protocol in the MCU.

PR #3442, opened 2025-01-19
Add calibration failed hint to TRVZB.md

kR1StYaN:
If you use the wrong adapter, did not mount it properly or at all, you cannot pair it. 
I had trouble with pairing and this hint would have helped me a lot. Took me some time to figure it out, why I could not pair it.


Koenkk:
Thanks!

PR #3445, opened 2025-01-21
Update TS0505B_1.md

DomiiBunn:
Add TechToy Smart Bulb RGB 9W E27 ZigBee to the whitelabel list


Koenkk:
Thanks for the PR but this part of the docs is generated, please update https://github.com/Koenkk/zigbee-herdsman-converters/blob/a590eda0fd71ad412a8965acccac1f688d3b4670/src/devices/tuya.ts#L1570 instead.

PR #3454, opened 2025-01-23
Update E2134.md

hansfriedringelpuetz:
Added a solution for the regularly appearing problem of the IKEA Vallhorn sensor reporting changes in occupancy with a delay of up to several seconds.

### DIFF
--- a/docs/devices/E2134.md
+++ b/docs/devices/E2134.md
@@ -34,6 +34,10 @@ After a few seconds it turns back on and pulsate. When connected, the light turn
 ### Common problems
 Many has experienced using 1.2v AAA batteries (instead of the common 1.5v) makes the device behave correctly. Using 1.5v might result in unexpected behavior like unable to connect and false-positive motion detection.
 
+In some cases, it has been found that the motion detector occasionally only reports occupancy after a delay of several seconds. This is most likely due to a configuration in the reporting settings of the device, which is often set unfavorable by default. If you find the "Min rep interval" value of the attribute "occupancy" to be set to something other than 0, setting it to 0 and applying the change might resolve the issue. By doing that you allow the device to report changes in occupancy as soon as they happen and not only every x seconds. This value can be altered in the device settings under the tab "Reporting", typically at endpoint 2. If you receive error messages, when trying to apply the new value, it can help to move in front of the sensor while clicking apply to keep it awake.
+![Vallhorn Reporting](https://github.com/user-attachments/assets/5dd830e5-e415-4b47-bf13-ea4b456170a3)
+
+
 ### Mods
 As an alternative to the use of 1.2V batteries as mentioned above, the battery voltage of normal batteries can also be reduced by using a diode. Experienced user can insert a diode in series in the battery circuit at the point marked in red. An MBR0520, for example, with a Vf of 0.45V worked well here. Other types will probably work well too. At the point shown in the photo, the conductor track must be scratched/split open and the diode inserted as a bridge between the battery plate and the soldering point of the polyfuse (FR1). Please note the direction! Alternatively, the polyfuse (FR1) can be removed and the diode inserted instead, then you don't have to sratch the conductor. 
 


Koenkk:
Would it be better if we make these the default values?

hansfriedringelpuetz:
First of all let me quickly thank you for creating this outstanding tool, that is zigbee2MQTT! I am using it for my Home Assistant for a long time now and it is just great! I dont know what I would do without it! :heart:

Lets get back to the topic: Yes, I think we should absolutely change the default settings! (I didn't know, that this is possible :smiley:) Since the IKEA Vallhorn is widely available and quite often in use, I think it is important to bring this configuration to as many Vallhorn users as possible. But it shouldn't be the illuminance reporting with min:0 like in your example code line, but rather the occupancy!

This is my occupancy reporting config at the moment: 
grafik

with the min rep interval for occupancy now set to 0 the sensor works rock solid. And it seems to be the same for lots of other people with the same issue: [https://www.reddit.com/r/homeassistant/comments/1hy7yay/found_a_solution_to_ikea_vallhorm_motion_sensors/](https://www.reddit.com/r/homeassistant/comments/1hy7yay/found_a_solution_to_ikea_vallhorm_motion_sensors/) (this is my post regarding the Vallhorn reaction delay issue on reddit)

Do you know what the default reporting config for the IKEA Vallhorn is? I don't remember any more. Tomorrow I will buy some more of the Vallhorn sensors at my local IKEA, so I will hopefully be able to check it for myself then.

Do you think we should check, if further changes to the other reporting default values would be advantageous (illuminance and remaining battery)? Or does that not make any sense, because they always get reported too, when occupancy is detected?

One last question at the end: is there a way to roll out the new reporting config to Z2M users, that are already running Vallhorn sensors? Maybe something like a pop-up after the next update that shortly describes the issue with the Vallhorn sensors and asks, if they want to automatically apply the changes to their configuration?

hansfriedringelpuetz:
> Would it be better if we make these the default values? Can be done by changing https://github.com/Koenkk/zigbee-herdsman-converters/blob/4471082e3040f3ca42c7cf80b3aa7eef9fe36ae4/src/devices/ikea.ts#L906 to
> 
> ```
>         extend: [addCustomClusterManuSpecificIkeaUnknown(), m.occupancy(), m.illuminance({reporting: {min: 0}}), m.identify({isSleepy: true}), m.battery()],
> ```

First of all let me quickly thank you for creating this outstanding tool, that is zigbee2MQTT! I am using it for my Home Assistant for a long time now and it is just great! I dont know what I would do without it! :heart:

Lets get back to the topic: Yes, I think we should absolutely change the default settings! (I didn't know, that this is possible :smiley:) Since the IKEA Vallhorn is widely available and quite often in use, I think it is important to bring this configuration to as many Vallhorn users as possible. But it shouldn't be the illuminance reporting with min:0 like in your example code line, but rather the occupancy!

This is my occupancy reporting config at the moment: 
grafik

with the min rep interval for occupancy now set to 0 the sensor works rock solid. And it seems to be the same for lots of other people with the same issue: [https://www.reddit.com/r/homeassistant/comments/1hy7yay/found_a_solution_to_ikea_vallhorm_motion_sensors/](https://www.reddit.com/r/homeassistant/comments/1hy7yay/found_a_solution_to_ikea_vallhorm_motion_sensors/) (this is my post regarding the Vallhorn reaction delay issue on reddit)

Do you know what the default reporting config for the IKEA Vallhorn is? I don't remember any more. Tomorrow I will buy some more of the Vallhorn sensors at my local IKEA, so I will hopefully be able to check it for myself then.

Do you think we should check, if further changes to the other reporting default values would be advantageous (illuminance and remaining battery)? Or does that not make any sense, because they always get reported too, when occupancy is detected?

One last question at the end: is there a way to roll out the new reporting config to Z2M users, that are already running Vallhorn sensors? Maybe something like a pop-up after the next update that shortly describes the issue with the Vallhorn sensors and asks, if they want to automatically apply the changes to their configuration?

Koenkk:
I changed this now, it is currently not automatically pushed, users can re-configure it by using the yellow refresh button on the frontend.

Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now.

> Do you think we should check, if further changes to the other reporting default values would be advantageous (illuminance and remaining battery)? Or does that not make any sense, because they always get reported too, when occupancy is detected? (when I look in the logs it seems like that)

I think the current reporting settings should be OK, note that more frequent reporting also decreases battery life.

hansfriedringelpuetz:
Perfect, thanks!! :+1:
can you say approximately when the change will reach stable?

Koenkk:
New release is every 1st of the month.

hansfriedringelpuetz:
great, cheers mate! :v:

PR #3457, opened 2025-01-24
Clarify ZBDongle-E instructions

io41:
Make it clear that the PDF flashing instructions link to old firmware, and where to find the correct firmware.


Koenkk:
Thanks!

PR #3499, opened 2025-02-07
Update RF_274_T.md

AlexanderLeporiday:
Updated all instances of innr to Innr to match all other Innr vendor devices (Innr appeared twice on the search and vendor page)

image


Koenkk:
Thanks!

PR #3501, opened 2025-02-08
Update PTM_215Z.md

marvinrabe:
Add GIra + Senic switch to white-label. It took me a while yesterday to find the correct documentation for my EnOcean Device that is used in the Gira + Senic Switch.


Koenkk:
This part of the docs is generated, please update https://github.com/Koenkk/zigbee-herdsman-converters/blob/3cc7e89985b4bfa938b51cc906a32cff6416c595/src/devices/enocean.ts#L32 instead.

PR #3510, opened 2025-02-12
Add usability warning to E2013 / IKEA PARASOLL device page

duckfullstop:
This PR adds a warning message to the E2013 device page warning about a known issue when paired with certain routers, as discussed in https://github.com/Koenkk/zigbee2mqtt/issues/22579

As this issue seems to be affecting a large number of users with seemingly no possible fix available or on the horizon, I'm recommending that users be dissuaded from buying these sensors if they're specifically looking to use them with Zigbee2MQTT for the time being. If that's too harsh, I can remove the last sentence.

I'm presuming that `::: warning` can be used on device pages? If not, let me know and I'll refactor.

### DIFF
--- a/docs/devices/E2013.md
+++ b/docs/devices/E2013.md
@@ -26,7 +26,7 @@ pageClass: device-page
 
 ::: warning Known Issues with This Device
 
-The IKEA PARASOLL door/window sensor is known to suffer from serious usability issues in some Zigbee networks, resulting in some adapters regularly going unavailable / draining their battery.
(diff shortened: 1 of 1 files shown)


duckfullstop:
There seems to be at least one or two people reporting that using a Ember-based coordinator resolves the issue - perhaps we should tweak the wording to specifically call out the TI coordinators? Equally, however, there are those earlier in the thread reporting the same problem with those, too.

PR #3513, opened 2025-02-14
Add new IKEA images

chris-1243:
It is following PR #8801

https://github.com/Koenkk/zigbee-herdsman-converters/pull/8801


Koenkk:
Thanks!

chris-1243:
Should we rename e1603-e1702-e1708.png to e1603.png only according to the change in PR?

Koenkk:
this is fine but we also need a `E160x-E170x-E190x.png`

chris-1243:
I would suggest a copy/paste of e1603-e1702-e1708.png and we rename the file. This is the EU version.

Those models look all different anyway.

PR #3520, opened 2025-02-16
Update 501.40.md

buz-ch:
Put clarification on using multiple lights simultaneously (works for all of my 3 remote bought on Amazon.de in Jan/Feb 2025). Unsure if there are other firmware versions?

I see something like:
``` 
MQTT publish: topic 'zigbee2mqtt/LR-RGB-Remote-1', payload '{"action":"on_1","battery":100,"linkquality":172}'
MQTT publish: topic 'zigbee2mqtt/LR-RGB-Remote-1/action', payload 'on_1'
MQTT publish: topic 'zigbee2mqtt/LR-RGB-Remote-1', payload '{"action":"on_2","battery":100,"linkquality":172}'
MQTT publish: topic 'zigbee2mqtt/LR-RGB-Remote-1/action', payload 'on_2'
```


Koenkk:
thanks!

PR #3540, opened 2025-02-23
"adapter" to "coordinator"

gam-st:
Zigbee implies 3 types of devices:

- **Zigbee Coordinator (ZC)**
-- There is exactly one coordinator in each Zigbee network.
-- Responsible for initiating and configuring the network (choosing channel, PAN ID, etc.).
-- Can route packets and also act as a parent for end devices.
-- Typically has more memory and power resources compared to the other devices.

- **Zigbee Router (ZR)**
-- Optional device type; can exist in a network to help relay data.
-- Can route packets on behalf of other devices.
-- Can act as a parent to End Devices, allowing them to join the network and forward their messages.

- **Zigbee End Device (ZED)**
-- Optional device type; can exist in a network.
-- Has the most limited functionality: cannot route packets for others; only communicates with its parent (Coordinator or Router).
-- Typically optimized for low power, can “sleep” to conserve energy.
-- Suitable for simple sensor nodes or actuators.

The “adapter” can operate in multiple modes (ZC, ZR, and even ZED (for example, to control its onboard LED)). Therefore, it makes more sense to use the term “Coordinator” instead of “Adapter.”

I propose changing only the frontend and leaving the backend (e.g., page names, etc.) as is, so as not to affect search engine rankings.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -11,7 +11,7 @@ actions:
       type: primary
 features:
     - title: Compatible
-      details: Zigbee2MQTT supports various Zigbee adapters and a big bunch of devices.
+      details: Zigbee2MQTT supports various Zigbee coordinators and a big bunch of devices.
     - title: Integrations
       details: Zigbee2MQTT integrates well with most home automation solutions because it uses MQTT.
     - title: Open Source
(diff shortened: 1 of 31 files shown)


Koenkk:
Thanks! I propose to also rename all files and folders. Old pages can redirect to the new ones like this: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/information/mqtt_topics_and_message_structure.md

gam-st:
I wont be able to make those changes in pages names, so i propose to rename this "visible" part at least.

Koenkk:
> I wont be able to make those changes in pages names

Can I help somehow?

PR #3541, opened 2025-02-24
Add pairing instructions for Paulmann 500.48

matsimon:
As depicted in vendor documentation:
https://content.paulmann.com/media/98/c4/7f/1600991028/50048_V01.pdf

I've added some notes on validating the device (properly) entered pairing mode properly since I had to take a couple of attempts.


Koenkk:
Thanks!

PR #3549, opened 2025-02-26
Add a section for null values in SJCGQ11LM

andreadecorte:
As per discussion in https://github.com/Koenkk/zigbee2mqtt/issues/18243, null values are expected for some properties.

As this behavior is somehow surprising, add a note in the docs about that.


Koenkk:
I'm not sure if this is the correct place to document since this happens for every device. It first needs to report the state once for z2m to have a value.

andreadecorte:
It's probably mostly visible for sensors that are triggered infrequently like this one. It was just to avoid such bug reporting by documenting clearly the behavior. Happy to also change to a better doc location if you have any suggestion

Koenkk:
I fixed the issue, state should now be read when pairing.

Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now and in the next release which is every 1st of the month.

PR #3552, opened 2025-02-27
Correct color to color_rgb in CL-L02D.md

adstheorganic:


Koenkk:
I dont believe this is correct, what doesnt work when using color?

adstheorganic:
I may have been a bit overzealous in the changes, it was gone 2 am when I made them, let me got and test all the different scenarios and undo any that are wrong

adstheorganic:
image I cannot get {"hue": HUE, "saturation": SATURATION} to work for color, color_rgb or color_hs, and color_hs does not seem to work at all

{"color": {"h": H, "s": S, "b": B}} works for the white light and {"color_rgb": {"h": H, "s": S, "b": B}} works for the rgb light, but this is currently only mentioned under the rgb light endpoint section, same is true of {"hsb": "H,S,B"}, {"h": H, "s": S, "v": V}, {"hsv": "H,S,V"},{"h": H, "s": S, "l": L}, and {"hsl": "H,S,L"}

Also the section underneath mentions Transtion being able to be applies, to any of the above, but so far I'm yet to find a single transition where it has worked, trying to change the color/brightness or color temp with a transition just seems to apply it immediately anyway

adstheorganic:
Ahh okay, I understand now, the documentation being inconsistent was confusing, the section names are expecting to be suffixed with _rgb or _white based on the type of light, but that's not specified in the examples which you'd think you can just copy and paste to test, so brightness_rgb or bright_white, is just being called brightness.

Still the color_hs_white and color_hs_rgb don't work, and so far the only one I've found to work with the transition was color_temp_rgb, not even color_temp_white, and none of the other rgb ones

adstheorganic:
Other weird oddities, the color temp presets on the white light `coolest`, `cool`, `neutral`, `warmest` actually represent the values of 153, 250, 370, 370 respectively (in home assistant there are 4 default options that I would assume to be the same but they are 153, 190, 251, and 370)

brightness_move/step color_temp_move/step apply to the white light when not specified (brightness_move_white, color_temp_step_rgb to specifiy for example). When using this unspecified version move immediately goes to 100% or 0% depending on +-, but step works correctly, when specify they both seem to work correctly. For saturation and hue, move does not seem to recognised at all, (message received but not error) and step is received and returns the value, but the value doesn't change.

I know most of these are probably on aqara, but I'm just mentioning them here so if someone take a look at why the docs are the way they are, they have this as a reference point, and this helps me keep a log which I can report to aqara directly

adstheorganic:
If you'd rather we leave the transition section as I'm sure I've not tested everything, then feel free to add it back in, but I tested a lot and that was the only one I found

docs/devices/CL-L02D.md
```diff
@@ -116,18 +118,20 @@ The direction of move and step can be either up or down, provide a negative valu
...
-  "brightness_move": 0, // Stop moving brightness
-  "brightness_step": 40 // Increases brightness by 40
-  "color_temp_move": 60, // Starts moving color temperature up at 60 units per second
-  "color_temp_move": "stop", // Stop moving color temperature
-  "color_temp_step": 99, // Increase color temperature by 99
-  "hue_move": 40, // Starts moving hue up at 40 units per second, will endlessly loop (allowed value range: -255 till 255)
-  "hue_step": -90, // Decrease hue by 90 (allowed value range: -255 till 255)
-  "saturation_move": -55, // Starts moving saturation down at -55 units per second (allowed value range: -255 till 255)
-  "saturation_step": 66, // Increase saturation by 66 (allowed value range: -255 till 255)
+  "brightness_move_rgb": -40, // Starts moving brightness down at 40 units per second
```
Koenkk, line 125: I don't think this works

adstheorganic: yeah it just goes to 0% or 100% (depending on positive or negative), but it does do something. I put that as a note above, as it's the same for all moves, as I've not looked into whether this is an z2mqtt issue that could be fixed, or an aqara issue and wont be, so maybe this will work in the future,

Koenkk: This is likely an Aqara specific issue, I suggest adding a section under the `## Notes` part of this document. (as these docs are auto generated and harder to adjust for just a single device)

PR #3565, opened 2025-03-03
Updates to Hue 2m outdoor light strip

trenchcoatjedi:


Koenkk:
This part is generated from https://github.com/Koenkk/zigbee2mqtt.io/blob/6b7c094f650f24f655a224513290a801b3a558f9/docgen/device_page_notes.ts#L14 so that needs to be updated instead, maybe rename `bulb` to `light` to be more generic?

Koenkk:
Thanks!

PR #3568, opened 2025-03-04
Update DJT11LM.md to clarify calibration behavior and firmware update…

drueppela:
… availability

This update adds a detailed note clarifying that x_calibration, y_calibration, and z_calibration  in Zigbee2MQTT do not alter the raw sensor data reported by the Aqara DJT11LM. Instead, these offsets  are applied internally for certain calculations, which can lead users to think calibration hasn't worked  when they still see unmodified x_axis/y_axis/z_axis values.

Additionally, a firmware updates section was added to explain that no publicly available firmware files  exist for the DJT11LM. As a result, firmware updates via Zigbee2MQTT are not possible. Any official  Aqara firmware updates would have to be delivered via the Aqara or Mi Home hub/app, if and when  they become available.


docs/devices/DJT11LM.md
```diff
@@ -73,7 +83,13 @@ You can fine tune the values of the offset by trying other sides and picking val
 Remember that the device sends accelerometer values a few seconds after the actual tilt event.
 <!-- Notes END: Do not edit below this line -->
```
Koenkk, line 84: Don't edit below this line

drueppela: @Koenkk : Sorry for my mistake. What do I have to do? Do you agree with the content of the proposals?

drueppela: @Koenkk : I transferred my proposal regarding the clarification of calibration Options to https://github.com/Koenkk/zigbee2mqtt.io/pull/3572

PR #3572, opened 2025-03-05
Update DJT11LM.md - Clarification regarding calibration behavior

drueppela:
This update adds a detailed note clarifying that x_calibration, y_calibration, and z_calibration in Zigbee2MQTT do not alter the raw sensor data reported by the Aqara DJT11LM. Instead, these offsets are applied internally for certain calculations, which can lead users to think calibration hasn't worked when they still see unmodified x_axis/y_axis/z_axis values.


Koenkk:
Thanks!

PR #3574, opened 2025-03-05
Tuya 6 gang wall switch

Jackylalala:
Tuya 6 gang wall switch
TS0601 model


Koenkk:
This should be called `SSW06G.png` but we already have a picture for it so closing this.

PR #3579, opened 2025-03-07
Add image for Envilar 7848 Dimmable LED Driver

HRading:
Add image for Envilar 7848 Dimmable LED Driver.
(I will also add the converter to the [zigbee-herdsman-converters] repo


Koenkk:
Thanks!

PR #3580, opened 2025-03-08
Adding Hue Go portable Accent light (915005822501)

thedavidthomas:
duplicated 7602031P7 as the device has a new model number of 915005822501


Koenkk:
Thanks!

PR #3589, opened 2025-03-11
Update connect_to_a_remote_sonoff_zbbridge.md

charliesjc:
Use zstack instead of ezsp


docs/advanced/remote-adapter/connect_to_a_remote_sonoff_zbbridge.md
```diff
@@ -26,7 +26,7 @@ Keep in mind that the EZSP support is currently **experimental**.
 ```yaml
 serial:
     port: 'tcp://192.168.2.13:20108'
-    adapter: ezsp
+    adapter: zstack
```
Koenkk, line 29: I believe this device uses a Silicon Labs chip? So `ezsp` should be correct

charliesjc: You're right. As soon as I saw this in my inbox I realized I'm using a ZBBridge Pro.

PR #3614, opened 2025-03-19
update 3R 3RCB01070Z

3reality-support:


3reality-support:
Where else is there an error？？？

3reality-support:
why，my picture is 512*512 and  the background transparent

Koenkk:
it is good now, thanks!

PR #3615, opened 2025-03-19
Update easyCodeTouch_v1.md

los7702:
Text update to reflect new changes in the converters repository

Aditional edit in the pincode category due to the easycode locks never accepts the /get  pin_code commands always error sending that command


Koenkk:
These docs are automatically generated based on the `exposes` of the definition, so no need to update it here.

PR #3616, opened 2025-03-19
Documentation/zbm5 1 c 120/action states with state commands

radovanobal:
There are some states missing for this device that get sent out when you change the on off state by sending the state on or off commands.
When the state command turns the switch off it publishes the action off_l1 and when it turns it on, then it sends on_l1


docs/devices/ZBM5-1C-120.md
```diff
@@ -84,3 +84,10 @@ Value can be found in the published state on the `action` property.
 It's not possible to read (`/get`) or write (`/set`) this value.
 The possible values are: `toggle_l1`.
 
+State change (e.g., a `state` command sent).
```
Koenkk, line 87: Please move above `<!-- Notes END: Do not edit below this line -->`

PR #3620, opened 2025-03-20
Update ID-UK21FW09.md

aalijag97:
Added Re-pairing description when using the reset button instead of the knob.


Koenkk:
Thanks!

PR #3635, opened 2025-03-26
Update the latest flashing method for SONOFF Dongle

Ash-xiao:
Update the latest flashing method for SONOFF Dongle.

### DIFF
--- a/docs/guide/adapters/emberznet.md
+++ b/docs/guide/adapters/emberznet.md
@@ -45,6 +45,7 @@ The use of `adapter: ezsp` is now deprecated. See [https://github.com/Koenkk/zig
     - Multi-devices by [@darkxst](https://github.com/darkxst/): [Silabs Firmware Flasher](https://darkxst.github.io/silabs-firmware-builder/)
     - For SkyConnect by [@NabuCasa](https://github.com/NabuCasa): [SkyConnect Flasher](https://skyconnect.home-assistant.io/firmware-update/)
     - SMLight: [Flasher](https://smlight.tech/flasher/)
+    - For SONOFF Dongle: [SONOFF Dongle Quick Flasher](https://dongle.sonoff.tech/sonoff-dongle-quick-flasher/)
 - Command-line based:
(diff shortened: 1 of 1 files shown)


docs/guide/adapters/emberznet.md
```diff
@@ -103,8 +104,7 @@ With external antenna. Only supports software flow control (make sure not to set
 
 - [Product Page](https://sonoff.tech/product/gateway-and-sensors/sonoff-zigbee-3-0-usb-dongle-plus-e/)
 - [Coordinator firmware](https://sonoff.tech/product-review/how-to-use-sonoff-dongle-plus-on-home-assistant-how-to-flash-firmware/)
-- [Flashing](https://sonoff.tech/wp-content/uploads/2022/08/SONOFF-Zigbee-3.0-USB-dongle-plus-firmware-flashing-.pdf)
-    - **Note**: The PDF links to an old firmware. See all [NCP_X.Y.Z](https://github.com/itead/Sonoff_Zigbee_Dongle_Firmware/tree/master/Dongle-E) versions and pick [7.4.4](https://github.com/itead/Sonoff_Zigbee_Dongle_Firmware/tree/master/Dongle-E/NCP_7.4.4) or newer.
```
marazmarci, line 107: I think it would be beneficial to keep the "old" instructions, because for example some users may want to perform the flashing on a headless (non-GUI) Linux machine, and the new quick flasher works in a browser, which requires a desktop environment on the host machine.

marazmarci (changes requested):

docs/guide/adapters/emberznet.md
marazmarci: Sorry, I realized that the command line flashing method in the PDF only applies to ZBDongle-P, and for ZBDongle-E there are only GUI instructions.

In this case I have another question: Don't you want to add the quick flasher also to [ZBDongle-P's instructions](https://www.zigbee2mqtt.io/guide/adapters/zstack.html)?

Ash-xiao: Yes. I think the "old" instructions for the ZBDongle-P are too complicated, and users may encounter difficulties during actual operation. Therefore, adding a Quick Flasher will help users operate more conveniently.

Ash-xiao: @marazmarci What should I do next?

PR #3652, opened 2025-03-31
Update SAGE206612.md - pairing instructions

antfurn:
Add pairing instructions.

Press and hold the reset button while inserting the battery. Hold until the LED starts blinking, this will force a reset/join. The LED will continue blinking to indicate its pairing.


Koenkk:
Thanks!

PR #3678, opened 2025-04-09
Update BHT-002_BHT-006.md

Onepamopa:
Add information about noisy devices.

### DIFF
--- a/docs/devices/BHT-002_BHT-006.md
+++ b/docs/devices/BHT-002_BHT-006.md
@@ -27,7 +27,8 @@ pageClass: device-page
 
 <!-- Notes END: Do not edit below this line -->
 
-
+### Warning. These devices are very, very noisy. Having more than 1 in your ZigBee network is a solid no-no.
+#### More information: https://github.com/Koenkk/zigbee2mqtt/discussions/21038
 
 
 ## Exposes
--- a/docs/devices/BHT-002_BHT-006.md
+++ b/docs/devices/BHT-002_BHT-006.md
@@ -27,7 +27,7 @@ pageClass: device-page
 
 <!-- Notes END: Do not edit below this line -->
 
-### Warning. These devices are very, very noisy. Having more than 1 in your ZigBee network is a solid no-no.
+### Warning. These devices are very, very spammy (more than 15 messages per second per device). Having more than 1 in your ZigBee network is a solid no-no.
 #### More information: https://github.com/Koenkk/zigbee2mqtt/discussions/21038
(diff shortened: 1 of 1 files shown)


docs/devices/BHT-002_BHT-006.md
```diff
@@ -27,7 +27,8 @@ pageClass: device-page
 
 <!-- Notes END: Do not edit below this line -->
 
-
+### Warning. These devices are very, very noisy. Having more than 1 in your ZigBee network is a solid no-no.
```
Koenkk, line 30: I guess you mean spammy (sending a lot of messages?) I initially thought you meant the device itself is making noise.

Onepamopa: Yes, extremely spammy to the point of 15+ messages per second per device..

nilshoell: I had issues even operating the z2m frontend if more than 4 devices were joined in the network. Denounce helps to reduce the mqtt messages, but of course does nothing against the congestion of the frequency. Devices also don't respond to any changes in the reporting settings.  
Here is a discussion with multiple affected people, including some frequency analysis to demonstrate the problem: https://github.com/Koenkk/zigbee2mqtt/discussions/21038.

@Onepamopa it might make sense to also add some info for troubleshooting: if one wants to remove the device from their network and keep using them standalone, it is adviced to first pull them into a new ZigBee network on a different channel, as the thermostats keep spamming even if removed from the network - or factory reset.

Onepamopa: I don't think removing the thermostats from the networks solves anything, because they still remain "configured" on that network (i.e. the antenna icon continues to be lit). I presume they keep sending data.

Onepamopa: BTW, from what I see - there's new comment in #21038 - with custom firmware that supposedly fixes the problem.

nilshoell: > I don't think removing the thermostats from the networks solves anything, because they still remain "configured" on that network (i.e. the antenna icon continues to be lit). I presume they keep sending data.
> 
> I still haven't found a way to reset them to factory settings in order to remove any trace of a zigbee connection.

Exactly, just removing it from z2m is not enough, that is why I advised to either move them to a temporary network on a different channel, or to do a factory reset.
The latter is possible from the on-device settings menu, which is reachable if you turn off the thermostat and press the mode an clock buttons iirc.

PR #3703, opened 2025-04-17
Add 4 switch model images.

devkaiwang:
Images for PR - https://github.com/Koenkk/zigbee-herdsman-converters/pull/9158

A same looking Tuya switch can be with or without scene support depending on the module, so I'm using same pictures for both cases.


Koenkk:
thanks!

PR #3705, opened 2025-04-19
Better images for ZG-101Z-D, ZD1-EN and Philips 8718699689308

ellnic:
1. A better 512px image for the ZG-101Z-D, and a 150px image that was missing.

2. The 512px image was not transparent despite appearing to have a checked background. Also added 150px image. Optimised both file sizes.

3. Better images for both 150px and 512px of Philips 8718699689308


Koenkk:
Thanks!

PR #3712, opened 2025-04-19
SNZB-02D: Hint at firmware version v2.3.0

raoulbhatia:
After Koenkk/zigbee-OTA#711 was merged,
help users to update their firmware more easily.


Koenkk:
Thanks!

PR #3721, opened 2025-04-22
Bacchus new devices

Bacchus777:
Added:
Mercury electricity meter
Mercury 3 phase electricity meter
Presence sensor v2
Water station
Water level meter


Koenkk:
Thanks!

PR #3727, opened 2025-04-24
Update RB_272_T.md

MSL-DA:


Koenkk:
This part of the docs is generated, please update https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/innr.ts

PR #3737, opened 2025-04-25
Add an image for the ZMS1-TYZ curtain motor.

kbuck1:
As requested in https://github.com/Koenkk/zigbee-herdsman-converters/pull/9235#issuecomment-2828332328

Image sourced from https://www.zemismart.com/products/zms1-tyz?VariantsId=17044


Koenkk:
Thanks!

PR #3777, opened 2025-05-13
add new devices

13717033460:
ZG-102ZM  ZG-204ZV


Koenkk:
Thanks! For next time `.jpg` images don't have to be added.

PR #3789, opened 2025-05-15
Fix incorrect formatting for Hue white ambiance GU10 with Bluetooth (9290019533)

fox34:
Formatted analog to 929001953301 and removed duplicate content.


Koenkk:
Thanks!

PR #3816, opened 2025-05-24
Add image for ZG9095B

ProtoxiDe22:


Koenkk:
thanks!

Koenkk:
@ProtoxiDe22 yes otherwise the device won't match the picture, you can create a new pr.

ProtoxiDe22:
@Koenkk I'm off my pc but vlcan absolutely do that tonorrow. Howeverl please note that the zigbee model (reported by the device in z2m) is ZG9095B, and after you merged this, i can already see the image properly on my install (pic attached). So i think it should be good this way?

Screenshot_20250526_215216_Home Assistant

It shows as not supported atm (as i removed the external converter), but the model number shown is without the sg- and i can already see the image (which i didn't put anywhere on my system)

PR #3841, opened 2025-06-02
Update TM-YKQ004.md

losip:
Added instructions for re-pairing

### DIFF
--- a/docs/devices/TM-YKQ004.md
+++ b/docs/devices/TM-YKQ004.md
@@ -26,8 +26,7 @@ pageClass: device-page
 ## Notes
 
 ### Reset/Re-pairing
-Unfortunately it is currently unknown how to bring the device in reset or re-pairing mode.
-Please be that kind and add this information, if your are in knowledge of it.
+Long press any key for about 10 seconds until the indicator light starts to flash.
 <!-- Notes END: Do not edit below this line -->
 
 


Koenkk:
thanks!

deathrattle666:
Hi losip,

the instruction does not work for both of my devices (I have a black and a white colored switch). Is the process for re-pairing really verified?

losip:
It works for me. Does your indicator light flash but then won't pair, or does the indicator light not flash when you hold down a button for 10 seconds?

deathrattle666:
No. I remove the battery and wait days. Then I activate pairing mode in HA (Z2M)  and re-insert a (fresh/new one)  battery.
I press any button on the switch and hold it for more than 10 seconds. But the indicator led fades approx. 3 seconds to off. Seems as it is in "normal" mode and not "re-pairing".
This with both of my devices. I have a white one and a black.

I once paired them as they were new (summer 2024). 

I examined Z2M via Z2M-Explorer. But no hint for the switch.

The original package says that the device code is "SMR-YKQ005-W4-ZIG".  Maybe a variant of the device and that could be the reason?

Is the picture of the device on the wiki correct? My is looking exactly as shown.

losip:
I've just tried again and the sequence is this:
1. Press and hold a button; light comes on immediately
2.  After about three seconds, light goes off
3. Keep holding button, 7 seconds later light comes on flashing
4.  Device is then discoverable in Z2M
5. 5. Light goes off when paired or after a further 30 seconds when it times out

Mine has the same part number

deathrattle666:
Thanks for your support and grateful passion!

My device fails at step 3. No more light after 7 seconds.
Gonna try it again several times. Maybe Iĺl need a better timing.

On the other side.... if it is indeed another device. I will ask Koen if I could send it to him as a gift and he can integrate it into upcoming versions of Z2M...

PR #3844, opened 2025-06-02
Update TS011F_plug_1.md

WCAve:
Added silvercrest plug to whitelabel devices.


Koenkk:
This part of the docs is generated, please add it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/c231daddcc02c3c989e27bb751cd85d9f01a0f02/src/devices/tuya.ts#L6516

PR #3848, opened 2025-06-06
Update E2134.md with illuminance behavior

zeroflow:
Add description that the VALLHORN illuminance sensor saturates at 1364lx (low overcast daylight)


Koenkk:
Thanks!

PR #3856, opened 2025-06-09
New Image für Device _TZE200_wzk0x7fq

Eisbaer2:
New Image für Device _TZE200_wzk0x7fq


Koenkk:
Thanks for the PR, however it is not needed anymore, this device should be detected as a ZG-102ZM, the picture has already been added to this repo.

PR #3868, opened 2025-06-12
New Adeo ZBEK-30 device

jpsminix:
New Adeo ZBEK-30 device


Koenkk:
Thanks! We will fix the issues in your zigbee-herdsman-converters PR

PR #3872, opened 2025-06-15
Update E2007.md

Wikibear:
Its possible to reset filter status, if you change a new filter, it's important to reset filter_replace and filter_age.


Wikibear:
Please check

Wikibear:
What I do is nonsense. But I will add possible way to reset filter message.

PR #3881, opened 2025-06-19
add Dlink motion sensor 

PharQ:


Koenkk:
A picture for this device is already available: https://github.com/PharQ/zigbee2mqtt.io/blob/add-Dlink-DCH-B122-motion-sensor/public/images/devices/DCH-B112.png

PR #3886, opened 2025-06-22
Basic airing ZBMINI

VincentSC:
Added more info on the basic pairing, as the original text only explained the exceptions.

### DIFF
--- a/docs/devices/ZBMINI.md
+++ b/docs/devices/ZBMINI.md
@@ -23,17 +23,28 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
+Here’s a more cohesive and unified version of the text, with a consistent tone and structure:
+
+---
+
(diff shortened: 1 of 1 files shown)


docs/devices/ZBMINI.md
```diff
@@ -23,17 +23,28 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
+Here’s a more cohesive and unified version of the text, with a consistent tone and structure:
```
Koenkk: Left over from ChatGPT?

VincentSC: Yes :)

I first wrote my part, which I mostly borrowed from the manual. There were now multiple styles of text, and I asked ChatGPT to make it more cohesive into one style. I seemingly only removed the last part of the text, and I did not notice the first line.

Koenkk:
Thanks!

PR #3888, opened 2025-06-23
Aqara H2 Warning about default Thread/Matter mode

LukeLR:
Aqara H2 devices are shipped with Thread/Matter firmware, Zigbee firmware requires flashing via the Aqara app.


Koenkk:
Thanks!

LukeLR:
You're welcome! Thanks for fast merging ;)

PR #3891, opened 2025-06-26
Update TS0726_4_gang_scene_switch.md with supported vendors

imammedo:
add notes to mention vendor ids "_TZ3002_aewsvjcu", "_TZ3002_phu8ygaw"

so they would come up during search

### DIFF
--- a/docs/devices/TS0726_4_gang_scene_switch.md
+++ b/docs/devices/TS0726_4_gang_scene_switch.md
@@ -23,8 +23,8 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
-
-
+## Notes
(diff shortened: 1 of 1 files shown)


imammedo:
it depends onn https://github.com/Koenkk/zigbee-herdsman-converters/pull/9549

Koenkk:
I'm not sure if this has added value, you have no clue what manufacturerName you bought until you pair it (and at the point it will show you it is supported or unsupported)

imammedo:
that's true and applies not only to manufacturerName.
But suppose one knows those values (aka preexisting network, or from another source),
then it increases visibility of what z2m supports vs others.

In nutshell it can't hurt.
Feel free to drop pull req if you think it has no value.

Koenkk:
The problem is that this will get outdated at some point (as it is not linked to the actual code). I will drop it for now.

PR #3894, opened 2025-06-27
add new sensor with custom fm TS0201-z-SlD

slacky1965:
Tuya temperature and relative humidity sensor _TS0201_TZ3000 with custom firmware


Koenkk:
Thanks!

PR #3910, opened 2025-07-02
Manual changes for B4Z and SC500ZB-v4

weichenb:
Hi Koen,
As indicated during last weekend, please find manual changes for B4Z and SC500ZB-v4 attached.
Picture for SC500ZB-v4 is fine-tuned.


docs/devices/B4Z.md
```diff
@@ -28,7 +28,6 @@ pageClass: device-page
 <!-- Notes END: Do not edit below this line -->
```
Koenkk, line 28: Please do not make changes below this line, the parts of the doc you edited are generated based on the definition, please modify that instead: https://github.com/Koenkk/zigbee-herdsman-converters/blob/e0e11cbafc9b25c7ac124a7ebe15ff305bbc82cc/src/devices/tuya.ts#L2861

weichenb: Hi Koen,
First the autogenerated content does not make sense to me. File that you referenced is fine in my eyes!
"indicator_mode" and "backlight_mode" are not exposed for devices "_TZ3210_xbpt8ewc", "_TZ3000_e3vhyirx", "_TZ3000_5iixzdo7", "_TZ3000_yruungrl". Despite that exclusion it is included in manual. Maybe it is worth looking in auto-generation-module to fix that issue? 
Second with such restrictions it is not clear to me how to add reasonable information to manual. In the moment I pause manual extension until there is more clarity.

Edit: 

- Can you please explain, why manual for "B4Z" also does not include exposed "switchTypeCurtain" ?
"Expose" is clearly included in https://github.com/Koenkk/zigbee-herdsman-converters/blob/e0e11cbafc9b25c7ac124a7ebe15ff305bbc82cc/src/devices/tuya.ts#L2861
- Where is an official document how to include documentation? I could not locate anything so far.
- Can you please explain, how entensive content gets entered in https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/046677476816.md ? There isn't anything defined in "notes"-section and in https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/philips.ts ?.

Koenkk: This is indeed a limitation of the doc generator, it does not have e.g. the manufacturerName available. To fix this a new definition in `tuya.ts` has to be created, can you make a PR for this?

weichenb: Hi @Koenkk,
for sure tuya.ts includes a definition of manufacturerName. Effectively for tuya-devices that definition is a mixture between a tuya-manfacturer-ID (not compliant with CSA manufacturer IDs used in Z2M) and a device-ID. 
But maybe you intended to reference on manufacturerCode instead of manufacturerName like some other device libs do?
Switch to manufacturerCode based on modernExtern would requrie from my point of view a complete redesign of tuya-device and libary definition. 
Potentially it might be much easier to adapt doc generation engine?

There are 3 obstacles to do that cleaning:
1. I have not found any guidance how to setup a comprehensive development environment under VScode that also allows a repository spanning analysis as well as a docker test installation. Up to now I perfomed local tests on my docker installation. However, that is not really efficient.
2. In oder to validate the proper device support after a redesign, the physical availability of the device would be required.
3. It seems that various different integration approaches have been applied for device support. In the moment I am not that profound yet, so that I would be able to redesign tuya-integration.

Concluding I am not able to implement changes and create a pull request for expected change. 

**Update:** Overall I think doc generator does not work properly. Beside the expected exclusion of backligh and indicator it did also NOT include "Switch type curtain" for both devices!

Koenkk: I fixed the doc generator, new page will be available after the next z2m release, preview: https://github.com/Koenkk/zigbee2mqtt.io/blob/dev/docs/devices/B4Z.md

PR #3914, opened 2025-07-05
Update TS0601_air_quality_sensor.md: added note about multiple message per second

GrumpyMeow:
Added note about multiple messages per second


docs/devices/TS0601_air_quality_sensor.md
```diff
@@ -23,7 +23,7 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
-
+This device is known to transmit a lot of messages, even multiple per second.
```
Koenkk, line 26: I would add a note here that this can degrade network performance

PR #3918, opened 2025-07-07
Update flashing_via_cc2538-bsl.md

gumida:
There is a small typo in the update instructions.
The update script is named `cc2538_bsl.py` instead of `cc2538-bsl.py`.
Following the flashing instructions from top to bottom, the extracted .py-file is located in a subfolder, so I suggest updating this in the instructions as well.


Koenkk:
thanks!

PR #3923, opened 2025-07-09
feat: generate settings page from schema

Nerivec:
Generate settings page from https://github.com/Koenkk/zigbee2mqtt/blob/dev/lib/util/settings.schema.json
Should allow to remove all manually-entered setting details throughout the docs and instead link to the appropriate heading. (the removal can be done in separate PRs)

Note: the algorithm is not for the whole JSONSchema7 but mostly only the elements used by Z2M (i.e. might need future updates if Z2M introduces new settings with new elements). It was adapted from [the one used by WindFront](https://github.com/Nerivec/zigbee2mqtt-windfront/blob/main/src/components/json-schema/SettingsList.tsx).

_If nothing else, a single-page of settings should hopefully teach GPT & the likes to stop screwing up/inventing Z2M settings..._

Also:
- fix mismatching package versions
  - quasar creates warning due to sass `@import` being deprecated, new syntax apparently won't be supported until [next quasar major](https://github.com/quasarframework/quasar/issues/9202)
- fix dark mode export changes
  - fixes some elements with wrong color
- fix some typing

### DIFF
--- a/docs/guide/configuration/all-settings.md
+++ b/docs/guide/configuration/all-settings.md
@@ -16,32 +16,31 @@ This page contains all currently supported settings in `configuration.yaml`.
 The code blocks show an **example value** for each setting (may be the default or any value derived from the possibilities/examples/boundaries).
 :::
 
-
 ## advanced
 
 Advanced
 
-### log_rotation 
+### log_rotation
 
 ```yaml
 advanced:
     log_rotation: true
 ```
 
-- Type: `boolean` 
+- Type: `boolean`
 - Default: `true`
 
-### log_console_json 
+### log_console_json
 
 ```yaml
 advanced:
     log_console_json: false
 ```
 
-- Type: `boolean` 
+- Type: `boolean`
 - Default: `false`
 
-### log_symlink_current 
+### log_symlink_current
 
 Create symlink to current logs in the log directory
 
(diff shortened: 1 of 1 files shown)


docgen/generate_settings.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    return nestedElements.length > 0 ? elements.concat(nestedElements) : elements;
+};
+
+export default async function generate_settings() {
+    console.log(`Generating settings`);
+
+    const schemaRsp = await fetch('https://github.com/Koenkk/zigbee2mqtt/raw/refs/heads/master/lib/util/settings.schema.json');
```
Koenkk:
```suggestion
    const schemaRsp = await fetch('https://github.com/Koenkk/zigbee2mqtt/raw/refs/heads/dev/lib/util/settings.schema.json');
```

(given I always run docgen before create a new release)

navbar.ts
```diff
@@ -24,6 +24,11 @@ export const navbar: NavbarConfig = [
                 link: '/supported-devices/',
             },
             '/guide/installation/',
+            {
```
Koenkk: I think it's better to move this under the Configuration section and call it "All settings" (having both Configuration and Settings is a bit confusing)

package.json
```diff
@@ -40,7 +41,7 @@
         "prettier": "3.6.2",
         "pug": "3.0.3",
         "pug-plain-loader": "1.1.0",
-        "sass": "1.77.6",
+        "sass": "1.89.2",
```
Koenkk, line 46: I think this causes this error (`pnpm run dev`)

Image

docgen/generate_settings.ts
```diff
@@ -0,0 +1,124 @@
+import type {JSONSchema7} from 'json-schema';
```
Koenkk, line 1: Maybe prefix the setting with it's parent? e.g. instead of `log_rotation` have `advanced` -> `log_rotation`, otherwise it's not immediately clear where the option should be put.

package.json
Nerivec: Yes, but the vuepress dependencies are mismatching otherwise. It's just a warning anyway, should be able to get rid of them once quasar is updated. Doesn't prevent anything from working, the scss vuepress config only ignores warning from deps, not local code, so these ones don't get silenced.

docgen/generate_settings.ts
Nerivec: I think the new code blocks should provide a better alternative?
_with https://github.com/Koenkk/zigbee2mqtt/pull/27961 to fix mismatching defaults & co_

navbar.ts
Nerivec: I guess it depends what you want to do with the big chunks of manually-maintained settings in `configuration/*` pages.
I think it would be better to link to the new generated page section in their place. Removes a lot of chores.
The `configuration/*` pages would then have "extra details" only.
Though not sure how to best arrange the menus either way.

Koenkk: The current docs provide more context about the settings, therefore I think its better to keep that as the main thing.

Nerivec: For sure should keep the more advanced details. But not quite sure what's the best way to organize all this.
For now I've swapped the location to under `Configuration`.

---

I'm thinking in a future PR, we can get rid of blocks like these:
https://github.com/Koenkk/zigbee2mqtt.io/blob/8b05e63c1af8e0b2b640fa9df2b2cc8749e74a26/docs/guide/configuration/logging.md?plain=1#L11-L49
https://github.com/Koenkk/zigbee2mqtt.io/blob/8b05e63c1af8e0b2b640fa9df2b2cc8749e74a26/docs/guide/configuration/frontend.md?plain=1#L18-L41

and instead just link to the appropriate `#` in the generated page. Lots of manual maintenance removed.

Koenkk: Agree, if they contain the same info they can be removed

docgen/generate_settings.ts
Koenkk: Looks really nice now!

Koenkk (approved):

Nerivec:
Should merge https://github.com/Koenkk/zigbee2mqtt/pull/27961 first. Then re-gen the docs here (& lint), and merge.

Koenkk:
thanks!

PR #3929, opened 2025-07-11
fix: add slzb06mg26 in EmberZNet adapters page

Nerivec:
Added in Recommended, no reason to think the newer ZigBee chip would be a problem, rest of hw is similar to mg24. Can re-evaluate in the future as needed.

_Didn't add a new image, would be the same as mg24, so, reused that one._


Koenkk:
thanks!

PR #3930, opened 2025-07-11
Fix broken OSRAM device pairing links

jpaju:
## Summary
- Fixed broken Belkin force.com URLs that were returning 404 errors
- Updated 13 OSRAM LIGHTIFY device documentation files with current Belkin support article URLs


Koenkk:
thanks!

PR #3936, opened 2025-07-15
Update groups guide to reflect 2.0.0 change

al4:
In verison 2.0.0, the ability to configure groups by configuration.yaml was removed.

See discussion at https://github.com/Koenkk/zigbee2mqtt/discussions/24198

> Removed configuring group members through configuration.yaml (groups.devices setting). This will not impact current group members; however, you will no longer be able to add or remove devices from a group through the configuration.yaml.

To save others wasting an hour of their life as I did, trying to get groups changes in configuration.yaml to work, let's update the guide to reflect this change.

### DIFF
--- a/docs/guide/usage/groups.md
+++ b/docs/guide/usage/groups.md
@@ -11,11 +11,11 @@ Groups are much more efficient than controlling devices separately as it signifi
 
 ## Creating a group
 
-Groups can be created via the frontend (easiest), [MQTT](./mqtt_topics_and_messages.md#zigbee2mqttbridgerequestgroupadd) or by adding them to the `configuration.yaml` as shown below.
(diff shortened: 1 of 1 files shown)


Nerivec:
It's group **members** that are no longer configurable through yaml, not the whole `groups`.

ref: https://github.com/Koenkk/zigbee2mqtt.io/pull/3104/file

al4:
I see. This came about because I was trying to set `optimistic: false` for a pre-existing group. The configuration was removed on restart, but it persisted after I set it via mqtt.

So if I was creating a new group, and set those configuration options on that group at the same time, it would work?

In that case, a note stating that group configuration (excluding members) cannot be updated after creation would be more appropriate - will update on confirmation.

(But if you are supposed to be able to update non-member configuration post-creation, it sounds as though I encountered a bug?)

Nerivec:
Were you editing the config file while Z2M was running?

PR #3964, opened 2025-07-22
fix: rename adapter sections

Nerivec:
Also removed a few zstack "specific chips" mentions since there are more around now.


Koenkk:
Thanks!

PR #3976, opened 2025-07-27
Add Hue Play wall washer image

Spirielle:
Added Hue Play wall washer image. 
File name is the UPC of the black version.


Koenkk:
thanks!

PR #3980, opened 2025-07-29
Add min and max brightness at qa devices

bernardesarthur:
I've tested that devices and they exposes that options: `min_brightness` and `max_brightness`


Koenkk:
Thanks for the pr, but this is generated because on the device definition (in zigbee-herdsman-converters), can you update that instead?

bernardesarthur:
I'll test in my env before to update that files and I'll open a new PR in that repo. Thanks.

PR #3984, opened 2025-07-30
Add 929003736501 for Philips Hue Datura Light

lordratner:
2 Commits by mistake, the first doesn't add anything. 

Adding the large version of the Hue Datura Light


Koenkk:
Thanks!

PR #3996, opened 2025-08-01
Add image for Mec Driver module

phihos:
Hi,

this PR adds an image for `Mec Driver module` from `L&S Germany`. I will create a corresponding PR in https://github.com/Koenkk/zigbee-herdsman-converters soon.


Koenkk:
Can you use the "Item-no" from the picture as name here? (and use it as `model` in the zigbee-herdsman-converters repo)

Bjk8kds:
~~@phihos Can you revise the file name to match the device?
In this case, it's 7061427.png~~

PR #4000, opened 2025-08-03
add note about defective capacitor causing Zigbee plug control issues

unite42:
Add troubleshooting section suggesting users check for faulty capacitors if sockets behave erratically or Zigbee2MQTT reports delivery failures when switching socket states. Include example error message for reference.


Koenkk:
Thanks!

PR #4004, opened 2025-08-04
Fix image of SLZB-07P10

FingerlessGlov3s:
Show correct image for the device SLZB-07P10 to match https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/images/SLZB-07P10.png


Koenkk:
Thanks!

PR #4009, opened 2025-08-05
Add BSEED Wall Socket Image for #28157

MilesTEG1:
Add image file for #28157


raveit65:
Hmm,
your picture is the french version with a pin, which has model no. FK86ZFRSK1W
EU Version with model no. FK86ZEUSK1W doesn't have a pin.

MilesTEG1:
Ho crap!
Can I change the image where I already push the PR?

raveit65:
Yes, 
- when in your branch
- replace the image in your local git folder
- commit with `git commit -a -m "your commit text"`
- `git push`

Edit: I guess reviewer will already squash both commits into one when merging.

MilesTEG1:
@raveit65 It should be now the right EU socket.
Thanks for pointing me my mistake.

Koenkk:
Thanks!

PR #4012, opened 2025-08-06
Update TS0601_water_valve.md

badr-ctrl:
added pairing guide for the valve since its not impossible to fugre out without the manual from ZPMETER


badr-ctrl:
id like to also add that the Brand that manufactures these is ZP Meter, hope the 213E version can be integrated soon

Koenkk:
thanks!

PR #4018, opened 2025-08-07
Update COZIGPMS.md, this device does not report battery state.

jolangse:


Koenkk:
Thanks for the PR, I updated the code to remove the `battery` expose for this device. Docs will be updated after the next release.

PR #4022, opened 2025-08-09
created 929003809701 which is just a new version of 929003045201

randriksen:
added a new version of philips hue centura white


Koenkk:
Thanks! Ignore my comment on zigbee-herdsman-converters.

PR #4025, opened 2025-08-11
Rename SFLZ02 Moes

bernardesarthur:
Correct as https://github.com/Koenkk/zigbee-herdsman-converters/pull/9745 did.


bernardesarthur:
I’ve checked https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/moes.ts and didn’t find that model SFL02-Z.

Koenkk:
Can you keep the old images? Otherwise the CI pipeline fails.

bernardesarthur:
Sure! Should I remove after next release?

Koenkk:
I would keep it as is, otherwise users on older releases would lose the pictures. Thanks!

PR #4029, opened 2025-08-12
Update home_assistant.md

ferbar:
Controlling Zigbee2MQTT via Home Assistant -> device list -> remove duplicate devices (breaks input_select) and sort options


Koenkk:
Thanks!

PR #4038, opened 2025-08-14
Add public/images/devices/ZC-W1.png

jspiros:
I copied IH-K009 as they are identical, though I have since encountered other form-factors that identify identically, [as noted in the original issue](https://github.com/Koenkk/zigbee2mqtt/issues/28167#issuecomment-3165783512).


Koenkk:
Thanks!

PR #4040, opened 2025-08-14
Update E1766.md

pilot1981:
change of real version of IKEA TRADFRI; added also specif action to wake up remote control and be sure to receive binding command


Koenkk:
Thanks!

PR #4046, opened 2025-08-17
Added image for Hive OTR1 single channel heating receiver.

absent42:
Added image for Hive OTR1 single channel heating receiver.


Koenkk:
Thanks! Ignore my comment on the zigbee-herdsman-converters pr.

PR #4058, opened 2025-08-21
Update Shelly 2PM docs to include both switch and cover mode

jfmlima:


jfmlima:
I guess everything is autogenerated and only notes can be updated, will update notes after https://github.com/Koenkk/zigbee-herdsman-converters/pull/9834 is merged

PR #4059, opened 2025-08-22
 Adds Nova Digital ZTS-MM device image

Ryaryu:
Adding device image as requested in https://github.com/Koenkk/zigbee-herdsman-converters/pull/9833#issuecomment-3214547329


Koenkk:
thanks!

PR #4061, opened 2025-08-23
Fix position of embedded images

Barokai:
On some lower resolutions, the images have some words in the same line (left of the image).
Moving the images to their own paragraph avoids this.

Screenshot of issue:
image


Koenkk:
Thanks!

PR #4062, opened 2025-08-23
Update zboss.md for ESP32-C6 firmware

diepeterpan:
Fix ESP32-C6 image and add information on more recent build with stability fixes.


Koenkk:
thanks!

PR #4067, opened 2025-08-25
Add T2 E26 device image

ds-sebastian:
Added image for the T2_E26 version which is nearly identical to the E27 version


Koenkk:
Thanks!

PR #4097, opened 2025-09-02
Added picture for mill gen 4 panel heaters

rurudev:


Koenkk:
Make sure that:
- The background is transparent (use e.g. [Adobe remove background](https://new.express.adobe.com/tools/remove-background))

Koenkk:
Thanks!

PR #4098, opened 2025-09-02
Update BSIR-EZ.md

nachonam:
Set switch 8 to OFF, otherwise the device left the network before adding to HA


nachonam:
Tested with RBSH-OS-ZB-EU model

Koenkk:
Thanks!

PR #4100, opened 2025-09-02
Add Sonoff plus vendor flashing instructions using web

KodeZ:
Added a new vendor flashing option for the Sonoff plus dongle. This was recommended on their website as the fastest option, and for me it surely was a seamless experience.


Koenkk:
Thanks!

PR #4106, opened 2025-09-04
fix: add cluster/cmd names in some docs

Nerivec:
Fixes https://github.com/Nerivec/zigbee2mqtt-windfront/issues/222


Koenkk:
Thanks!

PR #4111, opened 2025-09-05
add more pairing information for Bosch BSIR-EZ

fthiery:
Took me a while to find the information of how to re-set the device to pairing mode, so i propose adding initial status info and re-pairing instructions


Koenkk:
Thanks!

PR #4129, opened 2025-09-09
AD-RGBWH3001

pyrodex:
Adding new bulb that is not identified but so far fully works. 

Purchased from Amazon.

https://a.co/d/gL1qKVv


Koenkk:
Could you make the image 500x500 and make sure it has a transparent background?

pyrodex:
> Could you make the image 500x500 and make sure it has a transparent background?

Is this better? If so I'll resubmit it. 
AD-RGBWH3001

Koenkk:
This looks stretched, I would keep the original image and change the canvas size.

PR #4138, opened 2025-09-10
Add Hue A60 E27

agalera:


Bjk8kds:
Please make the background transparent.
You could use the [Adobe Express Remove Background tool](https://new.express.adobe.com/tools/remove-background) to make the background transparent.

agalera:
good point, fixed!

Koenkk:
Thanks!

PR #4140, opened 2025-09-10
Update MOSZB-153.md

irekzielinski:
added reset instructions as per this page:
https://help.frient.com/en/frient/motion-sensor-2-pet-6bf9ad/t/clhyldpqn07jdks0vlpybp6cb

those are not immediately obvious, so worth adding


Koenkk:
Thanks!

PR #4145, opened 2025-09-12
Optimize large png (for no particular reason)

Bjk8kds:
Optimize large png (for no particular reason) above 380 KB to reduced size


Koenkk:
Thanks!

PR #4147, opened 2025-09-12
Add images for FZB-1 switch from Nova Digital

brodock:
from: https://www.novadigitalsmart.com.br/produtos/interruptor-tecla-fisica-4x2-fzb-1-2-3---w-b

(please note models goes from FZB-1 to FZB-3, the number means the number of switches)


Koenkk:
Thanks!

PR #4162, opened 2025-09-17
Add Philips LCA015 image

fffe:
Add Philips LCA015 image


fffe:
Device PR is here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/10039

This would need to be renamed if I have the model name wrong there.

PR #4173, opened 2025-09-20
Bosch Door/Window Contacts: Add picture for `BSEN-C2D`, use correct picture for `BSEN-C2` and use same angle for `BSEN-CV`

DerDreschner:
I plan to add the `BSEN-C2D` as a separate device as it has slight differences between the `BSEN-C2`. During the development, I realized that the `BSEN-C2` had the wrong picture (only the `BSEN-CV` has such a large side box).

This pull request adds the picture for the `BSEN-C2D`, uses the same picture for the `BSEN-C2` (as that has the small side magnet, too) and uses the same picture angle for the `BSEN-CV` for uniformity.


Koenkk:
thanks!

PR #4178, opened 2025-09-22
Fix picture for Tuya TS0011 water leak valve

giDai7ja:
Updated the device image for the Tuya TS0011 water leak protection automatic ball valve to match the actual product.


giDai7ja:
Closing this PR as the TS0011 model number is correctly used for both switches and valves. The image does not require a fix.

PR #4179, opened 2025-09-22
Bosch Smart Plugs Compact: Add BSP-FD (new `[+M]` variant) and use same angle for all compact plugs

DerDreschner:
I will open a pull request for the new Matter-compatible variant of the Bosch Smart Plug Compact (known as `BSP-FD`). This pull request adds the picture for the device and uses the same angle for all other smart plug compact models from Bosch.


Koenkk:
Thanks!

PR #4188, opened 2025-09-26
Reduce device image sizes by 65%

Humberd:
I wrote a custom script that went through all the device images and reduced their size.

Results:
Optimized 7537 images
238.0 MB -> 83.1 MB
Saved 154.9 MB (65.1%)

I think the savings on the bandwidth for millions of requests will definitely be worth it.


Humberd:
Please let me know if you would like me to retain or remove the script since it was only one-time use.

Koenkk:
Could you explain what this script does? We already have the [imgbot optimizer](https://github.com/Koenkk/zigbee2mqtt.io/pull/2983)

Humberd:
I noticed a lot of 512x512 images were really heavy 300-500kB. But their only purpose is only informative about which device is connected. They don't need to be very detailed. As a result they now weigh 100kB+.

This script takes all the images and reduces their file size without decreasing their resolution.
It does some quality optimization (80%).

After the images were optimized I noticed that a lot of them were 32-bit color. Now they are 8-bit colors: 
image

Bjk8kds:
This is actually pretty good. But _unfortunately_, I think the converted image is quite _rough_.
In comparison, the (manual) conversion from the TinyPNG website algorithm can be much smoother (and smaller) and more "similar" to the original image.
This is just my personal _opinion_. The decision is ultimately up to @Koenkk.

Humberd:
Could you show me an example of the quite rough image quality?
When I glanced through the images I didn't realy noticed any visual differences.

Of course if the "roughness" is too much we can decrease the optimization from 80% to 90%

Bjk8kds:
It's not really noticeable at first glance, but upon closer inspection, the difference becomes apparent, at least for me.
Here I'll compare it with TinyPNG's algorithm, which is smarter at selecting replacement colors so _the difference isn't visible to the human eye_.
I can only provide a brief example because your PR resources are very heavy to load. I can't search/compare more complex images, for example photos of the original product, etc.

_Also please forgive me, perhaps I'm being too picky._
<br>
I also wrote the **image source** and **file size** in the file name.
<details>
<summary>roughness/noisier</summary>
<br>
If you look at the PR result image it looks rougher in converting the color to 8-bit, this is a comparison of TinyPNG which is smarter but larger in size.

Left is converted image, right original image, top PR image, bottom Tinypng.
Cuplikan layar 2025-09-27 203040-vert
</details>

<details>
<summary>darker/losing lightness</summary>
<br>
If you notice the part I marked on the original image (bottom), the PR image loses lightness (?) or becomes darker, compare it with Tinypng which retains its original appearance, while still having a small size.

</details>
 

_Once again, I apologize if I'm being too picky._

PR #4195, opened 2025-09-28
add missing level_config property to HK-LN-DIM-A

mSimon98:


Koenkk:
This part of the doc is generated based on the `expose` from the zigbee-herdsman-converters repo, please update that instead.

PR #4209, opened 2025-10-02
added SLZB-MR4

gam-st:


docs/guide/adapters/emberznet.md
```diff
@@ -289,6 +289,20 @@ SLZB-MR3 is a compact multi-radio smart home adapter featuring CC2674P10, EFR32M
 
 :::
 
+::: details SMLIGHT SLZB-MR4 MultiRadio Zigbee+Matter-over-Thread Ethernet USB POE WiFi LAN adapter
```
Koenkk, line 292: Given the descriptions of the MR1 to 4 are very similar, please merge them all into one so we just have a single entry: `SMLIGHT SLZB-MR1, SLZB-MR2, SLZB-MR3, SLZB-MR4 MultiRadio Zigbee+Matter-over-Thread Ethernet USB POE WiFi LAN adapter`, for e.g. MR3 and MR2 even the video review is the same as MR-1

gam-st: Well, following this logic, we have to merge, for example, Sonoff dongle E and mg24, shall we?

gam-st: And for the user, it is easier to have discrete records (lines) rather than all-in-one. Also, product links, where-to-buy links are different. In the next commit, I can add screenshots and video reviews specifically to each model.

Koenkk: > Well, following this logic, we have to merge, for example, Sonoff dongle E and mg24, shall we?

I don't agree because they don't look a lot like each other. MR-1 till MR-4 seems to be the same except chipset. 
This section is becoming very verbose now with almost the same product, do we have MR-10 in a year from now?

Screenshot 2025-10-02 at 21 26 24

gam-st: you won't believe, but MR-10 already exists :) https://smlight.tech/product/slzbmrw10 :)

by the way, @Koenkk  is the list sorted alphabetically or by the date items were added? I remember we talked a while back about keeping it chronological. Is that still the case?

Koenkk: >  I remember we talked a while back about keeping it chronological. Is that still the case?

It's in date added (because we got to a point where having the old devices as first doesn't make sense). I propose to merge MR1-MR4 and MR1U  into one and put it on top (of the Hybrid section). Having so many, almost the same product listed on this page will confuse users. The hub can stay in a separate entry (clearly a separate product)

gam-st: @Koenkk 
Got it, clear.  
Then, i will merge all mr1-4 tomorrow and will push updated commit.  
There is also MR1U commit, which is basically completely new device from the hardware perspective comparing to MR1, although it shares the same enclosure. So it has to be a separate line. And, because we will have mr1u-mr2u-mr3u-mr4u, i will keep those MR1U-MR4U also combined.

Koenkk: > which is basically completely new device from the hardware perspective comparing to MR1, although it shares the same enclosure. 

Agree!

PR #4218, opened 2025-10-03
Added new IKEA remote

Pepijn-DB:


Koenkk:
Thanks, but we already have a picture for this device: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/E2201.png

PR #4219, opened 2025-10-03
Image for Hue Festavia globe outdoor string lights (14 meter / 20 bulbs)

tannyl:
This is for the 14 meter version. Same image as 7 meter version.


Koenkk:
thanks

PR #4241, opened 2025-10-06
add photo of an Somfy 1871215B plug

pierrehenrymuller:
There is a new device from Somfy.


pierrehenrymuller:
sorry wrong filename

PR #4244, opened 2025-10-06
new device Frient REXZB-111

pierrehenrymuller:
Yes is the same image as SIRZB-111 because this two items have same enclosure, but this one is only a repeater for zigbee.


Koenkk:
Thanks!

PR #4247, opened 2025-10-07
Update E1766.md

Syzygy36au:
Updated OTA update section for clarity. Prior version states to press "the button" when there are three, only two of which appear to work.

### DIFF
--- a/docs/devices/E1766.md
+++ b/docs/devices/E1766.md
@@ -54,7 +54,7 @@ The [binding](../guide/usage/binding.md) functionality of this remote varies per
 - 23.0.75 and greater: supports binding to devices only; Also in this case, as for every battery zigbee device that sleep to save battery, before to send bind command from Z2M you have to press a button on remote control to wake up it and be sure remote control receive Z2M binding command.
 
 ### Starting OTA updates
-Press either the open or close button right before hitting the update button in Z2MQTT, and again every second or two for about 5 seconds until upgrade begins.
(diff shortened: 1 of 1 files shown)


docs/devices/E1766.md
```diff
@@ -54,7 +54,7 @@ The [binding](../guide/usage/binding.md) functionality of this remote varies per
 - 23.0.75 and greater: supports binding to devices only; Also in this case, as for every battery zigbee device that sleep to save battery, before to send bind command from Z2M you have to press a button on remote control to wake up it and be sure remote control receive Z2M binding command.
 
 ### Starting OTA updates
-Press the button on it right before hitting the update button in Z2MA, and again every second or two for about 5 seconds until upgrade begin.
+Press either the open or close button right before hitting the update button in Z2MA, and again every second or two for about 5 seconds until upgrade begins.
```
marazmarci: By the way, what is Z2MA? Isn't it supposed to be just Z2M? Could you please correct it if you're already here?

Syzygy36au:
Hi there, can jump back in and do that, no problem.

Chris Sheed

docs/devices/E1766.md
Syzygy36au: Yep saw that, but didn't want to change anything I wasn't 100% sure of. I'll update to Z2MQTT if that works?

Koenkk:
Thanks!

docs/devices/E1766.md
marazmarci: I saw you changed it to Zigbee2MQTT, thanks!

PR #4825, opened 2026-02-11
Add files via upload

HuismAndre:
Here's a picture.


Koenkk:
I propose to keep `public/images/devices/TS0222.png` as is, it's a model used by many devices. For correct detection see: 

> Refer to the [documentation](https://www.zigbee2mqtt.io/advanced/support-new-devices/02_support_new_tuya_devices.html#fixing-tuya-device-detection) to fix device detection issues.

PR #4873, opened 2026-02-28
docs: add time sync workaround for HS3AQ.md

SirY4:
The Heiman HS3AQ (Smart Air Quality Monitor) does not automatically synchronize its clock with the standard Zigbee time cluster. It requires the seconds since midnight of the current day to be pushed manually to the `genTime` cluster, otherwise the time on the display drifts or stays incorrect.

I have added a documentation section with a Home Assistant automation workaround that stochastically/regularly updates the time via MQTT to ensure the device display shows the correct local time.

This fix was tested and confirmed working with the HS3AQ model.

### DIFF
--- a/docs/devices/HS3AQ.md
+++ b/docs/devices/HS3AQ.md
@@ -71,3 +71,29 @@ Value can be found in the published state on the `temperature` property.
 It's not possible to read (`/get`) or write (`/set`) this value.
 The unit of this value is `°C`.
 
+### Time synchronization
+
+This device does not sync time using the standard Zigbee protocol, but instead requires the seconds since midnight of the current day. To fix the asynchronous time, you can regularly publish this value to the `genTime` cluster. 
+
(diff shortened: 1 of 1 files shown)


Koenkk:
I think we can fix this with the `writeTimeDaily` extend, examples: https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters%20writeTimeDaily&type=code

SirY4:
That sounds like a much cleaner solution! I am not a developer myself yet, so I'm not sure how to create a PR for the converters repository. If you could add m.writeTimeDaily({endpointId: 1}) to the HS3AQ definition, I would be happy to test it once it's in the edge version.
THX

Koenkk:
Added!

> Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now and in the next release which is every 1st of the month.

PR #4983, opened 2026-04-03
docs: replace device image for SONOFF SWV-ZNE and SWV-ZNU

CubeZ2mDeveloper:
Device: SONOFF SWV-ZNE & SWV-ZNU

**Changes:**

- Use a more accurate image for the devices.


Koenkk:
Thanks!

PR #5021, opened 2026-04-15
Kingswindsor patch 1

kingswindsor:


Koenkk:
Please resize to 512x512 and rename to `VES-ZB-PIR-21`

PR #5041, opened 2026-04-23
Update Tuya 3.12.6 apk download link

andrei-lazarov:
The old link is broken. Update or remove it?


andrei-lazarov:
I could not get it working on the emulator

- Android Studio can not emulate ARM (anymore?). Only x86 virtual devices
- Tuya APKs are ARM only

It partially works on a real phone
- Can't open devices newer than the app
- Newer app versions reject the custom certificate

Needs a different approach.. probably root the gateway

Koenkk:
Thanks!

PR #5048, opened 2026-04-24
Fix italics for windfront markdown

andrei-lazarov:
Replaced `_italic_` occurences (found in devices folder) with `*italic*` so it displays properly in windfront


Koenkk:
Thanks!

PR #5051, opened 2026-04-27
Update connect_to_a_remote_adapter.md

landolfi-us:
Update to show that the UI should be used to make the HA Z2M config changes rather than editing configuration.yaml directly. Also added directions for changing adapter, which may or may not be necessary but worked for me.

### DIFF
--- a/docs/advanced/remote-adapter/connect_to_a_remote_adapter.md
+++ b/docs/advanced/remote-adapter/connect_to_a_remote_adapter.md
@@ -82,11 +82,17 @@ reboot
 
 ## 3. Configure
 
-Now edit the Zigbee2MQTT `configuration.yaml` accordingly, replace `192.168.2.13` with the IP or hostname of your system where the adapter is connected to.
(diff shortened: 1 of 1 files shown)


docs/advanced/remote-adapter/connect_to_a_remote_adapter.md
```diff
@@ -82,11 +82,17 @@ reboot
 
 ## 3. Configure
 
-Now edit the Zigbee2MQTT `configuration.yaml` accordingly, replace `192.168.2.13` with the IP or hostname of your system where the adapter is connected to.
-
-```yaml
-serial:
-    port: 'tcp://192.168.2.13:20108'
+Now edit the Zigbee2MQTT using the App Configuration tab. Do not edit configuration.yaml, as the app will overwrite your changes at startup. Replace `192.168.2.13` with the IP or hostname of your system where the adapter is connected to.
```
Koenkk, line 85: > Do not edit configuration.yaml

This is specific for HA, for bare metal you need to edit the `configuration.yaml`

PR #5067, opened 2026-04-29
feat: add "securing" page

Nerivec:
Tried to cover all relevant aspects I could think of while remaining on "overall" architecture and every-day use (this is not a paper on Zigbee security :grin:).
Kept duplication to minimum, used links to relevant sections, and avoided stuff that is far better left to specific software's own docs (too many to even count anyway).
I also added some mentions on common misconceptions around PAN/IEEE/etc..
And consolidated a couple of sparse docs on the topic in that new page.

Let me know if you see anything else that might be worth adding, or if you spot any mistakes :grimacing: 

@burmistrzak @andrei-lazarov @sjorge 

@Koenkk thinking something like https://ecosystem.vuejs.press/plugins/tools/redirect.html might be useful to avoid keeping entire redirect pages? (it currently creates weird dupes when it's stuff that was in the sidebar - it remains there)

Code wise, it would be good to gate external JS extensions ([1](https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/controller.ts#L67), [2](https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/controller.ts#L77)) behind an option instead of automatically enabling them, since they are hugely permissive by design. This would "no-op" the related MQTT topics. Although until 3.0 it would have to be "enabled by default" to avoid a breaking change, unless we make an exception....

### DIFF
--- a/docs/guide/installation/14_securing.md
+++ b/docs/guide/installation/14_securing.md
@@ -150,7 +150,7 @@ The Network Address is randomly assigned on device join and usually remains the 
 The "permit join" state determines whether new devices are allowed to join the network.
 Joining is enabled temporarily (for 254 seconds by default) via the dedicated frontend button or via MQTT. You can also close the joining window manually once pairing is complete.
 
-:::warning WARNING
+:::tip TIP
 Freshly joined devices may automatically permit joining on themselves for a specific duration (max 254 seconds).
 :::
 
@@ -169,11 +169,12 @@ See also [Add install code via MQTT](../../guide/usage/mqtt_topics_and_messages.
 For stricter control over which devices are allowed on the network, use a passlist or blocklist.
 See [Device blocklist / passlist](../configuration/block-pass-list.md) for more details.
 
-:::tip
+:::tip TIP
 Devices that are not allowed are removed from the network on startup (e.g. configuration changes since last run), and on join attempts.
+Note: removal is a request sent to the targeted device to "ask it" to leave, a malicious device could purposely ignore it.
 :::
 
-:::tip
+:::tip TIP
 Using a passlist is the most restrictive and therefore most secure option, only explicitly trusted devices can join.
 :::
 
--- a/docs/guide/installation/14_securing.md
+++ b/docs/guide/installation/14_securing.md
@@ -177,6 +177,13 @@ Devices that are not allowed are removed from the network on startup (e.g. confi
 Using a passlist is the most restrictive and therefore most secure option, only explicitly trusted devices can join.
 :::
 
+### Inter-PAN
+
+Inter-PAN messages are **unsecured messages** sent to or received from unjoined devices 1-hop away.
+Touchlink (previously known as ZLL) uses inter-PAN messaging.
+
+Inter-PAN is usually reserved for highly specific operations (e.g. resetting a device to factory settings), and undesired messages are aggressively dropped; this limits the impact of its lack of security.
+
 ### Zigbee 4.0
 
 Zigbee 4.0 provides several security enhancements.
--- a/docs/guide/installation/14_securing.md
+++ b/docs/guide/installation/14_securing.md
@@ -183,7 +183,12 @@ Using a passlist is the most restrictive and therefore most secure option, only 
 Inter-PAN messages are **unsecured messages** sent to or received from unjoined devices 1-hop away.
 Touchlink (previously known as ZLL) uses inter-PAN messaging.
 
-Inter-PAN is usually reserved for highly specific operations (e.g. resetting a device to factory settings), and undesired messages are aggressively dropped; this limits the impact of its lack of security.
+Inter-PAN is usually reserved for highly specific operations (e.g. resetting a device to factory settings via Touchlink), undesired messages are aggressively dropped, and close physical proximity is required.
+This limits the impact of its lack of security.
+
+:::warning WARNING
+Avoid devices that keep Touchlink permanently enabled, especially in places with relative ease of access; a malicious user could otherwise disrupt the network.
+:::
 
 ### Zigbee 4.0
 
--- a/docs/guide/installation/14_securing.md
+++ b/docs/guide/installation/14_securing.md
@@ -56,7 +56,7 @@ Refer to the available documentation and guides for your broker.
 :::
 
 :::caution CAUTION
-Using `reject_unauthorized: false` in production is dangerous. Its disables TLS certificate validation and makes the connection vulnerable to man-in-the-middle attacks.
+Using `reject_unauthorized: false` in production is dangerous. Its disables TLS certificate validation and makes the connection vulnerable.
 :::
 
 ## Frontend
@@ -154,6 +154,16 @@ Joining is enabled temporarily (for 254 seconds by default) via the dedicated fr
 Freshly joined devices may automatically permit joining on themselves for a specific duration (max 254 seconds).
 :::
 
+#### Install codes
+
+Joining with an install code provides better security (if available).
+A code is randomly assigned during the manufacturing process of a device.
+The code allows to encrypt the initial network key transport from the Trust Center (coordinator) to the joining device.
+
+Ask the vendor or refer to each device's documentation to check if an install code is available (usually printed on the device, plainly or as a QR code).
+Codes can be added through the frontend.
+See also [Add install code via MQTT](../../guide/usage/mqtt_topics_and_messages.md#zigbee2mqttbridgerequestinstall_codeadd).
+
 ### Device passlist and blocklist
 
 For stricter control over which devices are allowed on the network, use a passlist or blocklist.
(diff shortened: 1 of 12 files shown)


docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,193 @@
...
+Zigbee2MQTT will replace `GENERATE` with a randomly generated key on startup.
+Of course, you can also set a specific key manually.
+
+See [Zigbee network configuration](../configuration/zigbee-network.md) for more details.
+
+### PAN ID and Extended PAN ID
+
+The PAN ID and Extended PAN ID identify your network. These are not security measures.
+
+Changing them is possible and require re-pairing all devices.
```
andrei-lazarov:
```suggestion
Changing them is possible and requires re-pairing all devices.
```

docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,193 @@
...
+
+The IEEE address is statically assigned to the Zigbee chip in the device (although it can be changed in some cases).
+Two devices with the same IEEE address cannot join a same network.
+
+The Network Address is randomly assigned on device join and usually remains the same until reset/rejoin (although it may change at the discretion of the device).
+
+### Joining (permit join)
+
+The "permit join" state determines whether new devices are allowed to join the network.
+Joining is enabled temporarily (for 254 seconds by default) via the dedicated frontend button or via MQTT. You can also close the joining window manually once pairing is complete.
```
andrei-lazarov, line 151: Maybe add one more warning here, something like:

> Pairing a device (without an install code) invloves emitting the network key - potentially allowing bad actors to decrypt all network traffic

docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,193 @@
...
+## Firmware updates (OTA)
+
+Firmware updates can change device behavior in ways that may affect Zigbee2MQTT compatibility, and could in theory introduce malicious or buggy functionality.
+
+Zigbee2MQTT retrieves OTA images from the [Koenkk/zigbee-OTA](https://github.com/Koenkk/zigbee-OTA) repository by default.
+
+:::caution CAUTION
+Only use firmware from trusted sources.
+Avoid using custom OTA index URLs unless you fully trust the source.
+:::
```
andrei-lazarov:
```suggestion
By default, Zigbee2MQTT retrieves OTA images from the [Koenkk/zigbee-OTA](https://github.com/Koenkk/zigbee-OTA) repository, which mirrors unaltered files coming from the device manufacturers.

Firmware updates are generally recommended to keep the network secure, but they can change device behavior in ways that may affect Zigbee2MQTT compatibility, and could in theory introduce malicious or buggy functionality.

:::caution CAUTION
Devices usually reject unofficial updates. Still, only use firmware from trusted sources.  
Avoid using custom OTA index URLs, unless you fully trust the source.
:::
```

andrei-lazarov (commented):
I think a section about Touchlink is also needed. That might be the least secure part?

docs/guide/installation/20_zigbee2mqtt-fails-to-start.md
```diff
@@ -1,4 +1,5 @@
 ---
+title: Zigbee2MQTT fails to start
```
andrei-lazarov, line 2: Whoops. Didn't notice it duplicated the sidebar entry.
(what if you use the same title, still creates a duplicate?)

You can remove the old page if you want.

docs/guide/installation/14_securing.md
burmistrzak: Yes, this is super important to highlight. Especially when pairing security devices.

**Edit:** We might need to make Install Codes more accessible. At the moment, it's quite buried and UX is poor because it requires copy&paste from a QR code reader.

docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,193 @@
...
+
+The Network Address is randomly assigned on device join and usually remains the same until reset/rejoin (although it may change at the discretion of the device).
+
+### Joining (permit join)
+
+The "permit join" state determines whether new devices are allowed to join the network.
+Joining is enabled temporarily (for 254 seconds by default) via the dedicated frontend button or via MQTT. You can also close the joining window manually once pairing is complete.
+
+:::warning WARNING
+Freshly joined devices may automatically permit joining on themselves for a specific duration (max 254 seconds).
```
burmistrzak, line 154: Do we know which devices do that?

docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,193 @@
+---
+sidebarDepth: 1
+next: ../configuration/
+---
+
+# Securing the installation
```
burmistrzak, line 6: Nitpick: Securing _your_ installation 😄

burmistrzak (commented):
Overall a solid first step and good summary of main attack vectors. 🙌

docs/guide/installation/14_securing.md
Nerivec: Silabs firmware do this to "extend the network" (180sec by default if memory serves). Not sure about others.

docs/guide/installation/20_zigbee2mqtt-fails-to-start.md
Nerivec: Should be fixed if we use the redirect plugin. Easier to maintain redirects in same location too, instead of empty files.
Out of scope for this PR though.

docs/guide/installation/14_securing.md
Nerivec: Was done on purpose. It's not necessarily "yours". I try to avoid "you" in docs unless there's no way around it, reads better that way.
_Takes the target off your back, so to speak_ :grin:

docs/guide/installation/14_securing.md
burmistrzak: Ok, so like 95% of devices on my network. 😅
I'll have to check that out sometime.

Does this "feature" also kick in when pairing via Install Code?

docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,217 @@
...
+### Encryption (HTTPS / WSS)
+
+Enable HTTPS/WSS directly within Zigbee2MQTT by providing a [certificate and private key](../configuration/frontend.md#advanced-configuration).
+
+Alternatively, terminate TLS at a reverse proxy (Nginx, Apache, etc.) placed in front of the frontend.
+See [Frontend configuration](../configuration/frontend.md) for proxy configuration examples.
+
+### Bind address
+
+By default, the frontend listens on all interfaces (`0.0.0.0`).
```
sjorge, line 87: I wonder if having it default to localhost and explicitly having to enable listening on all interfaces is a better. That might be considered breaking change I guess.

sjorge (commented):
Over all this looks pretty good to me, my only comment isn't really about the content 😅

docs/guide/installation/14_securing.md
burmistrzak: Yep, listening on `localhost` only would require a reverse proxy or SSH forwarding to access the frontend.

Koenkk: I think most people want `0.0.0.0` so I think it's better to keep that.

docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,217 @@
...
+Codes can be added through the frontend.
+See also [Add install code via MQTT](../../guide/usage/mqtt_topics_and_messages.md#zigbee2mqttbridgerequestinstall_codeadd).
+
+### Device passlist and blocklist
+
+For stricter control over which devices are allowed on the network, use a passlist or blocklist.
+See [Device blocklist / passlist](../configuration/block-pass-list.md) for more details.
+
+:::tip
+Devices that are not allowed are removed from the network on startup (e.g. configuration changes since last run), and on join attempts.
```
Koenkk, line 173: I think it's good to mention that it does not force the device to be removed, it's up to the device to honour this request. It can also always re-join the network temporarily.

Koenkk:
> @Koenkk thinking something like https://ecosystem.vuejs.press/plugins/tools/redirect.html might be useful to avoid keeping entire redirect pages? (it currently creates weird dupes when it's stuff that was in the sidebar - it remains there)

Agree! It's getting a bit messy

docs/guide/installation/14_securing.md
Nerivec: If I'm not mistaken, that's part of the BDB steering, so, this is expected behavior, per spec.
You can always close it by terminating the permit join early though (since that's broadcasted).

burmistrzak: Ah, I see!
Maybe we should put that into the docs as well? 👀

docs/guide/installation/14_securing.md
```diff
@@ -0,0 +1,218 @@
...
+:::tip TIP
+Using a passlist is the most restrictive and therefore most secure option, only explicitly trusted devices can join.
+:::
+
+### Inter-PAN
+
+Inter-PAN messages are **unsecured messages** sent to or received from unjoined devices 1-hop away.
+Touchlink (previously known as ZLL) uses inter-PAN messaging.
+
+Inter-PAN is usually reserved for highly specific operations (e.g. resetting a device to factory settings), and undesired messages are aggressively dropped; this limits the impact of its lack of security.
```
andrei-lazarov: I was thinking about how Touchlink acts as backdoor to Zigbee devices.

At least IKEA and Tuya keep it enabled non-stop afaik - letting anyone easily hijack the devices off a network. You can just factory reset them any time.

> [Touchlink]-capable devices accept unsecured inter-PAN messages for various use-cases.
> The only condition is a very short physical distance. If bad actors bypass the proximity test, they can gain control of said devices.
>
> For extra security, devices should be listening for inter-PAN messages only a short duration (after power-on).
> **Avoid devices that keep Touchlink enabled indefinitely.**

Someone linked this fun paper a while back 🙂 
https://ieeexplore.ieee.org/document/7958578

burmistrzak: @andrei-lazarov Aye yo! That's no good...
At least Hue turns off Touchlink after 30 minutes... Phew!

Koenkk:
Thanks!

PR #5070, opened 2026-04-30
Expand documentation for Cigol Connect device

kimtnielsen:
Added detailed notes on the Cigol Connect device, including its overview, endpoint structure, input behavior, direct Zigbee binding, output control, switch action configuration, interaction between input modes and bindings, binding considerations, and important notes.


Koenkk:
Thanks!

PR #5076, opened 2026-05-02
Added Tuya TZE204_apiu8k13 image

ohadvano:
Herdsman converter [PR #12089](https://github.com/Koenkk/zigbee-herdsman-converters/pull/12089)


andrei-lazarov (changes requested):
Name is not correct. Should pe `TS0601_power_monitoring_switch.png`
(if the converter is accepted like this)
```ts
fingerprint: [{modelID: "TS0601", manufacturerName: "_TZE204_apiu8k13"}],
model: "TS0601_power_monitoring_switch",
```

Similar logic for other devices. 

Also can you add all pictures in a single PR?

ohadvano:
Ok so I'll create a single new PR with corrected names once the converter is finalized

PR #5089, opened 2026-05-04
Add device images for Novato ZIS-03 and ZIS-04

bilgi-source:
Adds 512x512 PNG device images for Novato ZIS-03 and ZIS-04 24 GHz radar presence sensors.

Related to Koenkk/zigbee-herdsman-converters#12099


Koenkk:
Thanks!

PR #5093, opened 2026-05-05
Update E2490.md with direct binding limitations

jacauc:
Clarify limitations of direct binding with Bilresa remotes and emphasize the need for reporting setup.


Koenkk:
Thanks!

andrei-lazarov:
Thanks for the feedback on this document!
Though I've already written the limitation 3 times on the page. I don't think a 4th time is necessary.
Can you emphasize/reword these instead of writing the new paragraph?

> Partial and conflicting binding support: All remotes control the same devices. See Binding

> (with partial binding support: all BILRESA remotes will control the same 1-3 groups)

> All remotes directly control groups 21658, 21659, 21660 (channels 1-3).

PR #5102, opened 2026-05-09
Update Legrand 412170: add note about device_mode auto producing erratic behavior

alexhk:
Adds a notes section to the 412170 device page warning users that the factory-default `device_mode` produces erratic relay behavior with the L to C1 push-button wiring shown in Legrand's wiring instructions, and that setting `device_mode` to `switch` is required for predictable operation.

Goes with Koenkk/zigbee-herdsman-converters#12141 which updates the inline `withDescription` text in the converter. Background info including info on Wireshark capture against the official Legrand Gateway, Home+Control app comparison with the 412171, etc. are in the linked PR.

### DIFF
--- a/docs/devices/412170.md
+++ b/docs/devices/412170.md
@@ -27,7 +27,7 @@ pageClass: device-page
 
 ## Notes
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #5103, opened 2026-05-09
Add correct image for Mumubiz CZV20 water valve

nzlamb:
Replaces existing incorrect image showing a smart plug.

Image is formatted 512px square PNG with transparent background.


Koenkk:
thanks!

PR #5108, opened 2026-05-10
Fix brand reporting table

andrei-lazarov:
IDK about this table, kinda out-of-place... not all devices follow the "brand rule"..
Maybe we can have a Reporting doc page and move it there.

- I can confirm reporting fails for old Hue bulbs (even on 2024 firmware) with table_full / unsupported_manuf_general_command.
- I can confirm reporting works for new Hue bulbs, but not color temp.
- All reporting works on my IKEA bulbs (latest tradfri and kajplats)..
  - **changed color temp to Yes** 
- Changing brightness while bulb is off is not specific to IKEA.. should work on most bulbs
  - **removed**

cc @burmistrzak

### DIFF
--- a/docs/guide/usage/binding.md
+++ b/docs/guide/usage/binding.md
@@ -78,19 +78,18 @@ When a devices is being bound to, Zigbee2MQTT will automatically configure repor
 
 In order for this feature to work, the device has to support it. As devices from the same manufacturer (mostly) have the same features the table below might help to find out if your device supports it.
 
-| Brand            | On/Off | Brightness | Color | Color temperature | Color Mode |
-| :--------------- | :----: | :--------: | :---: | :---------------: | :--------: |
-| Philips Hue      |  N(1)  |    N(2)    |   N   |         N         |     N      |
-| Philips Hue (BT) |   Y    |     Y      |   Y   |         Y         |     N      |
-| Trådfri(3)       |   Y    |     Y      |   Y   |         N         |     Y      |
-| Innr             |   Y    |     Y      |   Y   |         Y         |     Y      |
-| GLEDOPTO         |   N    |     N      |   N   |         N         |     N      |
-| OSRAM            |   Y    |     Y      |   N   |         N         |     Y      |
-| Müller Licht     |   N    |     N      |   N   |         N         |     Y      |
+| Brand               | On/Off | Brightness | Color | Color temperature | Color mode |
+| :------------------ | :----: | :--------: | :---: | :---------------: | :--------: |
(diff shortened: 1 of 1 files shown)


burmistrzak:
We should probably add a paragraph about Hue fixtures (Amaze, Being, etc.) and Hue Connect modules?

Because everything that includes an old Hue Engine (green PCB, `010E` hardware platform) is ZLL-only and doesn't support reporting either.

andrei-lazarov:
I think that's way too specific for the binding doc.
Don't they fit the short description (ZB without BT)?

I think the Philips notes, that get applied to all Hue devices, is a better place.
Maybe later we can somehow draw the line and apply it only to affected devices

burmistrzak:
> I think that's way too specific for the binding doc. Don't they fit the short description (ZB without BT)?

Well, here's the thing: Some [fixtures](https://www.zigbee2mqtt.io/devices/1742930P7.html) with the new Hue Connect module on board do support reporting but have no BLE!
So a missing Bluetooth symbol on the box doesn't necessarily mean it's an older revision without full Zigbee 3.0 certification.

> I think the Philips notes, that get applied to all Hue devices, is a better place. Maybe later we can somehow draw the line and apply it only to affected devices

Fair point. 😊

At some point, I'd like to enrich the Z2M catalog with more metadata/labels (certified/uncertified, ZLL/ZB3.0, ZGP proxy/sink, etc.) to make it easier to find devices that support specific set of standards.

docs/guide/usage/binding.md
```diff
@@ -78,19 +78,19 @@ When a devices is being bound to, Zigbee2MQTT will automatically configure repor
...
+| IKEA               |   Y    |     Y      |   Y   |         Y         |     Y      |
+| Innr               |   Y    |     Y      |   Y   |         Y         |     Y      |
+| GLEDOPTO           |   N    |     N      |   N   |         N         |     N      |
+| OSRAM              |   Y    |     Y      |   N   |         N         |     Y      |
+| Müller Licht       |   N    |     N      |   N   |         N         |     Y      |
 
 1. Bulbs on old firmware (date 20170908 or older) do report On/Off
 2. Zigbee2MQTT will manual poll for change if a binding updates the bulb.
-3. The color/brightness of a Trådfri bulb can be changed while the state=off, it also reports back the change.
+3. Devices manufactured after ~2019
```
burmistrzak: AFAIK, the first BLE-enabled bulbs were released in late 2019 and the first Zigbee 3.0 Hue Connect module was certified in late 2018.

I'd phrase it something like

> Lamps & bulbs released starting around 2019

Koenkk:
thanks!

PR #5109, opened 2026-05-10
Pairing information from manual for 4512770.md

jskarpe:
Include information from manual on how to join the device into the Zigbee mesh


andrei-lazarov (changes requested):
Thanks! Please add `## Notes` above, and add one more `#` to all subtitles

Koenkk:
thanks

PR #5112, opened 2026-05-11
add: ZT08 device image

0leg7:
Companion to Koenkk/zigbee-herdsman-converters#12180 which renames the `_TZE284_hodyryli` definition from `ZY-ZTH03PRO` (a guess from #12004 that doesn't match any real product) to `ZT08` — the actual marking on the case — and adds LCD clock support.

The existing `ZY-ZTH03PRO.png` pictures a physically different device (rectangular, OUT/IN, no clock) and is left in place — it remains the correct image for whatever future definition uses that model name.


Koenkk:
could you make the background transparent?

0leg7:
Done — background made transparent (rembg/u2net), 512×512 PNG. Updated in the same branch.

Koenkk:
thanks

PR #5119, opened 2026-05-14
Update DWZTCGQ11LM.md

stfpari:
This adds two practical setup notes based on my experience:
1. The device is shipped in Matter/Thread mode and needs to be switched to Zigbee before pairing with Zigbee2MQTT.
2. In door/window mode, calibration is required by pressing the button 3 times while the door/window is closed.


Koenkk:
thanks!

PR #5120, opened 2026-05-14
Enhance documentation for power drop threshold

amaisano:
Added notes on instantaneous reporting and power drop threshold settings.

Out of the box, instant power doesn't really report (takes forever or doesn't change at all) - even which power cycling connected devices on the plug.

This was found by trial and error, and doesn't seem to be documented anywhere. I tried the normal Reporting Z2M settings first, but it had no effect. This did work, however.

image

### DIFF
--- a/docs/devices/3RSP02064Z.md
+++ b/docs/devices/3RSP02064Z.md
@@ -160,6 +160,11 @@ The unit of this value is `w`.
 ### Power drop threshold (numeric)
 Reports sudden power changes. Power rise and drop alerts can be enabled separately. Threshold adjustable..
 Value can be found in the published state on the `power_drop_threshold` property.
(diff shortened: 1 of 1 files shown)


docs/devices/3RSP02064Z.md
```diff
@@ -165,3 +165,7 @@ To write (`/set`) a value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/
 The minimal value is `1` and the maximum value is `3200`.
 The unit of this value is `w`.
 
+## Notes
```
Koenkk: Please move above "DO NOT EDIT BELOW THIS LINE"

Koenkk:
Thanks!

PR #5122, opened 2026-05-16
IKEA Matter -  warnings for unofficial support

andrei-lazarov:
Adds this to all IKEA devices with "Matter" in their description

> ## Unofficial support
> **This device is part of the IKEA Matter/Thread line-up.**  
> The hidden Zigbee mode is not officially supported by IKEA. It exists mainly to keep backwards compatibility via Touchlink.  
> As a result, the Zigbee firmware is stripped-down to the essentials, and could even be removed in OTA updates.

I made a new function, so this displays **above** the user-submitted notes


Koenkk:
thanks!

PR #5125, opened 2026-05-16
Update Legrand 067776A: Add info for pairing on older fw, calibration for venetian mode, tilt control in UI

alexhk:
Adds information
* for a pairing workaround needed for devices with older firmware
* info for the calibration procedure in mode venetian_bso (venetian blinds) 
* Info to restart z2m to show show tilt control in UI after changing mode


Koenkk:
thanks!

PR #5130, opened 2026-05-17
feat: add redirect plugin

Nerivec:
I think I caught them all :grin: 
Makes for a nice cleanup.

cc: @andrei-lazarov


Koenkk:
Cool, thanks!

PR #5134, opened 2026-05-18
Update notes for JY-GZ-01AQ version v21 issue

msnjdev:
Removed workaround and added fix details for version v21 issue.


Koenkk:
Thanks!

PR #5145, opened 2026-05-20
feat(device): add C-ZB-DM204V2 image

niracler:
Summary:
- replace the C-ZB-DM204V2 placeholder with the actual device image
- keep the existing C-ZB-DM204 image unchanged

Test:
- pnpm test


Koenkk:
Thanks!

PR #5148, opened 2026-05-21
Clarify baud rate update when changing adapters

barryshaffer:
Added note about updating baud rate when moving from an older adapter. (zbt-1 to zbt-2)


Koenkk:
Thanks!

PR #5161, opened 2026-05-25
S4EM-001PXCEU16: Add pairing & maintenance notes

xkr47:
Based on manual that came with the device + personal experience + wording from "Shelly S4SW-001P8EU".


Koenkk:
Thanks!

PR #5177, opened 2026-05-28
Update image link for HS1SA-E Lover device to match HS1SA-E

SmileyMan:
Updated firmware and now get a blank "No Image Available"

HS1SA-E-Lover is just a HS1SA-E with HS1SA_EF2_V15_20260407.ota firmware. Visually that are the exact same device.


Koenkk:
please update the image instead, this docs is re-generated on every release

SmileyMan:
ok

PR #5178, opened 2026-05-28
SM323_v2: update device image

fidusachates:
Replaces the existing `public/images/devices/SM323_v2.png` with an updated manufacturer-supplied product render.

- 512 × 512 PNG, white background
- Same filename and path – modifies the existing image rather than adding a new one


Koenkk:
Please make the background transparent.

fidusachates:
Thanks – updated.

Koenkk:
Thanks!

PR #5186, opened 2026-05-29
feat: add Novato ZSM-01 device image

bilgi-source:
Adds device picture for the Novato ZSM-01 curtain motor.

Related PR: Koenkk/zigbee-herdsman-converters#12330


Koenkk:
Thanks!

PR #5191, opened 2026-05-31
SM323_v2: sync description with herdsman-converters

fidusachates:
Syncs the auto-generated description in `docs/devices/SM323_v2.md` with the already-merged change in zigbee-herdsman-converters PR Koenkk/zigbee-herdsman-converters#12326 ("Zigbee retrofit dimmer 250W" → "Zigbee dimmer switch SM323").

The docs page is currently still showing the old description because the regenerated md hasn't been produced yet. This is the same line docgen will write on its next run, so the edit is just pre-empting that to keep the live page consistent with the merged converter source in the meantime.


Koenkk:
Thanks but this page will automatically updated when support for this lands in the z2m release.

PR #5193, opened 2026-05-31
SM323_v1: replace device image

fidusachates:
Replaces `public/images/devices/SM323_v1.png` with the correct manufacturer-supplied product render. The previous image was showing the wrong product.

- 512 × 512 PNG, transparent background
- Same filename and path – modifies the existing image rather than adding a new one

Note: `SM323_v2.png` also appears in the diff list, but the file is **byte-identical** to upstream master (no content change). It shows up only because of a small history divergence in the fork – the file content itself is unchanged.


Koenkk:
Thanks!

PR #5196, opened 2026-05-31
Document low light sensor sensitivity issue

sepich:
Added information about low light sensor sensitivity and its limitations.


Koenkk:
Thanks!

PR #5097, opened 2026-05-06
Allow whiteLabels of whiteLabels

andrei-lazarov:
- requires https://github.com/Koenkk/zigbee-herdsman-converters/pull/12132

WhiteLabels without fingerprints don't generate device pages. They are linked to the main model.

**I added the possibility to link them to a specific model** (to a whiteLabel with fingerprint).

I need this because firmware is not unique per brand.
I can not rename the `_TZ3000_zgyzgdua` fingerprint to Nous, because there is `_TZ3000_zgyzgdua` from Moes and more Tuya vendors.

inception
<p>

PR #5201, opened 2026-06-03
Fix last updated info again

andrei-lazarov:
Don't merge, please see below

## Issue

Every device shows the wrong date again: https://www.zigbee2mqtt.io/devices/ZBMINI.html
> Page was last updated on: 5/31/26, 10:33 PM

If we look into history: https://github.com/Koenkk/zigbee2mqtt.io/commits/master/docs/devices/ZBMINI.md
A new line was added by docgen in Bump zhc commit:

- Of course caused by my PR: https://github.com/Koenkk/zigbee2mqtt.io/pull/5097, or actually https://github.com/Koenkk/zigbee2mqtt.io/pull/5122

## Fix

We must reset the history before the docgen and remove the line

### What I've done already

1. Reset master before docgen
2. Reset dev before docgen

3. Added the image fixes from old dev to new master
4. Removed empty line in new master
5. Merged new master into new dev
6. Tested build: ZBMINI.md remains unchanged and date becomes correct  10/16/25 (lol, we use NA format)
7. Backed-up the current state: [old_master](https://github.com/andrei-lazarov/zigbee2mqtt.io/tree/old_master) and [old_dev](https://github.com/andrei-lazarov/zigbee2mqtt.io/tree/old_dev)

### What you must do

1. Hard-reset Koenkk/master to [andrei-lazarov/master](https://github.com/andrei-lazarov/zigbee2mqtt.io/tree/master) and force-push
2. Cancel the auto-actions, or wait for them to finish
3. Hard-reset Koenkk/dev to [andrei-lazarov/dev](https://github.com/andrei-lazarov/zigbee2mqtt.io/tree/dev) and force-push
4. The usual on master branch: bump zhc, docgen, merge to dev, deploy 🤞🏻


Koenkk:
I think we should look into a proper fix to prevent this from happening again, what about making docgen never change the add date?

andrei-lazarov:
Addition date is correct.
This is about the last updated date (git history of the page). It is working as expected now:
- If the file is different after docgen, date is updated
- If there is no change after docgen, date stays the same

The only issue is, that I added white space by accident.. This bumped the date for _all devices_ for no reason. I gave you the _option_ to correct it, because it is my fault 😁

If it's too complicated to fix, you can close and leave it like that. It's not important.

The only way to prevent this, is for me to triple-check when making a PR 😅

Koenkk:
Let's leave it as is for now, I think it's not the most annoying thing.

PR #5206, opened 2026-06-03
Update 1-ZB-WSD.md

nilservice:


Koenkk:
Thanks, but this will be automatically updated with the next release based on the contents in: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/shelly.ts

PR #5210, opened 2026-06-05
Updates Aqara H2 EU behaviour

TheNetworkGuy:
Removes invalid information regarding the Aqara H2 EU. After testing, it does NOT expose double_up, hold_up and release_up.

### DIFF
--- a/docs/devices/WS-K07E.md
+++ b/docs/devices/WS-K07E.md
@@ -142,7 +142,7 @@ If value equals `true` multi click is ON, if `false` OFF.
 
 ### Action (enum)
 Triggered action (e.g. a button click).
-Since the top button is a physical relais it only exposes 'single_up'. The down button exposes more states such as 'double_down'.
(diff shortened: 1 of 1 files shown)


andrei-lazarov (changes requested):
Everything below Notes is auto-generated, based on templates in this repository, and on the zigbee-herdsman-converters code for the respective device.

> ATTENTION: This file is auto-generated through docgen!
> You can only edit the "Notes"-Section between the two comment lines "Notes BEGIN" and "Notes END".

> Notes END: Do not edit below this line


Your changes will be overwritten.
Basically Z2M listens for those commands, in case the device ever sends them.
You can edit the source code to stop listening for those commands (and prevent display in docs), or add a Note here: "multi-press and long-press works only on the bottom button"

TheNetworkGuy:
Got it, i'll revert my change and update the source code accordingly. Thanks for pointing me in the right direction.

PR #5216, opened 2026-06-06
fix: Clarify ZBT-2 hardware flow control

burmistrzak:
Comments in the Nabu Casa firmware repository [confirm](https://github.com/NabuCasa/silabs-firmware-builder/blob/ffbb9030b026b92b7347e127828803e84461363c/manifests/nabucasa/zbt2/zbt2_zigbee_ncp.yaml#L70) that the ZBT-2 only features hardware flow control **between** the **ESP32-S3** (acting as USB-UART adapter) and the **MG24** SoC.

The host application (i.e. Z2M) should use software flow control instead.

### DIFF
--- a/docs/guide/adapters/emberznet.md
+++ b/docs/guide/adapters/emberznet.md
@@ -54,12 +54,12 @@ Each category is ordered by chip, newer series first.
 
 Chip: [EFR32MG24A420F1536IM40](https://www.silabs.com/wireless/zigbee/efr32mg24-series-2-socs/device.EFR32MG24A420F1536IM40?tab=specs)
 
-With external antenna and hardware flow control support. Make sure to set the baudrate to 460800.
+With external antenna and internal hardware flow control (between ESP32-S3 and MG24). Make sure to set the baudrate to 460800.
 
 ```yaml
 serial:
     baudrate: 460800
-    rtscts: true
+    rtscts: false
 ```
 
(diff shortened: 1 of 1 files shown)


puddly:
Let's keep the docs as-is. I'd rather not have fragmented serial config when we release updated UART firmware.

ZHA uses hardware flow control for all ZBT-2s.

burmistrzak:
> Let's keep the docs as-is. I'd rather not have fragmented serial config when we release updated UART firmware.

Sure. 👌
Just that I get it right: You're planning to add flow control between the host and ESP32 as well, but it's currently not ready yet?

burmistrzak:
I assume you were referring to this https://github.com/NabuCasa/zwave-esp-bridge/pull/15 PR?

puddly:
Sort of. We'll likely have an ESPHome firmware released at some point that replaces the native esp-idf serial proxy.

burmistrzak:
> Sort of. We'll likely have an ESPHome firmware released at some point that replaces the native esp-idf serial proxy.

Oh hell yeah! 🤟
Let me know if you need any helping hands.

puddly:
Totally! I'll tag you in a roadmap when we formalize things a bit more in the future. For the most part everything is already done, there isn't much, just gotta plan out the ESPHome-over-raw USB part a bit more and see what it takes to get the ESPHome API working over non-TCP transports.

Koenkk (approved):

PR #5218, opened 2026-06-07
Add Eglo 901463

panther7:
- Update icon for Eglo 900566
- Add Eglo 901463
  - https://github.com/Koenkk/zigbee-herdsman-converters/pull/12391

Any tip, how add "901463" to search devices, i try create new md file with redirect.

PR #5222, opened 2026-06-09
Fix links and pics for EGLO 900566 & 901463

andrei-lazarov:
Fixes:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/12391
- https://github.com/Koenkk/zigbee2mqtt.io/pull/5218

@panther7 The images must have the exact names as the white-labels.

With wrong name they don't display on z2m website:
https://www.zigbee2mqtt.io/supported-devices/#s=led%20with%20adjustable

Here is an example: IKEA Matter plug
https://www.zigbee2mqtt.io/supported-devices/#e=switch&v=IKEA&s=matter

<p>

Unfortunately in Z2M we can't differentiate between them. It will display the generic image (which can be whatever, one of the models is ok). The users can change the icon in their Z2M instances.

---

The white-label redirect links are a nice addition. But don't add too many of them - we should automate this process instead


panther7:
Thanks, I asked how to do that, but no answer.

Maybe add new image for "merge" both images.

andrei-lazarov:
I missed the pull request 🙂 so I answered now and fixed it.

You can send an image with both models merged if you want.
But it's also ok like this (There could be 3, 4.. 100 models, who knows)

panther7:
Ok, bad is that AwoX have same "IDs" for multiple models alá _EBF_RGB_Zm_ :-/

Koenkk:
Thanks!

PR #5243, opened 2026-06-18
docs: SONOFF SNZB-02DR2: document external sensor display

shaggyone:
This is an update for an existing device.

Sonoff SNZB 02DR2 supports displaying temperature and humidity from external source. The solution is based on a discussion:
https://github.com/Koenkk/zigbee2mqtt/discussions/31115

Converter is also updated:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/12463


Koenkk:
thanks!

PR #5245, opened 2026-06-19
Add _TZ3210_qlmnxmac

phyzical:
This is actually the exact picture used for  MG-AUZG01, but does look like the device. 

duplicated for sake of process


phyzical:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/12451#issuecomment-4744976303

Koenkk:
Thanks!

PR #5246, opened 2026-06-19
Image for Heimgard Entry Lock (HT-SLM-3)

oddeirik:


Koenkk:
thanks but it seems we already have an image for this device: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/HT-SLM-3.png

PR #5250, opened 2026-06-21
add image for NoDieby ND-01

MarianKuschel:
Device picture for the NoDieby ND-01 (512x512, transparent background). Companion to the converter PR in zigbee-herdsman-converters.


Koenkk:
Thanks!

PR #5261, opened 2026-06-23
Document Shelly 2PM cover tilt discovery

MaxRink:
## Summary
- Document that Shelly 2PM Gen4 cover tilt is exposed only when slat control is enabled.
- Document the `cover_tilt_enabled` auto/override behavior.

## Related code PR
- Koenkk/zigbee-herdsman-converters#12497

## Validation
- `git diff --check`
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.12.1 pretty:check`

## Draft status

Kept as draft until Koenkk/zigbee-herdsman-converters#12497 is ready and the final default/auto-detection behavior is confirmed. This docs branch should match the source PR exactly before leaving draft.


Koenkk:
thanks!

PR #5266, opened 2026-06-23
Update SWV.md

pieterdevriesch:
Improved description of Valve Work State entity. The name itself is not very clear and I had to find this out for myself when creating an irrigation automation.


Koenkk:
Thanks!

PR #5267, opened 2026-06-23
Update SNZB-01M.md with button binding details

sephore:
Added a note documenting the binding behavior introduced in firmware 1.1.0, as obtained directly from SONOFF support.


Koenkk:
thanks!

PR #5273, opened 2026-06-23
Document raw attribute readback

MaxRink:
Documents converter-author guidance for custom cluster raw attribute readback.

Related code PR: Koenkk/zigbee-herdsman-converters#12509

Verification:
- git diff --check
- PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.12.1 pretty:check


docs/advanced/support-new-devices/01_support_new_devices.md
```diff
@@ -81,6 +81,12 @@ Repeat this last step until your device no longer produces log messages like `No
 
 If none of the existing converters fit, you can add custom ones, an example for this can be found [here](https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/externalConvertersExample/freepad_ext.mjs).
 
+For manufacturer-specific or custom clusters, generated definitions can also
```
Koenkk: I don't think this is the correct place to document this, maybe it's better to document in the zhc `README.md`?

PR #5274, opened 2026-06-23
Add image for Somfy Sonesse 40

mikesorrenti:
Somfy Sonesse 40 wirefree zigbee li-ion usb-c

From https://www.somfypro.com/products/-/e-cat/1245993


Koenkk:
Thanks!

PR #5283, opened 2026-06-27
docs: document mqtt.server_name TLS SNI override

rohankapoorcom:
Documents the new `mqtt.server_name` setting added in Koenkk/zigbee2mqtt#32417.

`mqtt.server_name` overrides the TLS SNI / certificate-hostname used when connecting to the MQTT broker, without disabling certificate validation. This is useful when connecting to the broker via a hostname or IP that is not listed in the broker certificate's Subject Alternative Names (e.g. an internal Docker/Kubernetes service name) while the certificate is only valid for a public domain.

Adds the setting to the MQTT server connection example in `docs/guide/configuration/mqtt.md`, grouped with the other TLS options.

Made with [Cursor](https://cursor.com)


Koenkk:
Thanks!

PR #5287, opened 2026-06-29
Expand calibration and actions documentation for DWZTCGQ11LM

absent42:
Added calibration instructions for object mode and clarified actions are only reported when device is in Object mode.


Koenkk:
Thanks!

PR #5290, opened 2026-06-29
Add MakeGood-MG-GPO02Z.png

KablammoNick:
Add image for MakeGood MG-GPO02Z to pair with https://github.com/Koenkk/zigbee-herdsman-converters/pull/12586

Noting that there is also a v2 whitelabel from retailer Sparkelec that does look different - but internally they are the same.


Koenkk:
Thanks!

PR #5299, opened 2026-07-02
Add Nova Digital SA-WK device image

felipemathies:
512×512 transparent PNG for the **Nova Digital SA-WK** (Safira "Full Switch": 1 gang + 20A socket + USB-A + USB-C, 4x2).

Pairs with converter PR Koenkk/zigbee-herdsman-converters#12610, which adds `_TZ3210_lqb7lcq9` as a `whiteLabel` (model `SA-WK`) of `TS011F_2_gang_2_usb_wall`.


Koenkk:
Thanks!

PR #5300, opened 2026-07-03
Add image for Innr RCL242C

hidrissi-hass:
512×512 transparent PNG for Innr Round Ceiling Light 42cm (White & Colour) (RCL242C)


Koenkk:
Thanks!

PR #5301, opened 2026-07-03
Damien alt sudo patch led2402 c3 pairing

damien-alt-sudo:
Update pair instructions for LED2402C3 (KAJPLATS E14 bulb) after trying 15 times and failing to pair. 

Worked with 12 as seen on the HomeAssistant Forums.


andrei-lazarov (changes requested):
Sorry for the confusion. I know the clear E27 models need 15 cycles, and I assumed the clear E14 models would work the same.

For the PR, this part is auto-generated. It will get over-written.
You can edit here:
https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docgen/device_page_notes/ikea_kajplats.ts
I guess we can do a generic "12/15 times, depending on model"

PR #5305, opened 2026-07-05
Add image for Moes ZRS-USC-WH

radinsky:
Adds a device image for the Moes ZRS-USC-WH smart curtain switch white-label entry.

Image source: user-provided product photo of the physical device.


Koenkk:
Thanks!

PR #5318, opened 2026-07-11
SONOFF MINI-ZB2GS doesn't support groups on endpoint 2

andrei-lazarov:
- https://github.com/Koenkk/zigbee2mqtt/issues/32534


Koenkk:
Thanks!

PR #5320, opened 2026-07-13
docs: add Simplified Chinese (zh-CN) locale

trphoenix:
This PR adds a Simplified Chinese translation of the documentation.

## What's included

- VuePress i18n setup: site/theme/docsearch `locales` config with a language switcher (`vuepress.config.ts`, `navbar.ts`, `sidebar.ts`). The English site keeps its original paths and content — no English pages were modified.
- Chinese homepage and translations of all 70 guide + advanced pages, mirrored under `docs/zh/`.

## Translation conventions

- Code blocks, inline code, config keys/values, MQTT topics, CLI commands and URLs are kept untranslated.
- A consistent glossary is used (e.g. pairing = 配对, coordinator = 协调器), with the English term noted in parentheses on first occurrence per page.
- Image references use absolute `/images/...` paths so both locales share the same assets.
- Auto-generated device pages (`docs/devices/`, `docs/supported-devices/`) are not translated; links from the Chinese pages point to the English device pages.

## Verification

- File parity check: 70 English pages ↔ 70 Chinese pages, matching heading structure and code-fence counts.
- `prettier --check` passes; full `vuepress build` succeeds and emits `dist/zh/` correctly.

Known limitation: DocSearch results only cover English content until the crawler indexes `/zh/`.


Koenkk:
Thanks for the PR, but given the maintenance we will not support translations for the docs

PR #5323, opened 2026-07-13
Recommend Shelly firmware updates

### LINKED ISSUE #12681: S4PL-00116US (Shelly Plug US Gen4): activePower reports negative values on high-draw loads
## Device

**Model:** S4PL-00116US (Shelly Plug US Gen4)
**Z2M version:** 2.12.1
**zigbee-herdsman-converters version:** 26.81.0

## Issue

The activePower attribute on the haElectricalMeasurement cluster reports **negative values** for loads above approximately 280-310W. The value is consistent (not varying), which is characteristic of an INT16 sign-bit overflow rather than reactive power.

**Observed behaviour:**
- Kitchen fridge at compressor start (~310W draw): power reported as -310W
- Standing desk at typical draw (~200W): power reported correctly as +200W
- Negative value appears during high-draw cycles and is always the same negative value for that load level

## Existing fix

The device definition already includes shellyModernExtend.shellyPowerFactorInt16Fix(), which patches the powerFactor attribute type from INT8 to INT16. However this only addresses the powerFactor attribute. The activePower attribute (0x050B) appears to be affected by a similar INT16 issue.

## Hypothesis

The activePower attribute may be reported by the Shelly firmware using a value or multiplier/divisor combination that overflows INT16 interpretation at higher wattages. The threshold behaviour (positive below ~280W, negative above) is consistent with a sign-bit flip.

## Workaround

Applying Math.abs() to the power and current values in the electrical_measurement fromZigbee converter resolves the display issue. Currently working around this with an HA template sensor.

## Notes

- Confirmed on 2 separate S4PL-00116US units (kitchen fridge and standing desk)
- The standing desk does not exhibit this because its typical draw stays below the threshold
- Happy to test any proposed fix on hardware

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/12681

### DIFF
--- a/docgen/device_page_warnings/index.ts
+++ b/docgen/device_page_warnings/index.ts
@@ -1,7 +1,10 @@
 import {Definition, Expose} from 'zigbee-herdsman-converters';
 import {ikeaMatter} from './ikea_matter';
+import {shellyFirmware} from './shelly_firmware';
 
-const warnings = [ikeaMatter];
+// Warnings are placed above Notes
+
+const warnings = [ikeaMatter, shellyFirmware];
 
 export function getWarnings(definition: Definition, exposes: Expose[]) {
     return warnings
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #5337, opened 2026-07-18
Set 4-space indent for GitHub YAML files

Rotzbua:
Update `.editorconfig` to apply `indent_size = 4` for `.github/**.yml` files as currently used.

This keeps GitHub workflow/config YAML formatting consistent.


Koenkk:
thanks!

PR #5347, opened 2026-07-20
Update FPS-4-1-00 image and add humidity notes

nthiers:
## Summary
- Replace the NodOn FPS-4-1-00 device image with an updated official product render.
- Add a `## Notes` entry warning that the exposed `humidity` value is not reliable (the onboard humidity sensor does not function correctly on this model) — only `temperature` should be trusted.

## Test plan
- [x] Image is 512x512, consistent with other device images.
- [x] Notes section keeps the required `Notes BEGIN`/`Notes END` markers, uses no h1/h2 headings, no raw HTML tags.


Koenkk:
Thanks!

PR #5351, opened 2026-07-22
Replace AOYAN AY301Z image

zyjsmile857:
Replace the AOYAN AY301Z device image with the updated product image.\n\nImage file: public/images/devices/AY301Z.png


Koenkk:
thanks!

PR #5354, opened 2026-07-23
Add device picture for NodOn SEM-4-3-20

nthiers:
Device picture for the NodOn SEM-4-3-20 three-phase energy meter, added ahead of zigbee-herdsman-converters#12752.


Koenkk:
Thanks!

PR #5370, opened 2026-07-30
add docker-compose example with EMQX

id:
Adds an external end-to-end Docker Compose example that runs Zigbee2MQTT with EMQX as its MQTT broker. The current Docker guide covers Zigbee2MQTT itself but assumes a broker is already available.

Disclosure: I authored the linked guide and work for EMQ. This does not change Zigbee2MQTT’s Mosquitto recommendation.


Koenkk:
thanks!

PR #5382, opened 2026-08-03
Add notes about device endpoint configuration

fst-dresden-elektronik:
Added notes section explaining device endpoint behavior based on DIP switch configuration.


Koenkk:
thanks!

PR #5400, opened 2026-08-11
Update ZHT-002.md

Di4bLo84:
Update ZHT-002 documentation to match the new converter


Koenkk:
Thanks but this will automatically be generated after the next release

Di4bLo84:
> Thanks but this will automatically be generated after the next release

Even better :-D

PR #5402, opened 2026-08-12
Update OWON DWS312 device image

jacky202509:
Update the OWON DWS312 device image.

DWS332-E is the current replacement model for DWS312 and shares the same device definition through whiteLabel. Since DWS312 is no longer manufactured, update the shared image to the current model image.


Koenkk:
Thanks!

PR #5407, opened 2026-08-14
Revise pairing instructions for Range Extender Zi

GeVanCo:
Updated instructions for pairing the Range Extender Zi, including button press duration and LED behavior.


Koenkk:
thanks!

PR #5410, opened 2026-08-14
Image and docs note for 929004296801

reeett:
Adding Image and docs note for 929004296801 Hue wired wall switch module 2 channel


Koenkk:
thanks!

PR #5413, opened 2026-08-15
Update SBHT-103C.md

buz-ch:
Explain how Zigbee Clock Sync can be activated

Have validated the process on a newly unpacked device.


Koenkk:
Thanks!

PR #5418, opened 2026-08-18
feat: add usage troubleshooting page

Nerivec:
Been repeating some of this quite a lot recently, figured a docs page for the most common troubleshoots might be better. Might also stop AIs from giving crappy/convoluted answers (at least a little...).
I moved some paragraphs that were stashed in the crash page for lack of better place, which there now is.

I suggest we keep this page to _most_ common issues. And also do not add device-specific ones (which as mentioned at top of page, should be in device-specific pages).
We should not put links to issues in there, if necessary, we should add a recap instead (issues often contain off-topic/outdated, that will just confuse).
I also don't think we should put temporary issues in there; that will just confuse again (including search engines/AIs).

There's probably more stuff that could be moved to this new page, but we can do that as someone stumbles on it :grimacing: 

CC: @andrei-lazarov


andrei-lazarov (approved):
Very nice 🙂 This was really needed

Koenkk:
Nice, thanks!

PR #5420, opened 2026-08-18
Put correct image for ThirdReality E3 smart plug

GeVanCo:
The current image for the ThirdReality E3 Smart Plug was still the old ThirdReality E2 Smart Plug.  This, while the new image was already in the repo, be it with an extra space at the end of the name.
Removed the space and as such, replaced the old image.


Koenkk:
Thanks!

PR #5428, opened 2026-08-20
feat: add gfm alert support

Nerivec:
Moves all `:::` alerts to GitHub style alerts.
It cleans up the docs quite nicely, and of course, has the added benefit of being visible if consulting the docs on GitHub directly.
Added the dep explicitly, even though it was already pulled in by vuepress internally; more stable since we're using the import now.
Also cleaned up a few levels as I went through the list to use the full range.

https://ecosystem.vuejs.press/plugins/markdown/markdown-hint.html

I suggest we reject the use of `:::` for alerts in future PRs, and reserve it for `details`.

Also fixed the new troubleshooting page, which wasn't added to the sidebar. :hammer: 

cc: @andrei-lazarov

### DIFF
--- a/docgen/generate_settings.ts
+++ b/docgen/generate_settings.ts
@@ -221,9 +221,8 @@ pageClass: settings-page
 
 This page contains all currently supported settings in \`configuration.yaml\`.
 
-::: tip NOTE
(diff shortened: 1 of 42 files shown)


andrei-lazarov (approved):

Koenkk:
thanks!

PR #5429, opened 2026-08-21
docs: add Rti-Tek STHZB device image

QuAnnan2019:
Adds the 512x512 transparent product image for the Rti-Tek STHZB temperature and humidity sensor.


Koenkk:
thanks!

PR #5434, opened 2026-08-22
fix: add sparkfun mgm240p not rec

Nerivec:
Been meaning to add that one for a while...

CC: @chris-1243


chris-1243:
The description is perfect. This is clearly what you might expect from it.
To do some test, it is fine but I would never use it ad a main corrdinator!

Koenkk:
thanks!

PR #5453, opened 2026-08-28
Add SC0002 device image

Ales2-k:
Adds a 512×512 PNG device image for the Selena SC0002 desktop environmental monitoring station.


Koenkk:
thanks!

PR #5455, opened 2026-08-29
added LF101W2-Mod lamp

slacky1965:
Industrial lamp. Sold at a local online store in Moscow - https://ledron.ru/catalog/umnyy-dom/umnye-lampy/smart-lamp-d60/


Koenkk:
thanks!

PR #5460, opened 2026-08-30
feat(add): CS-T10C device picture

Gren-95:
Device picture for **EZVIZ CS-T10C** (water leak sensor), as requested in Koenkk/zigbee-herdsman-converters#12992.

- `public/images/devices/CS-T10C.png`
- 512x512, PNG, transparent background

The filename matches the `model` in the converter definition (`CS-T10C`), alongside the existing `CS-T2C.png` from the same vendor family.


Koenkk:
thanks!

PR #5466, opened 2026-08-31
Add 42-050 device picture

mikkelke:
Device picture for Light Solutions 42-050 (rebranded Sunricher SRP-ZG9105-12-50CVT), requested in Koenkk/zigbee-herdsman-converters#12943.


andrei-lazarov (commented):
Can you fix the picture, please?

mikkelke:
Fixed in #5499 — thanks for catching it. The floodfill I used to cut the background out bled through the casing (the housing is the same pure white as the backdrop), which put holes in the product body. Re-masked from the outline instead; housing is intact now.

PR #5471, opened 2026-09-02
feat(device): add Nova Digital ZCMR-1 image

caio-emoraes:
Adds the device image for the Nova Digital ZCMR-1.

The hardware is visually identical to the existing CMR-1, so its existing image was reused.

Related device support PR: Koenkk/zigbee-herdsman-converters#13078


Koenkk:
thanks!

PR #5494, opened 2026-09-06
feat(add): MW786Z device picture

isaackennylee:
Device picture for the Mowe (Möwe Smart Home) TS0001 white label added in Koenkk/zigbee-herdsman-converters#13134:

- `MW786Z.png` — water heater switch

512x512 PNG with a transparent background, produced from the manufacturer's own product render for this SKU on the Mausklick Series water heater switch page at mowesmarthome.com.

Note: this render is the same one the vendor uses for the Mausklick MW781Z 1 gang switch, because the two products share a faceplate — the MW786Z is the 30A water heater variant. The image is therefore very similar to the existing `MW781Z.png`, which is intentional rather than a mistaken duplicate.


Koenkk:
thanks!

PR #5497, opened 2026-09-06
fix: correct ECODO device picture for PSL-24V/RGBCW/ECD

isaackennylee:
Companion picture for Koenkk/zigbee-herdsman-converters#13139, which corrects the ECODO whiteLabel for `_TZ3210_rnj5wxxg` from `ECD-SS12` (a recessed downlight — see the existing `ECD-SS12.png`, a completely different fixture) to `PSL-24V/RGBCW/ECD`, an all-in-one 240 W power supply for RGBCW/RGBCCT LED strip.

docgen sanitizes `/` to `-` for image filenames (`getImageName()` in `docgen/utils.ts`), so this file is named `PSL-24V-RGBCW-ECD.png` to match the corrected model string.

No clean studio product shot exists for this vendor's product (small reseller, no manufacturer site with press photos — same situation as the earlier Mowe whiteLabels). This is a real installation photo showing the unit's own printed label (`ECODO Constant Voltage Dimming Toning LED Driver ... Output: DC24V 10A Max 240W Max ... ZigBee`), confirming both the brand and the 240 W RGBCW spec firsthand.

The wall has been keyed out and the driver cut onto a transparent background to match the other device images — 512x512 RGBA, same as the existing `ECD-SS12.png`. Two things are trimmed deliberately: the dangling output wire, which is the installer's rather than part of the product, and the two thin mounting-bracket ears at the right-hand end, because the wire loops through the gap between them and the ears cannot be kept without keeping a patch of wall along with them. The left end sits flush with the frame because the driver runs off the edge of the original photo there.


Koenkk:
Please make the background transparent

isaackennylee:
Checked the pushed PNG (public/images/devices/PSL-24V-RGBCW-ECD.png) directly: it's RGBA, 512x512, alpha channel confirms genuine transparency — ~81% of pixels are fully transparent (alpha=0) including all four corners, ~18% fully opaque (the driver), ~1% partial alpha at the antialiased cutout edge. So the background is transparent in the file itself. If it's rendering with a background in your viewer, that may be how GitHub's PR diff preview or a cache is displaying it rather than the PNG's actual alpha data — happy to re-export or adjust if you're seeing something specific.

isaackennylee:
Pushed an update: the image is now squared to 512x512 (matching the other device images, including the existing ECD-SS12.png this PR corrects) with the driver centered on a transparent background. Confirmed the transparency is genuine (RGBA, alpha channel verified) rather than a viewer artifact — let me know if it still doesn't look right on your end.

PR #5499, opened 2026-09-07
fix(image): redo 42-050 device picture

mikkelke:
Follow-up to #5466 — @andrei-lazarov you were right, that picture was broken.

The cut-out was made with a corner floodfill, and the driver's housing is the same pure `#FFFFFF` as the photo background, so the fill bled straight through the casing and punched holes in the product body. It only showed up against a dark background.

Re-masked from the outline instead of the colour: threshold the non-white pixels, morphologically close the outline, fill the interior, and clip the mirror reflection under the front face. Same source photo (the 12&nbsp;V variant, so the label still reads `SRP-ZG9105-12-50CVT`), same 512x512 transparent PNG, housing now intact.


Koenkk:
thanks!

PR #5500, opened 2026-09-07
Add CBE 1443ZK device image

animaluslex:
Adds the device image for the CBE 1443ZK thermostatic radiator valve.

Zigbee model: TS0601
Manufacturer: _TZE284_16m4bgsv


Koenkk:
thanks!

PR #5501, opened 2026-09-07
Image for Immax Neo Smart Water Leak Sensor (07519L)

michal-kocarek:
Adding image for 07519L (https://www.immax.eu/immax-neo-smart-water-leak-sensor-zigbee-3-0-p14936/)

Definition is to follow

image


michal-kocarek:
As mentioned in the device support PR: https://github.com/Koenkk/zigbee-herdsman-converters/pull/13149, there are two reported flags that are not supported. So the device support is not complete at this point. Is this worth adding into the model markdown?

Koenkk:
thanks!

PR #5505, opened 2026-09-09
fix(image): remove orphaned ECD-SS12 device picture

isaackennylee:
`ECD-SS12` no longer exists in zigbee-herdsman-converters. Koenkk/zigbee-herdsman-converters#13139 (in 26.108.0) replaced that whiteLabel outright rather than renaming it, because the hardware behind `_TZ3210_rnj5wxxg` had been misidentified — it is an ECODO `PSL-24V/RGBCW/ECD` constant-voltage LED driver, not a recessed downlight. The string appears nowhere in zhc master now, so `public/images/devices/ECD-SS12.png` is unreferenced. Its replacement picture is #5497.

`docs/devices/ECD-SS12.md` is deliberately left alone: docgen renames device pages on the next zhc bump by itself (as it did for `STH1Z.md` → `STHZB.md` in 2dee44c), so it needs no manual edit here.

One ordering note — that page is still on master until the bump lands, and it embeds this image, so if this merges first the stale page will show a missing picture until then. Merging it after the next zhc bump avoids the gap entirely; happy to hold it if you prefer.


isaackennylee:
Closing this — the premise was wrong on two counts, sorry for the noise.

`ECD-SS12` is a real ECODO product. It lost its zhc entry only as a side effect of Koenkk/zigbee-herdsman-converters#13139, which reassigned `_TZ3210_rnj5wxxg` to the PSL-24V/RGBCW/ECD driver once the manufacturer ID turned out to have been read off the wrong unit. The downlight's own fingerprint hasn't been recorded yet, so the model is unsupported-by-name rather than retired — and this picture is a genuine photo of it that a future definition would want under exactly this filename. Better to keep it.

CI disagreed on the repo's own terms too: the image test walks the docs tree, and `docs/devices/ECD-SS12.md` is still on master, so the build fails with `Missing image for Model ECD-SS12`. The picture was never orphaned in the sense I claimed.

PR #5507, opened 2026-09-09
docs: CCTFR6000: explain demand percentage, cycle time and missing fail-safe

lxhunter:
Adds a Notes section for the Schneider Electric CCTFR6000 (Wiser UFH controller, sold as CCTFR6600/CCTFR6610).

The device is a time-proportional actuator: a relay only closes while the channel state is ON and the manufacturer-specific demand percentage is above 0, and a demand value persists until overwritten. Both facts were verified on a CCTFR6610 with Zigbee2MQTT 2.14.1 and are needed to use the device at all; the exposes are added in Koenkk/zigbee-herdsman-converters#13162. Also documents the pairing procedure from the vendor manual.

### DIFF
--- a/docs/devices/CCTFR6000.md
+++ b/docs/devices/CCTFR6000.md
@@ -24,8 +24,18 @@ pageClass: device-page
 
 
 <!-- Notes BEGIN: You can edit here. Add "## Notes" headline if not already present. -->
+## Notes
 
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

lxhunter:
My pleasure!

PR #5531, opened 2026-09-17
Correction of syntax on #5506

badr-ctrl:
i accidentally added an exclamation mark that turned the link into an image (?) rather than a hyperlink
original PR : #5506


Koenkk:
Thanks!

PR #5536, opened 2026-09-18
feat(add): TS0601_wsek35um image

mfarad:
Adds the 512×512 transparent product image for the Tuya TS0601_wsek35um radiator thermostat.


Koenkk:
Thanks!

PR #5541, opened 2026-09-20
Add device image for LUUMR Smart LED G9 2,5W CCT Clear

kaistian:
Device image for LUUMR Smart LED G9 2,5W CCT Clear model 10010339.


Koenkk:
Thanks!

PR #1661, opened 2022-10-26
EFR32MG21.Router

BWFCGUIXT:
Hi Koen,

Not sure if the .MD is correct, should I do it differently please let me know.

PR #5543, opened 2026-09-20
docs: SONOFF ZBBridge router (EFR32MG21.Router.1) and ZBDongle-P router (ti.router)

alichaudry:
Two routers that are already supported but documented for the wrong hardware or not at all.

**EFR32MG21.Router.1 (SONOFF ZBBridge flashed with router firmware)**
- Picture: the page showed the UT-02 USB-stick image, copied over when the model was split in 602f4ec6. Replaced with a picture of the ZBBridge.
- Notes: the previous notes (red LED states, "press BOOT", RouterForDongle firmware link) were written for the UT-02 dongle firmware (#1661) and copied to this page on the split. Replaced with what applies to the bridge: firmware source (digiblur / xsp1989, 6.7.10), flashing via Tasmota following digiblur's guide, pairing, LED behaviour. Pairing and LED text follow the guide; on my unit the bridge rejoins automatically after a power cycle and both LEDs are off in normal operation.

**ti.router, including the SONOFF ZBDongle-P**
- Picture: `ZBDongle-P.png` for the white label added in Koenkk/zigbee-herdsman-converters#13246. The "White-label" row on the ti.router page is generated by docgen, so it is not part of this diff.
- Notes: firmware now lives in the Z-Stack-firmware releases (the old `router/Z-Stack_3.x.0/bin` link is dead); which build the ZBDongle-P uses; all boards report `ti.router`; `transmit_power` needs firmware 20221102+ (my ZBDongle-P router on 20220125 returns UNSUPPORTED_ATTRIBUTE on read); LED: the router firmware has no LED status indication (all router builds set `CUI_DISABLE` in firmware.patch), and the ZBDongle-P's red LED is a power indicator per the Z-Stack-firmware hardware table ("DIO_7 (Not welded), Power LED (Red)").

**zstack.md, ZBDongle-P section**
- One line on the LED and `disable_led`, same source.

Pictures: both are ITEAD's own product shots (itead.cc) with the background removed, as the contribution guide describes. If you'd rather not carry vendor images for these two, say so and I'll replace them with my own photos.

Not verified by me: LED behaviour on boards other than the ZBDongle-P and ZBBridge. That statement rests on `CUI_DISABLE` in the router firmware patch.

Companion converters PR: Koenkk/zigbee-herdsman-converters#13246


Koenkk:
Thanks!

